PR #203, opened 2014-09-23
[WIP] Rewrite ByPropertyIdArray

Benestar:
This should replace `ByPropertyIdArray`. Just a draft for now, needs lots of testing to make sure it actually replicates the behaviour of the old version of the class. However, I'm already quite proud of using only half as much codes as the old implementation.


Benestar:
@JeroenDeDauw @thiemowmde is it worth to work on this PR again or are there already alternatives to `ByPropertyIdArray`?

JeroenDeDauw:
There is the `ByPropertyIdGrouper` that you created... no further work has been done in this direction IIRC

thiemowmde:
I find this class scary, before and after the rewrite. I will not spend time on this.

Benestar:
New approach made in #479

PR #272, opened 2014-11-04
Add requirements for language fall backs.

JanZerebecki:


src/Term/AliasGroupFallback.php
```diff
@@ -0,0 +1,82 @@
...
+
+	/**
+	 * @param string $languageCode
+	 * @param string[] $aliases
+	 * @param string $actualLanguageCode fallen back to
+	 * @param string|null $sourceLanguageCode for transliteration
+	 *
+	 * @throws InvalidArgumentException
+	 */
+	public function __construct( $languageCode, array $aliases, $actualLanguageCode, $sourceLanguageCode ) {
```
brightbyte: So *all* aliases come from the same fallback? That doesn't make sense to me. The application should be free to choose how fallabck is applied to aliases - e.g. all aliases from all applicable fallback languages could be collected. In that case, the the actual language and source language can be different for each alias.

thiemowmde: I think it's not helpful to discuss this on a closed pull request. Please open a ticket.

PR #289, opened 2014-11-17
StatementList constructor accepts Statement

thiemowmde:
See https://gerrit.wikimedia.org/r/#/c/173828/.


thiemowmde:
I reworked this patch to accept _any_ number of `Statement` parameters, not only one. I will do the same for other classes, if requested (please tell me which are relevant and which are not), but not in this patch.

PR #290, opened 2014-11-17
Make all Item constructor parameters optional

thiemowmde:
Main reason to touch this was the `Item` constructor. The first parameter can be null but the others can not? This doesn't make much sense. Why not simply make them all optional?

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -12,8 +12,9 @@
 * Added `ItemLookup` and `PropertyLookup` interfaces
 * Added `ItemNotFoundException`
 * Empty strings are now detected as invalid language codes in the term classes
-* `Item` constructor parameters for site links and statements are optional now
+* Made the `Item` constructor parameters optional
 * Made the `Fingerprint` constructor parameters optional
+* Deprecated `Item::newEmpty` in favour of `new Item()`
 * The `StatementList` constructor now accepts `Statement` objects in variable-lenght argument list format
 
 ## Version 2.4.1 (2014-11-26)
--- a/src/Entity/Diff/ItemDiffer.php
+++ b/src/Entity/Diff/ItemDiffer.php
@@ -113,7 +113,7 @@ class ItemDiffer implements EntityDifferStrategy {
 	 */
 	public function getConstructionDiff( EntityDocument $entity ) {
 		$this->assertIsItem( $entity );
-		return $this->diffEntities( Item::newEmpty(), $entity );
+		return $this->diffEntities( new Item(), $entity );
 	}
 
 	/**
@@ -124,7 +124,7 @@ class ItemDiffer implements EntityDifferStrategy {
 	 */
 	public function getDestructionDiff( EntityDocument $entity ) {
 		$this->assertIsItem( $entity );
-		return $this->diffEntities( $entity, Item::newEmpty() );
+		return $this->diffEntities( $entity, new Item() );
 	}
 
 }
--- a/src/Entity/Item.php
+++ b/src/Entity/Item.php
@@ -41,18 +41,18 @@ class Item extends Entity implements StatementListProvider {
 	 * @since 1.0
 	 *
 	 * @param ItemId|null $id
-	 * @param Fingerprint $fingerprint
+	 * @param Fingerprint|null $fingerprint
 	 * @param SiteLinkList|null $siteLinks
 	 * @param StatementList|null $statements
 	 */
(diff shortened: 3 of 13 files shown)


snaterlicious:
The change absolutely makes sense to me. Using the constructor while omitting its parameters is a much more natural way to instantiate an empty object. Having multiple ways to achieve the same result tends to be confusing. Consequently, deprecating `newFromEmpty` is a reasonable step. I would opt for finally removing `newFromEmpty` in version 3.0. If there are no serious objections, I will merge the change this week.

JeroenDeDauw:
> Having multiple ways to achieve the same result tends to be confusing.

In some cases perhaps. Thought there are clear cases where this is not so. For instance, you can construct the object empty and then add a label to it, or you can construct it with a label in the first place. Do you think having `Item::newEmpty` is confusing? What do you think will go wrong by having it?

> If there are no serious objections, I will merge the change this week.

As noted, I think it makes little sense to do the deprecation, and doing so, at least at this point, is not worth the cost. I do not have the impression this has been thought about, certainly it has not been acknowledged that there is a cost.

It's only that part of the change I -1 against. Just a -1, so no strong objection. In that regard, you do also not have to wait a week with merging.

thiemowmde:
> you can construct the object empty and then add a label to it, or you can construct it with a label in the first place.

How is that an argument for having a static method?

> I do not have the impression this has been thought about, certainly it has not been acknowledged that there is a cost.

You are repeating your "cost" argument over and over again. So I'm not sure if it's helpful to repeat my cost argument, but I will try: There is also a cost for keeping a duplicate, redundant method. That cost is, in my opinion, much higher than the cost of having a deprecate tag.

The cost of finally dropping that method is much higher. But that's not what we are discussing here.

snaterlicious:
I understand that by assigning a "deprecated" tag, the "cost" is triggered instantly because users of the class are urged to apply changes to their code. Eventually, that whole discussion seems to be more or less about rather fundamental principles on our workflow and our philosohpy on "change": Whether to pursue a more liberal fast-paced approach or a more conservative approach focusing on stability. I am a supporter of the idea that the development still can (and should) happen in a liberal way. In my opinion, there are still some things to catch up in the PHP data model (e.g. naming) that need to be addressed before switching to a more conservative procedure.
Let alone personal opinions, there obviously is a lack of a decision process or authority on such subjects. Who decides on such a conflicts? Is it the majority of developers? Is it the majority of developers "interested"? In fact, the change here influences a public interface--should the decision bubble up to product management or whoever is declared "owner" of the component?

thiemowmde:
I would love to decide such things based on arguments I can understand and agree with, not by an authority. Here the only counter-argument I hear still sounds like "I don't want people to know" to me. Which is unsolvable if that's exactly what I want. I want people to know there is a duplicate method and which one they should prefer.

thiemowmde:
What does `@deprecated` mean? (To me. An essay.)

For itself, it means nothing. Everybody can ignore it and absolutely nothing will happen. Just like a wrong `@return` means nothing to the code and can be ignored without anything bad happening.

All a `@deprecated` really does is making code visually different from non-deprecated code, either via strike-through in the IDE or by collecting such lines of code in analysis tools. That's all. It "does" nothing to the code. In that way it's only documentation, nothing else.

What's the "cost" of documentation? Sure, it does have cost because people will see it and, depending on what they think it means, will act on it.

But what it really means depends on what's written in the release notes and in the `@deprecated` line.

1. It _should_ say since when stuff is deprecated,
2. it absolutely __must__ say what the alternative is
3. and it's _nice to have_ an idea when deprecated stuff is going to finally die.

That's also why a lonely, unexplained `@deprecated` is almost worthless. It's only 1 bit of information, right?

1. and 2. obviously aren't a problem here. Both is there. So what we are really talking about is 3. What if I add something like "will be kept forever for compatibility reasons"? Or "will be kept in all 3.x releases and dropped in 4.0"?

snaterlicious:
In my opinion, discussing about the "deprecated" tag is an aspect but does not hit the fundamental point. I am in support of this change and the lack of any proper decision process / authority leads to the definitely frustrating situation that it takes just two developers to change less or more important parts of the code. (How important is a public interface..?) Merging this change demonstrates the failure arising from the current situation. Somebody else may simply submit a pull request reverting this change and just by finding someone else to acknowledge (at least, that is our low-level standard procedure), that action would still comply to the non-existing process.
Back to the change: Personally, I would like `Property:newFromType` to become deprecated as well.

PR #296, opened 2014-11-19
Allow constructing empty Fingerprint

thiemowmde:
This is split from #290. I hope we all agree that the concept of an empty Fingerprint is in the scope of the DataModel.


snaterlicious:
As per comment on #290, that change is a reasonable improvement from my perspective. Going to merge unless there are no serious objections.

PR #299, opened 2014-11-20
Normalize namespaces of all tests

### LINKED ISSUE #252: correct namespace of tests
Like in #250 just for the rest.

thiemowmde:
Fixes #252.

Please note that I picked a naming scheme that was not used in this component before, but it is already in use in the Wikibase Client/Lib/Repo tests.


thiemowmde:
I'm rebasing this for two months now. What's the problem? Pining @JanZerebecki, @JeroenDeDauw, @tobijat.

PR #306, opened 2014-11-25
Add GUID to Statement constructor

thiemowmde:
Both `ClaimList::addNewClaim` and `StatementList::addNewStatement` support an optional GUID parameter. These "new" methods should mirror the constructors. Everything else is just confusing. This is only one step to make them equal again, the other step is #268.

### DIFF
--- a/tests/unit/Statement/StatementTest.php
+++ b/tests/unit/Statement/StatementTest.php
@@ -3,13 +3,16 @@
 namespace Wikibase\DataModel\Tests\Statement;
 
 use DataValues\StringValue;
+use InvalidArgumentException;
 use Wikibase\DataModel\Entity\PropertyId;
 use Wikibase\DataModel\Reference;
 use Wikibase\DataModel\ReferenceList;
 use Wikibase\DataModel\Snak\PropertyNoValueSnak;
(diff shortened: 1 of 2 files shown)


thiemowmde:
I'm sorry? Reason?

JeroenDeDauw:
Because the change no longer applied to our current codebase. I see you modified it now.

thiemowmde:
It still applied to Statement. I just rebased it, removed Claim and added tests.

tests/unit/Statement/StatementTest.php
```diff
@@ -24,6 +27,62 @@
...
+			array( $snak, null, null, 'guid' ),
+			array( $snak, $qualifiers, $references, 'guid' ),
+		);
+	}
+
+	/**
+	 * @dataProvider invalidConstructorArgumentsProvider
+	 * @expectedException InvalidArgumentException
+	 */
+	public function testConstructorWithInvalidArguments(
```
JeroenDeDauw: Two simple methods here would be less code, less complexity and more clarity

thiemowmde: True. I was not happy with this provider too. See #468.

PR #312, opened 2014-11-26
Deprecate Entity::setId

JeroenDeDauw:


JeroenDeDauw:
That suggests option 3 as a solution everyone is reasonably happy with

PR #322, opened 2014-11-28
Clarify contracts of toArray methods

thiemowmde:
As requested in https://github.com/wmde/WikibaseDataModel/pull/311#issuecomment-64713212.

Please note that `StatementList` always reindexed the array. It stopped doing this with #285 (released with 2.4.0).

I think this must be the responsibility of the `StatementList` class (as well as `ClaimList`, see the TODO) and not the calling code.

### DIFF
--- a/src/Claim/ClaimList.php
+++ b/src/Claim/ClaimList.php
@@ -27,6 +27,7 @@ class ClaimList implements \IteratorAggregate {
 
 	/**
 	 * @param Claim[]|Traversable $claims
+	 *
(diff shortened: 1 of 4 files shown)


thiemowmde:
Wow. A super-trivial documentation change blocked for 2 months now. Pinging @tobijat since we apparently still need a nurse.

JeroenDeDauw:
Note how this is not blocked at all (no -2 has been given). I asked you if you could do me a favour, and you got mad about it. Then nobody looked at this.

PR #324, opened 2014-12-10
Remove misleading return doc

thiemowmde:
`Serializable::unserialize` isn't supposed to `@return` anything. Warning, this is a possible breaking change (in contrast to all other `unserialize` implementations in all other components, they never returned anything, but this one did).

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -3,6 +3,7 @@
 ## Version 2.6.0 (dev)
 
 * Empty strings are now detected as invalid in the `SiteLink` constructor
+* Fixed `EntityIdValue::unserialize` accidentally returning itself
 
 ## Version 2.5.0 (2014-01-20)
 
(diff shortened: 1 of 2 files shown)


thiemowmde:
#326 got merged, which makes this change super-trivial now. Please note that I'm not documenting this as a breaking change on purpose. If anybody ever relied on the return value of PHP's core `Serializable::unserialize` they are doing it wrong.

JeroenDeDauw:
Needs rebase. Also, not reason to file this for 3.0, it can go into 2.6. In case we decided to not do 2.6, we can easily move all 2.6 items in the release notes to 3.0 at that point.

thiemowmde:
Since I consider this a bugfix, this could also be a 2.5.1. I don't care much, I just picked what's going to be the next release anyway, in my opinion.

mariushoch:
I agree with Thiemo here: I don't think this will case any harm, thus should certainly be ok for a `2.6`.

PR #317, opened 2014-11-27
Remove claim

JeroenDeDauw:
Note that this is a pull request against the 3.0.x branch, not against master. So we can still make as many 2.x releases as we want before taking this step.

PR #325, opened 2014-12-10
Decrease inheritance in Statement

thiemowmde:
This is a no-op change (it does not change any behavior and can be merged in any branch) split from #268 (a.k.a. #317) to make future patches easier to review.


JeroenDeDauw:
Good idea. Can you also get to reviewing the part of #317 that I did, so we can merge it into the 3.x dev branch?

PR #328, opened 2014-12-18
Add ItemIdSet::getSerializations

thiemowmde:


tests/unit/Entity/ItemIdSetTest.php
```diff
@@ -16,6 +16,35 @@
...
+		$set = new ItemIdSet();
+		$iterator = $set->getIterator();
+
+		$this->assertInstanceOf( 'Traversable', $iterator );
+	}
+
+	/**
+	 * @dataProvider serializationsProvider
+	 */
+	public function testGetSerializations( array $expected ) {
```
thiemowmde: "Mirror". Yes. That's exactly what a test should do. I don't understand why you want to "avoid" that.

tests/unit/Entity/ItemIdSetTest.php
```diff
@@ -16,6 +16,21 @@
...
 class ItemIdSetTest extends \PHPUnit_Framework_TestCase {
 
+	public function testGetIterator() {
+		$set = new ItemIdSet();
+		$iterator = $set->getIterator();
+
+		$this->assertInstanceOf( 'Traversable', $iterator );
+	}
+
+	public function testGetSerializations() {
```
thiemowmde: What "questions" are you referring to? What "technique"?

src/SiteLink.php
```diff
@@ -18,10 +18,18 @@
  * @licence GNU GPL v2+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
  * @author Michał Łazowik
+ * @author Thiemo Mättig
```
thiemowmde: I fail to see an "honest" question. Why are you asking if the added `@author` is unrelated? You can see it is unrelated. It takes you seconds to look at the history and blame of this class and see that I'm a co-author of relevant parts of the code.

Why are you adding that suggestive sentence about me not knowing how `@author` tags should be used?

tests/unit/Entity/ItemIdSetTest.php
JeroenDeDauw: Two main reasons:

* It makes the tests more complex, harder to change, and harder to maintain
* You are more likely to repeat the same mistake (ie missing a case you should handle) in both places

src/SiteLink.php
JeroenDeDauw: I did look at the git blame and did not see much done by you. I checked again now and getting a different impression. Perhaps I incorrectly checked ItemIdSet. My bad, I don't object to you adding your name here.

Can you please try to assume good faith and not throw accusations at me?

JeroenDeDauw:
+1. Needs rebase though.

src/SiteLink.php
thiemowmde: Please try to do the same.

"This change is rather unrelated no?" <- This is closed question. Like you are stating the fact that this is unrelated and asking why I'm doing it nevertheless.

"Are you adding your name as author because you are adding some docs for private code?" <- Again, a closed question. Like the only possibility you can think of is that I'm adding myself because of some private docs. Please ask open questions if you want to know what I think. Such closed questions put me in a corner, where I can't do much but fight.

I know I'm making the exact same mistake all the time. Please make me aware if I do that.

tests/unit/Entity/ItemIdSetTest.php
thiemowmde: What's the "less complex" alternative? Really, I was thinking about this but everything I can come up with is equally "complex" (probably depends on what "complex" means to you). Something like that?
```php
public function serializationsProvider() {
	return array(
		array( array( new ItemId( 'Q1' ) ), array( 'Q1' ) ),
	);
}
public function testGetSerializations( array $itemIds, array $serializations ) {
	$set = new ItemIdSet( $itemIds );
	$this->assertEquals( $serializations, $set->getSerializations() );
}
```

adrianheine: I also tend to make tests too complex in order to be DRY. As @JeroenDeDauw wrote, this introduces the risk of reproducing mistakes you did in the production code, thereby masking these mistakes in the production code.

What I would definitely recommend is moving the instantiation to the provider. The test method should not have to prepare its test input, that should be the responsibility of the provider in my opinion.

thiemowmde: I'm still not sure how this should look like instead, in your opinion. Like I did it now?

src/SiteLink.php
JeroenDeDauw: "This change is rather unrelated no?"

-> No it is not, what makes you think it is?

> "Are you adding your name as author because you are adding some docs for private code?"

-> No, because I made changes to the code in the class before.

I asked the question because I found what I saw very strange. Which Is why I did not say "-1. this change is unrelated". If you had simply answered my question, we'd have saved us a lot of time.

PR #329, opened 2014-12-18
Optional Item constructor parameters

thiemowmde:
This is a part of #290. I hope we can at least agree on that. The two parameters made optional by this patch are:
* `$statements` are already optional in `Property`, but not in `Item`, which is just a bit inconsistent.
* `$siteLinks` are specific for `Item`s.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -10,6 +10,7 @@
 * Added `ItemLoookup` and `PropertyLookup` interfaces
 * Added `ItemNotFoundException`
 * Empty strings are now detected as invalid language codes in the term classes
+* `Item` constructor parameters for site links and statements are optional now
 
 ## Version 2.4.1 (2014-11-26)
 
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
Ok, I'm not sure why you did not do this for all parameters, but I'll just merge this now and submit a follow up.

thiemowmde:
> I'm not sure why you did not do this for all parameters

What the ...? Are you kidding? What do you think #290 is?

JeroenDeDauw:
I'm not sure. You submitted this change after you submitted #290. Why are you submitting the same thing twice?

thiemowmde:
As I said in the commit message this is a resubmission of a non-controversial fragment of #290 because #290 was and still is stuck. Now you are acting like you didn't read that.

JeroenDeDauw:
I did not read that yesterday, at least not in enough detail to get that part. What I saw was a commit you put up to be merged, where I had some concern before, you replied to the concern, and I now decided to just fix the concern in a follow up.

thiemowmde:
I wanted this merged as an acceptable intermediate step everybody can agree on to make #290 smaller and more focused. As said in the commit message.

You are giving me the impression that you have a problem with both
* making all parameters optional (#290, including all the necessary steps to make that a complete, valid patch)
* as well as making only some parameters optional (#329, which was the most minimal change that did not required to discuss deprecation).

PR #333, opened 2015-01-12
Drop static keywords from data providers

thiemowmde:
I'm not sure if this was required in an older PHPUnit version. But I'm sure it's not needed with our current PHPUnit version(s).


JeroenDeDauw:
This was not needed when it was written. In fact, people for some reason added the static to code that did not have it >_>

PR #334, opened 2015-01-12
More descriptive Term/AliasGroup docs

thiemowmde:
* Main reason for this patch is the inline documentation. Note that the sub classes are changing the meaning of the `$languageCode` field in the parent class!
* ~~Inlined some trivial private setters. Some where already inlined, some not. This makes the code look more equal in the 4 classes.~~
* ~~Removed some `gettype`. They are not worth the additional complexity in my opinion.~~
* ~~"need" -> "must".~~
* ~~Avoid calling getters if not necesarry.~~
* ~~Avoid `is_null`, it's barely used in DataModel.~~

(Almost everything was re-submitted and merged in #335.)

### DIFF
--- a/src/Term/AliasGroup.php
+++ b/src/Term/AliasGroup.php
@@ -19,7 +19,7 @@ use InvalidArgumentException;
 class AliasGroup implements Comparable, Countable {
 
 	/**
-	 * @var string
+	 * @var string Usually the requested and/or actual language of the aliases.
 	 */
 	private $languageCode;
 
@@ -29,7 +29,7 @@ class AliasGroup implements Comparable, Countable {
 	private $aliases;
 
 	/**
-	 * @param string $languageCode
+	 * @param string $languageCode Usually the requested and/or actual language of the aliases.
 	 * @param string[] $aliases
 	 *
 	 * @throws InvalidArgumentException
--- a/src/Term/Term.php
+++ b/src/Term/Term.php
@@ -16,7 +16,7 @@ use InvalidArgumentException;
 class Term implements Comparable {
 
 	/**
-	 * @var string
+	 * @var string Usually the requested and/or actual language of the text.
 	 */
 	private $languageCode;
 
@@ -26,7 +26,7 @@ class Term implements Comparable {
 	private $text;
 
 	/**
-	 * @param string $languageCode
+	 * @param string $languageCode Usually the requested and/or actual language of the text.
 	 * @param string $text
 	 *
 	 * @throws InvalidArgumentException


src/Term/AliasGroup.php
```diff
@@ -19,7 +19,7 @@
 class AliasGroup implements Comparable, Countable {
 
 	/**
-	 * @var string
+	 * @var string Requested and actual language of the aliases.
```
JeroenDeDauw: "requested"? That seems to be about a usecase you have for the object, not something intrinsically part of the object.

thiemowmde: These two meanings do have separate fields in the sub class but are in a single field in this base class. I find it helpful to have this documented.

JeroenDeDauw: Ah, so you are putting the documentation of what the field means in the subclass here because that is where the field is defined? The issue with that is that you have the base class docs know about a subclass that it does not depend on. That is confusing for someone who is just interested in the base class. Because of this I think it's much better to document this in the derivative.

thiemowmde: > you are putting the documentation of what the field means in the subclass here

Yes and no. The sentence here is true with and without the subclass.
> document this in the derivative.

It is, or isn't it?

src/Term/AliasGroupFallback.php
```diff
@@ -17,40 +17,40 @@
...
 	private $actualLanguageCode;
+
+	/**
+	 * @var string|null Source language if the aliases are transliterations.
+	 */
 	private $sourceLanguageCode;
 
 	/**
-	 * @param string $languageCode
+	 * @param string $requestedLanguageCode Requested language, not necessarily the language of the
```
thiemowmde: > document this in the derivative.

Here it is.

src/Term/AliasGroup.php
JeroenDeDauw: > Yes and no. The sentence here is true with and without the subclass.

Try tasking the concept of an AliasGroup on it's own. Forget about all that is build on top of it. Where in this concept is a "request" present? This implies a request is being made, for which presumably the AliasGroup is the result. And while there definitely usecases in which that can happen, there are also those in which there is no such request. Hence "request" is not part of the AliasGorup concept. Please drop that part from the doc if you put it in this class.

Concrete example: suppose I deserialize an entity from the Wikidata API into a PHP Entity. That has a list of AliasGroup. Yet there is no notion of a language being requested there.

> actual language of the aliases

What does this "actual" mean exactly? "original" language? I'm not sure the above issue is absent here, since it seems to again depend on context.

Concrete example: what I have a place where I do not want to differentiate between original and fallback language, and want to construct an AliasGroup with a fallback language as language?

JeroenDeDauw: I just saw the derivative has an `$actualLanguageCode` field. So it seems "actual" is used for two different things then?

thiemowmde: > Where in this concept is a "request" present?

You are nitpicking about a word. I don't find this helpful. Sure, there are no requests in the DataModel. On the other hand the DataModel is useless without code that uses it. The documentation should say how it's supposed to be used. That's why the variables are called "languageCode", for example, even if there is nothing to ensure it's really a language code.

I could change the comment to something like "usually the requested and actual language code of the aliases".

> what I have a place where I do not want to differentiate between original and fallback language, and want to construct an AliasGroup with a fallback language as language?

If you don't need the additional information you should use the base class and store the fallback language (which is the actual language of the aliases then) in the $languageCode field. Judging from this comment I'm not sure if you fully understood what was going on in #272. That's exactly why I'm trying to make the documentation more clear.

JeroenDeDauw: > The documentation should say how it's supposed to be used.

Yeah. The problem here is that the documentation introduced here talks about a subset of all valid use cases. What about the concrete example I provided?

> You are nitpicking about a word.

That is your impression perhaps. I'm objecting against the meaning of the documentation.

JeroenDeDauw:
-2 to the doc change in AliasGroup and Term. +1 to the other changes

src/Term/AliasGroup.php
thiemowmde: > subset of all valid use cases.

What "subset"?

> What about the concrete example I provided?

What are you missing in my answer above?

thiemowmde: @JeroenDeDauw I don't understand your answer, see above. How to proceed here?

snaterlicious:
All in for better documentation. But that description is a rather misleading as it makes an assumption about how the AliasesGroup is used / from what context it is created. Capturing that is rather an objective of high-level documentation. Hence, I do not understand why the language should be regarded the "requested and/or actual language of the aliases" here. As far as I am concerned, it is just the language code of the language the aliases are in.

thiemowmde:
@snaterlicious, sorry, but that's exactly what the language code is **not**. That's why I'm trying to make the description more specific. And the fact that everybody gets it wrong tells me that my change is correct and should be merged.

JeroenDeDauw:
I'm still -2 to these doc changes, for the reasons originally stated.

thiemowmde:
Could one of you please, please suggest an alternative instead of just blocking stuff?

snaterlicious:
OK, please explain the nature of that language code then. I mean, the word "usually" is kind of odd in some documentation string without spending a word on what would be in an "unusual" case.
My impression is that both, Term and AliasGroup simply are containers... but since I already have been accused of being wrong, how should I make a suggestion if I do not know it better?

thiemowmde:
The way the fallback classes are implemented contains an OOP violation (maybe LSP, who knows). The subclasses are reusing the $languageCode field for different purpose than the base classes.

The field is meant to tell you the actual language of the text. **But** there is nothing this class can ever do to enforce this. Everybody is free to use the fields independently, no matter what some comments say. And that's exactly what the subclasses do.

It's not even guaranteed to be a language code. So what should the documentation say?
* The truth is what I suggest in #384.
* How the class is **meant** to be used is that the language should tell the actual language of the text. But that's misleading and even **wrong** if you take the subclasses into account.
* This leaves you with something in between, as proposed in this patch.

And since any documentation is better than none and nobody came up with a better suggestion I still want to see this merged.

snaterlicious:
OK, got the issue! Thanks for explaining.
Yes, technically, #384 is correct. However, one could argue about whom the documentation is directed to -- "users" of the class that just read the instantiated object's contents (which my perception was referring to, basically stating what *should* be) or "developers" writing code interacting with the class (which is what this pull request and #384 refers to). Not sure I want to argue about that as I can imagine quite some places in the code (at least in JS) where particular checks are kind of out of scope (language codes are a good example). Of course, the best situation would be addressing "users" and "developers" alike by having implemented straight-forward logic (validating the language code).
But following your logic, why is the variable named `$languageCode` then at all, why is `$aliases` not named `$texts`?
Anyway, back to my first impression the propsed documentation being a bit confusing. Just combine this pull request with #384. Result should be kind of:
`Language code of the language the aliases are in. (Might be a random string since the language code is not validated.)`
Still not very "user"-friendly but (1) better than no documentation and (2) correct (I hope).

thiemowmde:
@JeroenDeDauw is right when he says I'm trying to fix an issue that was introduced in the fallback subclasses.

PR #335, opened 2015-01-14
More descriptive Term/AliasGroupFallback docs

thiemowmde:
This is a re-submit of #334 since, apparently, having no documentation is better than having any.

Additional changes:
* Rearranged the setAliases code to avoid iterating the array twice.
* Inlined a few more trivial private setters.
* Dropped all `gettype` (introduced in #276) from the `Term` namespace since I find the added code complexity not worth it.

### DIFF
--- a/src/Term/AliasGroup.php
+++ b/src/Term/AliasGroup.php
@@ -35,36 +35,26 @@ class AliasGroup implements Comparable, Countable {
 	 * @throws InvalidArgumentException
 	 */
 	public function __construct( $languageCode, array $aliases = array() ) {
-		$this->setLanguageCode( $languageCode );
-		$this->setAliases( $aliases );
-	}
-
-	private function setLanguageCode( $languageCode ) {
 		if ( !is_string( $languageCode ) ) {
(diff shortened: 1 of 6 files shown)


src/Term/AliasGroup.php
```diff
@@ -35,36 +35,26 @@ class AliasGroup implements Comparable, Countable {
...
-			}
-		}
-
 		$this->aliases = array_values(
-			array_filter(
-				array_unique(
-					array_map(
-						'trim',
-						$aliases
+			array_unique(
```
JeroenDeDauw: Cool you got rid of the double iteration. This is kinda nested a lot now though. Perhaps a next step would be to extract part into it's own function.

thiemowmde: Doesn't matter much. The test cases are key here. A simple way of refactoring such code is to linearize it, ...
```php
$b = fn1( $a );
$c = fn2( $b );
$d = fn3( $c );
```
... but I don't think that's worth it in this case.

PR #338, opened 2015-01-14
Make all Item constructor parameters optional

JeroenDeDauw:
This is a follow up to https://github.com/wmde/WikibaseDataModel/pull/329


thiemowmde:
-2. This is an incomplete (test missing, now obsolete constructor calls untouched) fragment of #290.

PR #339, opened 2015-01-14
Allow constructing empty Fingerprint

JeroenDeDauw:
This is a split from #296, containing the non-controversial part.


thiemowmde:
-1 because of the missing documentation.

-100 because of the self-merge. This is not how a team should work. This is not how discussions are ended.

PR #342, opened 2015-01-16
Optional fingerprint constructor parameter in Property

thiemowmde:
Since the `Fingerprint` constructor parameter is now optional in `Item`, it should be optional in `Property` too.

This also adds tests for the `Property` constructor which was not tested at all, from what I can see.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -14,6 +14,7 @@
 * Empty strings are now detected as invalid language codes in the term classes
 * Made the `Item` constructor parameters optional
 * Made the `Fingerprint` constructor parameters optional
+* Made the `Property` constructor parameter for fingerprint nullable
 * Deprecated `Item::newEmpty` in favour of `new Item()`
 * The `StatementList` constructor now accepts `Statement` objects in variable-lenght argument list format
 
--- a/src/Entity/Property.php
+++ b/src/Entity/Property.php
@@ -38,18 +38,18 @@ class Property extends Entity implements StatementListProvider {
 	 * @since 1.0
 	 *
 	 * @param PropertyId|null $id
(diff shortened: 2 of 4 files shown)


RELEASE-NOTES.md
```diff
@@ -14,6 +14,7 @@
 * Empty strings are now detected as invalid language codes in the term classes
 * Made the `Item` constructor parameters optional
 * Made the `Fingerprint` constructor parameters optional
+* Made the `Property` constructor parameter for fingerprint nullable
```
JeroenDeDauw: Not very consistent

thiemowmde: Not sure what you mean. The property's type parameter can't be optional.

JeroenDeDauw: Wording in the rel notes

thiemowmde: I know you are talking about the release notes, but what's "not very consistent" here?

thiemowmde: @JeroenDeDauw: So instead of answering my question you did this https://github.com/wmde/WikibaseDataModel/commit/1a9577ff102137de52033c36efa423ccf079c35b, which is incorrect from my perspective, since the parameter is not "optional". You can not skip the parameter and  call `new Property( new PropertyId( 'P1' ), 'string' )`.

What now? I don't want to revert your change without an agreement.

JeroenDeDauw: Just fix it then. The README is documentation, and this never required RFCs to modify

thiemowmde: So you want me to revert to the wording you complained about? Sorry, but no. This is not a way forward. I still don't know what the problem was that made you changed it.

Issue #301, opened 2014-11-20
Replace Fingerprint with Label/Description/Alias interfaces


thiemowmde:
After working with the so called "Fingerprint" for a while now I never really understood and still wonder why we introduced this in the first place. (And I know other team members feel the same way.) What it basically did was moving some methods deeper. I'm not sure what actual problem that solved (that could not been solved in an other, less disruptive way) but I know it introduced quite a few new ones (like not being able to use `clone` on entity objects any more). I suggest to drop it and replace it with the following interfaces in DataModel 3.0.0:

``` php
interface LabeledDocument {
    public function getLabels();
    public function setLabels( /* TBD */ );
    public function hasLabel( $languageCode );
    public function getLabel( $languageCode );
    public function setLabel( $languageCode, $text );
    public function removeLabel( $languageCode );
}
interface DescribedDocument {
    public function getDescriptions();
    public function setDescriptions( /* TBD */ );
    public function hasDescription( $languageCode );
    public function getDescription( $languageCode );
    public function setDescription( $languageCode, $text );
    public function removeDescription( $languageCode );
}
interface AliasedDocument {
    /* TBD */
}
```

Also see [how the Java DataModel does this](https://github.com/Wikidata/Wikidata-Toolkit/blob/master/wdtk-datamodel/src/main/java/org/wikidata/wdtk/datamodel/interfaces/TermedDocument.java).

PR #343, opened 2015-01-19
Rename Fingerprint to EntityTerms

thiemowmde:
This does __not__ make #301 (now https://phabricator.wikimedia.org/T78286) obsolete, but is a first step.

https://phabricator.wikimedia.org/T87237


JeroenDeDauw:
This rename causes a pile of work - we need to update a lot of code using the class and the methods in most of the PHP components that depend on DataModel. Plus rename things such as fingerprint view, causing further breaking changes, including in our JS components. No reasoning has been provided for why this rename should happen, and I certainly don't see anything that justifies the cost. so -2 from my side.

thiemowmde:
> including in our JS components.

Already renamed.

> No reasoning has been provided for why this rename should happen

This is just not true. We talked about this a gazillion times.

addshore:
As the class itself isn't changing structure could we leave a class alias and also keep getFingerprint for at least one release? Thus not immediately breaking everything that uses these bits of code ?

thiemowmde:
Since this is the first time (I'm aware of) we working on a branch, why is this not mergeable any more and how do I rebase this?

thiemowmde:
Erm, why? We did not agreed on a way forward, this is still a possibility. Reopen please.

JeroenDeDauw:
GitHub automatically closed this when I removed the 3.0.x-dev branch. Needs a new PR against master.

PR #345, opened 2015-01-20
Disallow empty strings in SiteLink

thiemowmde:
This also removes some of the `gettype` calls as reasoned in #335.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,9 @@
 # Wikibase DataModel release notes
 
+## Version 3.0.0 (dev)
+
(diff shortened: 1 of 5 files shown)


RELEASE-NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Wikibase DataModel release notes
 
+## Version 3.0.0 (dev)
```
JeroenDeDauw: Why not 2.6?

JeroenDeDauw:
Missing tests

RELEASE-NOTES.md
thiemowmde: No special reason. I will change it.

PR #346, opened 2015-01-20
Inline trivial methods in DispatchingEntityIdParser

thiemowmde:
~~A parser with an empty array can't do anything but throw an exception. Better throw this exception as early as possible.~~

* Added special message if the builders array is empty.
* Re-throw the original message if one of the EntityId constructors throws an InvalidArgumentException, which should not happen since they aren't (currently) doing anything but checking against the same regex again.
* Inlined trivial methods.
* Added a test.

### DIFF
--- a/src/Entity/DispatchingEntityIdParser.php
+++ b/src/Entity/DispatchingEntityIdParser.php
@@ -9,6 +9,7 @@ use InvalidArgumentException;
  *
  * @licence GNU GPL v2+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ * @author Thiemo Mättig
  */
 class DispatchingEntityIdParser implements EntityIdParser {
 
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
> Better throw this exception as early as possible.

True, in case the exception would always be thrown. That is not a given though. One might end up constructing a parser without ever using it.

A second problems comes from the change in when it is thrown. Imagine you configure the supported entity types somewhere, and set this to no supported entities. You'll then end up with no associated parsers. And after this change you'll get the error here. While you likely either want to not get the error during application config, and get it when an ID is parsed, or get it at the place where the config is wrong, which will as far as I can tell always be before you reach this point.

thiemowmde:
I find it quite confusing if other peoples patches are closed with no reason given. :-(

Benestar:
Sorry, I thought this to be a dead pull request which was open in error as the subject of this patch already got adressed in other patches. This only removes some methods and makes a method more complicated, where I don't see any reason to merge these functions.

Again sorry for closing this without asking first, I really thought this was out of interest.

Issue #310, opened 2014-11-25
Add setId to EntityDocument


JeroenDeDauw:
`Wikibase\Repo\Store\WikiPageEntityStore` has 

```php
public function assignFreshId( Entity $entity ) {
	if ( $entity->getId() !== null ) {
		throw new StorageException( 'This entity already has an ID!' );
	}

	$contentModelId = $this->contentFactory->getContentModelForType( $entity->getType() );
	$id = $this->idGenerator->getNewEntityId( $contentModelId );

	$entity->setId( $id );
}
```

This cannot be changed to use `EntityDocument` do to the call to `setId`. This suggests adding a `setId` method to `EntityDocument`.

JeroenDeDauw:
As per discussion here and on #312, it seems preference goes to doing what was originally suggested: adding a `setId` method to `EntityDocument`.

PR #348, opened 2015-01-22
Add setId to EntityDocument

JeroenDeDauw:
Pull request into 3.0.x-dev, not into master.

This fixes https://github.com/wmde/WikibaseDataModel/issues/310 and is the approach that came out of the discussion there. (We decided to not go with https://github.com/wmde/WikibaseDataModel/pull/312)

Note how we cannot use `EntityId` as PHP type hint, since then derivatives would not be able to do things such as accept an integer. This is needed since we support `$item->setId( 42 )`.

Also note that if we would just (doc) type hint against `EntityId`, we'd have a LSP violation. This is why the doc outlines a more detailed contract. (I'd rather avoid this situation here, since it can be avoided by not having the method, which is part of why I favoured the approach of deprecation. We have to do something though, and this approach seems benign enough to me.)

Both these points apply to the existing `setId` method in `Entity`, so in a way, we are just keeping what is already there.

### DIFF
--- a/src/Entity/EntityDocument.php
+++ b/src/Entity/EntityDocument.php
@@ -6,7 +6,7 @@ namespace Wikibase\DataModel\Entity;
  * Minimal interface for all objects that represent an entity.
  * All entities have an EntityId and an entity type.
  *
- * @since 0.8.2
+ * @since 0.8.2, modified in 3.0
  *
  * @licence GNU GPL v2+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
@@ -30,5 +30,18 @@ interface EntityDocument {
 	 * @return string
 	 */
 	public function getType();
+
+	/**
+	 * Sets the id of the entity. If the id is not of the correct type,
+	 * an exception will be thrown. A specific derivative of EntityId is
+	 * always supported.
+	 *
+	 * @since 3.0
+	 *
+	 * @param EntityId $id
+	 *
+	 * @throws \InvalidArgumentException
+	 */
+	public function setId( $id );
 
 }
--- a/src/Entity/Entity.php
+++ b/src/Entity/Entity.php
@@ -49,8 +49,6 @@ abstract class Entity implements \Comparable, FingerprintProvider, EntityDocumen
 		return $this->id;
 	}
 
-	public abstract function setId( $id );
-
 	/**
 	 * Sets the value for the label in a certain value.
 	 *
--- a/tests/fixtures/EntityOfUnknownType.php
+++ b/tests/fixtures/EntityOfUnknownType.php
@@ -18,4 +18,7 @@ class EntityOfUnknownType implements EntityDocument {
 		return 'unknown-entity-type';
 	}
 
+	public function setId( $id ) {
+	}
+
 }


src/Entity/EntityDocument.php
```diff
@@ -31,4 +33,15 @@ public function getId();
 	 */
 	public function getType();
 
+	/**
+	 * Sets the id of the entity. If the id is not of the correct type,
+	 * an exception will be thrown. A specific derivative of EntityId is
+	 * always supported.
+	 *
+	 * @param EntityId|mixed $id
```
thiemowmde: I suggest something like `mixed|EntityId|int|null`, so it can be `ItemId|int|null` in the subclasses without violating LSP.

JeroenDeDauw: Doing that will actually cause LSP violations. If you have an EntityId type that is not numeric, it can't construct itself from an integer, so this method won't be able to accept it.

thiemowmde: I think because `mixed` is in the list, no addition can be a LSP violation.

JeroenDeDauw: That depends on how you interpret `mixed`. If you interpret it as "must support all types", then yes.

I'm interpreting this one as "and possibly something else". I started with just having `EntityId` as type hint, though then you'd get IDE warnings... ... no wait, you only will when not using the subtype, in which case you're doing it wrong... so yeah, should just be dropped then.

thiemowmde: There is no interpretation of what `mixed` means, it means "all types". So whatever you do, no matter if you start with `EntityId` and restrict it to a subtype or if you start with `mixed` and restrict it to `ItemId|int|null`, it will always be a LSP violation. Which makes me wonder. Is there even a "correct" way to solve this?

src/Entity/EntityDocument.php
```diff
@@ -31,4 +31,17 @@ public function getId();
...
 	public function getType();
 
+	/**
+	 * Sets the id of the entity. If the id is not of the correct type,
+	 * an exception will be thrown. A specific derivative of EntityId is
+	 * always supported.
+	 *
+	 * @since 3.0
+	 *
+	 * @param EntityId $id
```
thiemowmde: At least `null` is missing now.

src/Entity/EntityDocument.php
```diff
@@ -31,4 +31,17 @@ public function getId();
...
+	 *
+	 * @throws \InvalidArgumentException
```
thiemowmde: Why did you reverted my change here?

JeroenDeDauw: I merely did not re-apply it since I don't see reason to. Feel free to change it if you want though

src/Entity/EntityDocument.php
JeroenDeDauw: Is there an issue here that you think needs to be solved before this is merged?

src/Entity/EntityDocument.php
JeroenDeDauw: That was intentionally left out. Do we need to support this?

PR #350, opened 2015-01-26
[WIP] Throw exception on creation of nonsensical Fallbacks

JanZerebecki:
Such Fallbacks should be normal Terms instead.

### DIFF
--- a/src/Term/TermFallback.php
+++ b/src/Term/TermFallback.php
@@ -28,8 +28,9 @@ class TermFallback extends Term {
 	 * @param string $requestedLanguageCode Requested language, not necessarily the language of the
 	 * text.
 	 * @param string $text
-	 * @param string $actualLanguageCode Actual language of the text.
-	 * @param string|null $sourceLanguageCode Source language if the text is a transliteration.
+	 * @param string $actualLanguageCode Actual language of the text which is the fall back.
+	 * @param string|null $sourceLanguageCode Source language if the text is a transliteration,
+	 *		null if no transformation was done.
 	 *
 	 * @throws InvalidArgumentException
 	 */
@@ -44,6 +45,28 @@ class TermFallback extends Term {
 			|| ( is_string( $sourceLanguageCode ) && $sourceLanguageCode !== '' )
 		) ) {
 			throw new InvalidArgumentException( '$sourceLanguageCode must be a non-empty string or null' );
+		}
+
+		if ( $sourceLanguageCode === null && $actualLanguageCode === $requestedLanguageCode ) {
+			throw new InvalidArgumentException(
+				'$actualLanguageCode and $requestedLanguageCode must be different when $sourceLanguageCode is null, '.
+					'use Term without Fallback when no fall back did occur'
+			);
+		}
+
+		if ( $actualLanguageCode === $sourceLanguageCode ) {
+			throw new InvalidArgumentException(
+				'$actualLanguageCode and $sourceLanguageCode must be different, '.
+					'set $sourceLanguageCode to null when no transformation took place'
+			);
+		}
+
+		if ( $requestedLanguageCode === $sourceLanguageCode ) {
+			throw new InvalidArgumentException(
+				'$requestedLanguageCode and $sourceLanguageCode must be different; '.
+					'Transforming one language to itself makes no sense; Each variant, '.
+					'that can be transformed or transliterated to, needs a different language code'
+			);
 		}
 
 		$this->actualLanguageCode = $actualLanguageCode;
--- a/tests/unit/Term/TermFallbackTest.php
+++ b/tests/unit/Term/TermFallbackTest.php
@@ -27,34 +27,29 @@ class TermFallbackTest extends \PHPUnit_Framework_TestCase {
 		$this->assertNull( $term->getSourceLanguageCode() );
 	}
 
-	/**
-	 * @dataProvider invalidLanguageCodeProvider
-	 * @expectedException InvalidArgumentException
-	 */
-	public function testGivenInvalidActualLanguageCode_constructorThrowsException( $languageCode ) {
-		new TermFallback( 'foor', 'bar', $languageCode, 'foos' );
-	}
-
-	public function invalidLanguageCodeProvider() {
-		return array(
-			array( null ),
-			array( 21 ),
-			array( '' ),
-		);
+	public function testConstructorWithActualSameAsRequestedLanguage() {
+		new TermFallback( 'foor', 'bar', 'foor', 'foos' );
 	}
 
 	/**
-	 * @dataProvider invalidSourceLanguageCodeProvider
+	 * @dataProvider invalidArgumentCombinationsProvider
 	 * @expectedException InvalidArgumentException
 	 */
-	public function testGivenInvalidSourceLanguageCode_constructorThrowsException( $languageCode ) {
-		new TermFallback( 'foor', 'bar', 'fooa', $languageCode );
+	public function testGivenInvalidArgumentCombinations_constructorThrowsException( $requestedLanguage, $sourceLanguage, $actualLanguage ) {
+		new TermFallback( $requestedLanguage, 'bar', $sourceLanguage, $actualLanguage );
 	}
 
-	public function invalidSourceLanguageCodeProvider() {
+	public function invalidArgumentCombinationsProvider() {
 		return array(
-			array( 21 ),
-			array( '' ),
+			'nullSource' => array( 'foor', null, 'fooa' ),
+			'integerSource' => array( 'foor', 21, 'fooa' ),
+			'integerActual' => array( 'foor', 'foos', 21 ),
+			'emptyStringSource' => array( 'foor', '', 'fooa' ),
+			'emptyStringActual' => array( 'foor', 'fooa', '' ),
+			'actualSameAsSource' => array( 'foor', 'foos', 'foos' ),
+			'actualSameAsRequested' => array( 'foor', 'foos', 'foor' ),
+			'requestedSameAsSourceAndNoActual' => array( 'foor', 'foor', null ),
+			'requestedSameAsSourceAndActual' => array( 'foor', 'foor', 'foor' ),
 		);
 	}
 
@@ -81,7 +76,7 @@ class TermFallbackTest extends \PHPUnit_Framework_TestCase {
 			'actualLanguage' => array( new TermFallback( 'foor', 'bar', 'spam', 'foos' ) ),
 			'sourceLanguage' => array( new TermFallback( 'foor', 'bar', 'fooa', 'spam' ) ),
 			'null sourceLanguage' => array( new TermFallback( 'foor', 'bar', 'fooa', null ) ),
-			'all' => array( new TermFallback( 'ham', 'nom', 'nom', 'nom' ) ),
+			'all' => array( new TermFallback( 'ham', 'nom1', 'nom2', 'nom3' ) ),
 			'instance Term' => array( new Term( 'foor', 'bar' ) ),
 		);
 	}
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,8 +1,10 @@
 # Wikibase DataModel release notes
 
-## Version 2.6.0 (dev)
+## Version 3.0.0 (dev)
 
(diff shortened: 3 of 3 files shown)


src/Term/TermFallback.php
```diff
@@ -46,6 +46,18 @@ public function __construct( $requestedLanguageCode, $text, $actualLanguageCode,
 			throw new InvalidArgumentException( '$sourceLanguageCode must be a non-empty string or null' );
 		}
 
+		if ( $actualLanguageCode === $sourceLanguageCode ) {
```
thiemowmde: I'm not sure if this is a good idea. Why shouldn't a user of this class be allowed to use it like that if he wants?

Alternative proposal: Check if `source` is identical to `actual`, and if it is, set `source` to `null`.

src/Term/TermFallback.php
```diff
@@ -46,6 +46,18 @@ public function __construct( $requestedLanguageCode, $text, $actualLanguageCode,
...
+			throw new InvalidArgumentException(
+				'$actualLanguageCode and $sourceLanguageCode must be different'
+			);
+		}
+
+		if ( $sourceLanguageCode === null && $actualLanguageCode !== $requestedLanguageCode ) {
```
thiemowmde: I think this is wrong. If "requested" is, for example, `de-at`, "actual" can be `de` with no "source" (no "source" that's different from "actual").

src/Term/TermFallback.php
JanZerebecki: Why should the user of this class be deprived of the notification that they did something wrong in favor of silently doing something that maybe was not intended?

src/Term/TermFallback.php
JanZerebecki: Thanks, fixed to actually check what the exception message says.

src/Term/TermFallback.php
thiemowmde: The question is, "wrong" in which context? This is probably moving application logic into a base component.

thiemowmde:
Don't forget the tests.

src/Term/TermFallback.php
JanZerebecki: Wrong in the context of intended API usage.

src/Term/TermFallback.php
```diff
@@ -46,6 +46,24 @@ public function __construct( $requestedLanguageCode, $text, $actualLanguageCode,
...
+				'$actualLanguageCode and $sourceLanguageCode must be different'
+			);
+		}
+
+		if ( $requestedLanguageCode === $sourceLanguageCode ) {
+			throw new InvalidArgumentException(
+				'$requestedLanguageCode and $sourceLanguageCode must be different'
+			);
+		}
+
```
adrianheine: What about `$actualLanguageCode === null`. Can happen, too, apparently.

src/Term/TermFallback.php
thiemowmde: Do you understand that I think this could be a problem? Why should knowledge about the API be baked into the DataModel? @JeroenDeDauw recently blocked me for exactly that reason in #334.

thiemowmde:
-1. I don't think this code belongs here, at least not in the way it currently is. The exceptions don't explain anything. Why should it be forbidden to use the classes in that way? The messages don't give any reason or alternative.

src/Term/TermFallback.php
adrianheine: But the other way round, why should anybody use this? What does `new TermFallback( 'en', 'text', 'en' )` express?

src/Term/TermFallback.php
JanZerebecki: No even before this change the is_string check should prevent that.

src/Term/TermFallback.php
thiemowmde: I suggest to implement well tested `isFallback` and `isTransliteration` (and maybe `isVariant`) methods instead of throwing exceptions. In the given example all these methods will return `false`. So yes, this means you can have a TermFallback object that is *not* a fallback. I find this much more convenient: Code that constructs Term objects does not need complicated `if` constructs, it can be straight and create TermFallback objects for all terms, no matter how the given fallback actually resolves for each of them.

JanZerebecki:
I added explanations and suggestions what would be correct.

src/Term/TermFallback.php
```diff
@@ -28,8 +28,8 @@ class TermFallback extends Term {
 	 * @param string $requestedLanguageCode Requested language, not necessarily the language of the
 	 * text.
 	 * @param string $text
-	 * @param string $actualLanguageCode Actual language of the text.
-	 * @param string|null $sourceLanguageCode Source language if the text is a transliteration.
+	 * @param string $actualLanguageCode Actual language of the text which was used as a fall back.
```
thiemowmde: Huh? Why "was"? What does "used" mean?

thiemowmde:
I suggest to implement well tested `isFallback` and `isTransliteration` (and maybe `isVariant`) methods instead of throwing exceptions. In the given example all these methods will return `false`. So yes, this means you can have a TermFallback object that is *not* a fallback. I find this much more convenient: Code that constructs Term objects does not need complicated `if` constructs, it can be straight and create TermFallback objects for all terms, no matter how the given fallback actually resolves for each of them.

adrianheine:
@thiemowmde wrote:
> I suggest to implement well tested isFallback and isTransliteration (and maybe isVariant) methods instead of throwing exceptions. In the given example all these methods will return false. So yes, this means you can have a TermFallback object that is not a fallback. I find this much more convenient: Code that constructs Term objects does not need complicated if constructs, it can be straight and create TermFallback objects for all terms, no matter how the given fallback actually resolves for each of them.

I suggest having a `TermFactory` which constructs the correct `Term` depending on whether it is a fallback or not. That way, the class used actually communicates something to the users.

JanZerebecki:
A potential $term->isFallback() would after this just be ( $term instanceof TermFallback ).
A potential $term->isTransliteration() would just be ( $term->getSourceLanguageCode() !== null ).
AFAIK we never indiscriminately fall back, but have code to decide if you need to invoke the fall back logic, so I don't think a TermFactory would be used, but if there is a need for it, I would prefer that over TermFallback being able to have isFallback() === false. Same with transliteration.

src/Term/TermFallback.php
JanZerebecki: Changed, is that better?

thiemowmde:
My problem is, I think code that wants to work with these classes will need to repeat the logic in this patch. For each term the code needs to check what the language fallback actually did: Was a term in the requested language found or not? If not, is the actual term the result of a transliteration or not? Depending on all these questions an other combination of a specific class and specific constructor parameters needs to be called.

I don't think it makes sense to simply `try` instantiating a fallback object and if it fails, create a non-fallback term object instead. How should the calling code know what the exception actually meant and if it's ok to drop the information that the term may be the result of a language fallback?

Factory code (wherever it lives) needs either specific exceptions for the relevant cases (where it's supposed to create a `Term` object instead of a `TermFallback` object) or the proposed `is...` methods.

A factory that just repeats the logic proposed in this patch doesn't make much sense in my opinion. Or the other way around: Bake that logic into a factory, not into the DataModel classes.

src/Term/TermFallback.php
```diff
@@ -46,6 +47,28 @@ public function __construct( $requestedLanguageCode, $text, $actualLanguageCode,
...
+				'$actualLanguageCode and $sourceLanguageCode must be different, '.
+					'set $sourceLanguageCode to null when no transformation took place'
+			);
+		}
+
+		if ( $requestedLanguageCode === $sourceLanguageCode ) {
+			throw new InvalidArgumentException(
+				'$requestedLanguageCode and $sourceLanguageCode must be different. '.
+					'Transforming one language to itself makes no sense. Each variant, '.
+					'that can be transformed or transliterated to, needs a different language code.'
```
thiemowmde: Inconsistent punctuation, better remove everywhere.

src/Term/TermFallback.php
```diff
@@ -46,6 +47,28 @@ public function __construct( $requestedLanguageCode, $text, $actualLanguageCode,
 			throw new InvalidArgumentException( '$sourceLanguageCode must be a non-empty string or null' );
 		}
 
+		if ( $sourceLanguageCode === null && $actualLanguageCode === $requestedLanguageCode ) {
```
thiemowmde: One of the later two `if`s is redundant and can be removed. What about these checks and messages:
1. "$actual must be different from $requested, if no transliteration took place; use Term without fallback instead"
2. "$source must either be null, if no transliteration took place, or different from $requested"

src/Term/TermFallback.php
JanZerebecki: Done.

src/Term/TermFallback.php
JanZerebecki: No, 3 variables  means 3 combinations of how 2 can be equal. The comparison with null is && not ||.

JanZerebecki:
It seems the only user in Wikibase.git currently uses TermFallback also when no fall back is necessary, see https://github.com/wikimedia/mediawiki-extensions-Wikibase/blob/2cf9ea683c3542ea32ac432fc9ccb23acbc1a77f/lib/includes/store/LanguageFallbackLabelLookup.php#L57 .

src/Term/TermFallback.php
thiemowmde: As I said, what's relevant here are the tests. If they cover everything, anybody can safely remove/refactor these checks.

tests/unit/Term/TermFallbackTest.php
```diff
@@ -53,11 +54,20 @@ public function testGivenInvalidSourceLanguageCode_constructorThrowsException( $
...
+			'integer' => array( 21 ),
+			'emptyString' => array( '' ),
+			'sameAsRequestedLanguageCode' => array( 'foor' ),
 		);
 	}
 
+	/**
+	 * @expectedException InvalidArgumentException
+	 */
+	public function testGivenInvalidRequestedSameAsActualAndNoTransliteration_constructorThrowsException() {
```
thiemowmde: I suggest to write individual tests like this one, either alternatively or in addition to the test cases in the providers. I come up with the following relevant combinations:
* `( 'requested', 'text', 'actual', 'source' )` => valid
* `( 'requested', 'text', 'actual', null )` => valid
* `( 'requested', 'text', 'actual', 'actual' )` => exception
* `( 'requested', 'text', 'actual', 'requested' )` => exception
* `( 'requested', 'text', 'requested', 'source' )` => exception
* `( 'requested', 'text', 'requested', null )` => exception
* `( 'requested', 'text', 'requested', 'requested' )` => exception

tests/unit/Term/TermFallbackTest.php
```diff
@@ -53,11 +54,20 @@ public function testGivenInvalidSourceLanguageCode_constructorThrowsException( $
...
-			array( '' ),
+			'integer' => array( 21 ),
+			'emptyString' => array( '' ),
+			'sameAsRequestedLanguageCode' => array( 'foor' ),
 		);
 	}
 
+	/**
+	 * @expectedException InvalidArgumentException
+	 */
```
JeroenDeDauw: Looks like the `Invalid` should be dropped

thiemowmde: Agree, I find the wording (of the test method name) hard to read.

tests/unit/Term/TermFallbackTest.php
JanZerebecki: > ( 'requested', 'text', 'requested', 'source' ) => exception

This is valid, added an explicit test for it.

> ( 'requested', 'text', 'requested', 'requested' ) => exception

I still need to add this and I should probably move all the exception throwing test variants into one data provider. The rest already have tests.

tests/unit/Term/TermFallbackTest.php
JanZerebecki: Done.

tests/unit/Term/TermFallbackTest.php
JanZerebecki: Done.

thiemowmde:
Still [WIP]? Or ready for a final review? Please rebase and remove the [WIP] if you think this is the case. Thanks.

PR #352, opened 2015-01-28
Remove misleading type from EntityDocument

thiemowmde:
Main change here is: Remove the `@param` line. Note that the original `Entity` class did not had that either. It's misleading (subclasses don't fulfill this contract and aren't supposed to) and better described in text.

### DIFF
--- a/src/Entity/EntityDocument.php
+++ b/src/Entity/EntityDocument.php
@@ -8,7 +8,7 @@ use InvalidArgumentException;
  * Minimal interface for all objects that represent an entity.
  * All entities have an EntityId and an entity type.
  *
- * @since 0.8.2
+ * @since 0.8.2, modified in 3.0
  *
  * @licence GNU GPL v2+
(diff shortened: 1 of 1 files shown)


src/Entity/EntityDocument.php
```diff
@@ -2,11 +2,13 @@
...
 namespace Wikibase\DataModel\Entity;
 
+use InvalidArgumentException;
+
 /**
  * Minimal interface for all objects that represent an entity.
  * All entities have an EntityId and an entity type.
  *
- * @since 0.8.2, modified in 3.0
+ * @since 0.8.2
```
JeroenDeDauw: why are you removing this?

thiemowmde: I think this does not belong here. This line should say since when a class exists. If a method signature changed, the since tag of this method should change, not the since tag of the class.

Feel free to click edit and revert my change in this line. It's unrelated to the rest of this patch anyway.

PR #354, opened 2015-02-03
[RFC] Introduce generic Item + Property differ

thiemowmde:
This patch introduces a generic differ that allows comparing items and properties. Note that both do have statements now. The only remaining difference is that properties don't have site links.

If you don't like the name, please suggest a better one. Maybe `ItemPropertyDiffer` is better?

This is an RFC for the moment. Tests and release notes are missing.

### DIFF
--- a/src/Entity/Diff/EntityDiffer.php
+++ b/src/Entity/Diff/EntityDiffer.php
@@ -37,15 +37,12 @@ class EntityDiffer {
 	 * @throws RuntimeException
 	 */
 	public function diffEntities( EntityDocument $from, EntityDocument $to ) {
-		$this->assertTypesMatch( $from, $to );
-
-		return $this->getDiffStrategy( $from->getType() )->diffEntities( $from, $to );
-	}
-
-	private function assertTypesMatch( EntityDocument $from, EntityDocument $to ) {
-		if ( $from->getType() !== $to->getType() ) {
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
This class pretends to be able to diff all types of entities, which is true if you just have items and properties, but not if an extension defines a new type of entity.

What problem are you addressing with this change?

thiemowmde:
> What problem are you addressing with this change?

Being able to diff different types of entities that share common aspects (namely FingerprintProvider and StatementListProvider).

> pretends to be able to diff all types of entities

It tries to do exactly that but will miss aspects it doesn't know about, sure. Please suggest a better name or other change that will make this acceptable. What about some kind of `instanceof Item || instanceof Property` check?

JeroenDeDauw:
There already is code to diff entities, so I still don't understand what this change gains us (apart from the OCP violation).

thiemowmde:
To what code are you referring to?

thiemowmde:
Also, I don't see how **this** is introducing a violation that wasn't there before. The code I touched was already build around the assumption that there are no other entity types but items and properties. This is not changed.

thiemowmde:
Dropped the interface.

Still needs tests.

thiemowmde:
Resubmitted as https://github.com/wmde/WikibaseDataModelServices/pull/61.

PR #356, opened 2015-02-05
Fix HashArray accepting anything

thiemowmde:
Calling `new SnakList( $snak )` with a single snak object created an empty snak list. I added a specific test and immediately found that mistake in StatementTest. Whoops.

### DIFF
--- a/src/HashArray.php
+++ b/src/HashArray.php
@@ -70,11 +70,17 @@ abstract class HashArray extends \ArrayObject implements \Hashable, \Comparable 
 	 * @param array|Traversable|null $input
 	 * @param int $flags
 	 * @param string $iterator_class
+	 *
+	 * @throws InvalidArgumentException
 	 */
 	public function __construct( $input = null, $flags = 0, $iterator_class = 'ArrayIterator' ) {
 		parent::__construct( array(), $flags, $iterator_class );
 
 		if ( $input !== null ) {
+			if ( !is_array( $input ) && !( $input instanceof Traversable ) ) {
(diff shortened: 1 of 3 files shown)


snaterlicious:
Very nice. The only thing that bugs me is that those generic negative test cases are added to the SnakListTest. SnakList seems to be the only subclass of HashArray and I know HashArray being abstract, but is there no simple way to mock a HashArray subclass and use that for generic testing?

thiemowmde:
"Generic"? I actually find it much, much better that way. This means we can possibly inline the base class in the future without loosing any tests. Abstract classes can't and shouldn't be tested, in my opinion (we had this discussion with @Benestar at some point). The tests should be concrete, in every subclass.

JeroenDeDauw:
Good to not mirror the code reuse via inheritance. That'd just make it harder to change the implementation.

snaterlicious:
Alright, then, assuming derivation of another class from HashArray, that class would need to test the HashArray base functionality again -- with some test cases being exactly the same. The state now is kind of leaving a door open having the paper blown through the air. ;)

JeroenDeDauw:
If you write a new list class, you'll need to test the public behavior of this list class. This holds true if it uses HashArray or not, which ultimately is an implementation detail the test should not bind itself to. If you have common behavior to test for list classes, you can always use composition in your tests where appropriate.

We also do not need to agree on that. There is no planned new derivative of HashArray, and if we end up needing one, and it indeed runs into the raised concerns, fixing it is not going to be hard.

PR #362, opened 2015-02-11
Drop unused EntityTest::makeClaims method

thiemowmde:
Same as #361, but in the master branch.


JeroenDeDauw:
I suggest we close this to avoid merge hassle, and let the unused code be there until we merge 3.0.x into master

JeroenDeDauw:
I'll merge if someone volunteers to do the merge hassle (which I'm actually not 100% will be there)

PR #367, opened 2015-02-12
Favor StatementList's argument list constructor

thiemowmde:
Supported since #289.


JeroenDeDauw:
Why are you doing this on the 3.x branch? [1] If you start submitting changes like this here, and we have similar ones on master, we'll end up with a big evil merge of doom.

[1] not a criticism, we should however make sure this is coordinated properly

JeroenDeDauw:
Did not see this was also submitted against master in #368. I'm thinking it's probably easier if we merge master into 3.0.x.

Issue #372, opened 2015-02-12
SnakList construction with argument list


thiemowmde:
Make it possible to use `new SnakList( $snak1, $snak2 )`.

Issue #371, opened 2015-02-12
Type safe ReferenceList constructor


thiemowmde:
Currently this `SplObjectStorage` based thing accepts anything and is used wrong even in DataModel tests.

JeroenDeDauw:
This has overlap with #212

PR #376, opened 2015-02-12
[WIP] Rework ...List constructors

### LINKED ISSUE #370: ReferenceList construction with argument list
Make it possible to use `new ReferenceList( $reference1, $reference2 )` without `array()`.

thiemowmde:
This changes all three
* `Reference`
* `ReferenceList` and
* `SnakList`

classes to

1. not accept `null`,
2. reject mismatching elements and
3. accept variable constructor arguments.

This fixes #370, #371 and #372.

Note how the ReferenceList constructor was used wrong in StatementTest.

WIP because some of these new features are missing tests. And I should probably create three pull requests for the three classes.

### DIFF
--- a/src/Reference.php
+++ b/src/Reference.php
@@ -26,22 +26,25 @@ class Reference implements \Hashable, \Comparable, \Immutable, \Countable {
 	/**
 	 * An array of Snak is only supported since version 1.1.
 	 *
-	 * @param Snaks|Snak[]|null $snaks
+	 * @param Snak[]|Snaks|Snak $snaks
+	 * @param Snak [$snak2,...]
+	 *
 	 * @throws InvalidArgumentException
 	 */
-	public function __construct( $snaks = null ) {
-		if ( $snaks === null ) {
-			$this->snaks = new SnakList();
+	public function __construct( $snaks = array() /*...*/ ) {
+		if ( $snaks instanceof Snak ) {
+			$snaks = func_get_args();
 		}
-		elseif ( $snaks instanceof Snaks ) {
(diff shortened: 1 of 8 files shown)


JeroenDeDauw:
Why is this being done on the 3.0.x branch?

thiemowmde:
Because I expect this to become a big breaking change that can't go into the 2.x branch. It#s WIP anyway, I will split this and try to submit into the 2.x branch if possible.

snaterlicious:
:+1: Just a question: Do we really need the constructors to be variadic? And, to me, testing the generic HashableObjectStorage mechanisms in the ReferenceList test seems a bit odd -- same as on #356.

JeroenDeDauw:
> Do we really need the constructors to be variadic?

I don't think so, but if we don't, Thiemo will has a sad. Do you think it's harmful to make them variadic as well?

snaterlicious:
Every additional functionality is prone to errors, needs to be tested, needs to be documented, needs to be maintained. Talking about "cost", you know. ;)
OK, maybe that is not a major functionality, but it feels like implementing functionality just because we can... :-/

JeroenDeDauw:
Yes, the cost of this feature is quite clear:

* initial implementation
* maintenance
* users are presented with two ways of using the constructor

All in all, these do not add up to much considering how small this is. Hence I do not think it is harmful, though I also think that this is doing something we do not really benefit from much, while there are many pressing issues to be worked upon.

thiemowmde:
Wh... hey, I'm still working on this!

thiemowmde:
Resubmitted as #416.

PR #380, opened 2015-02-12
Merge master into 3.0.x

JeroenDeDauw:


src/Entity/Item.php
```diff
@@ -208,7 +208,7 @@ public function getType() {
 	 * @return Statement
 	 */
 	public function newClaim( Snak $mainSnak ) {
-		return new Statement( new Claim( $mainSnak ) );
```
thiemowmde: If this is a pull request that merges master into 3.x, how is it possible these changes pop up here? Something is still messed up, I'm afraid.

JeroenDeDauw:
The derp is strong in this one

#382

PR #386, opened 2015-02-20
More specific exception messages in DispatchingEntityIdParser

thiemowmde:
This patch does not change semantics, but:
* Inline trivial methods.
* Re-throw the original message if one of the builders fails, which should not happen because all they (currently) do is checking against the same regex.
* Reword the final message.

This is split from #346 to make it much more focused.

### DIFF
--- a/src/Entity/DispatchingEntityIdParser.php
+++ b/src/Entity/DispatchingEntityIdParser.php
@@ -9,6 +9,7 @@ use InvalidArgumentException;
  *
  * @licence GNU GPL v2+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ * @author Thiemo Mättig
  */
 class DispatchingEntityIdParser implements EntityIdParser {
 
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
I don't agree with this inlining.

Is there an actual problem being solved here? My guess is that this supposedly improves performance, though I'm kinda dubious about this change having much of an impact in that field.

thiemowmde:
Yes, there are two actual problems:

1. Readability. This could be a really, really simple implementation, but instead you have to jump back and forth between all these super-trivial methods to finally understand what's happening. I find such an excessive splitting not helpful.
2. It's impossible to have separate messages for the three cases "parsing failed", "no matching parser found" and "no parser configured at all".

JeroenDeDauw:
I disagree this makes readability better

thiemowmde:
So you agree with 2. then? Fine. Can we move on?

thiemowmde:
After rebasing this is now identical to #346.

PR #388, opened 2015-02-20
Move implementations from Claim to Statement

thiemowmde:
This is split from #317 to make it easier to review. This introduces code duplication that is going to be removed in #317.

Note that I'm ordering the methods different here. This also avoids some mistakes still in #317, like calling the class a "claim".


JeroenDeDauw:
Cannot make protected field private in derivative

thiemowmde:
Temporarily reverted to protected.

PR #390, opened 2015-02-23
Don't access vendor/bin in composer ci

thiemowmde:
Remove "vendor/bin/" as suggested by @Krinkle in https://github.com/wmde/Diff/commit/acd15a05c6317c0a5a1ea446eca01123fce4bca1#commitcomment-9861167.

I'm also moving the `ci` command to the top.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,16 +44,16 @@
 		}
 	},
 	"scripts": {
+		"ci": [
+			"composer test",
+			"composer cs"
+		],
 		"test": [
 			"composer validate --no-interaction",
 			"phpunit"
 		],
 		"cs": [
-			"vendor/bin/phpcs src/* tests/* --standard=phpcs.xml --extensions=php -sp"
-		],
-		"ci": [
-			"composer test",
-			"composer cs"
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
pfft

Restoring correct behavior by reverting

thiemowmde:
@JeroenDeDauw, "incorrect" means, it runs perfectly fine, but an old version? Why is this crucial?

Krinkle:
@thiemowmde No, it's crucial because it wasn't using the version installed by Composer. Instead of throwing an error (as would happen in most cases), it returns the old version on @JeroenDeDauw 's machine because he has a global install exposed in his `PATH` that is responding to the second command.

Anyway, the screenshot posted here is irrelevant because that's not how composer works. It doesn't change your bash environment. Only its own. And that worked fine until a recent update broke it. See wmde/Diff#37.

JeroenDeDauw:
> Anyway, the screenshot posted here is irrelevant because that's not how composer works.

Yeah. I made a screenshot of the wrong console output.

PR #395, opened 2015-03-03
Unify variable function params docs

thiemowmde:
Reason for not using spaces in `,...`:
* Most common style across MediaWiki and our code base.
* It's less confusing for tools without a space.

~~I'm removing the "coding standards" tags because I don't see why they are needed and because these two places are the only places where we do this for variable argument functions. It's not a problem in all the other places, why is it here?~~

I re-purposed the patch and reworked the methods to either accept variable-length arguments or an array.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -2,6 +2,7 @@
 
 ## Version 3.0.0 (dev)
 
+* `ReferenceList::addNewReference` and `Statement::addNewReference` support an array of Snaks now
 * The concept of `Claim` is no longer modelled
 	* The `Claim` class itself has been removed, though `Claim` is now a temporary alias for `Statement`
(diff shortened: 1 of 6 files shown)


JeroenDeDauw:
Needs rebase

thiemowmde:
Rebased.

RELEASE-NOTES.md
```diff
@@ -2,6 +2,7 @@
 
 ## Version 3.0.0 (dev)
 
+* `ReferenceList::addNewReference` and `Statement::addNewReference` support an array of Snaks now
```
JeroenDeDauw: Which is kinda less significant than the `Claim` drop :)

thiemowmde: I agree. Not sure why the line ended here.

PR #398, opened 2015-03-05
Ignore empty references when added to ReferenceList

JeroenDeDauw:


src/ReferenceList.php
```diff
@@ -70,6 +70,17 @@ public function addReference( Reference $reference, $index = null ) {
 		}
 	}
 
+	/**
+	 * @see parent::attach
```
thiemowmde: `@see SplObjectStorage::attach` please.

src/ReferenceList.php
```diff
@@ -70,6 +70,17 @@ public function addReference( Reference $reference, $index = null ) {
...
+	 * @param Reference $reference
+	 * @param mixed $data
+	 */
+	public function attach( $reference, $data = null ) {
+		if ( !$reference->isEmpty() ) {
```
thiemowmde: I think you need to add this check to `addReference`. Calling `addReference( new Reference(), 0 )` will still be allowed but shouldn't for consistency.

tests/unit/ReferenceListTest.php
```diff
@@ -141,41 +142,61 @@ public function testRemoveReference( ReferenceList $array ) {
...
+
+	public function testGivenEmptyReference_addReferenceDoesNotAdd() {
+		$reference1 = new Reference( array( new PropertyNoValueSnak( 1 ) ) );
+		$reference2 = new Reference( array( new PropertyNoValueSnak( 2 ) ) );
+		$emptyReference = new Reference( array() );
 
-		$array->rewind();
-		$array->next();
+		$references = new ReferenceList( array( $reference1, $reference2 ) );
+		$references->addReference( $emptyReference );
```
thiemowmde: As hinted above, please add the same test with `addReference( ..., 0 )`.

src/ReferenceList.php
JeroenDeDauw: Done

tests/unit/ReferenceListTest.php
JeroenDeDauw: I've added the test to verify the behaviour is as expected. The bug you hinted at appears to not be there

JeroenDeDauw:
New version in #399. For some reason having problems pushing to this one.

> git push origin empyrefs -f
error: src refspec empyrefs does not match any.
error: failed to push some refs to 'git@github.com:wmde/WikibaseDataModel.git'

tests/unit/ReferenceListTest.php
thiemowmde: Oh, right. I double-checked and addReference depends on append in all cases, with and without index.

PR #403, opened 2015-03-09
Clean-up dataProvider docs and type hints

thiemowmde:
* Rename copy-pasted `$labelText` in description tests.
* Make existing docs more specific.
* Add some missing type hints.
* No empty line between `@dataProvider` and `@param`. Why? The params are what the provider provides, these two things are very strongly connected.
* No `@param` doc for "invalid" providers. Having docs that say `mixed` doesn't help. Some of the docs I'm removing in this patch were wrong.

### DIFF
--- a/tests/unit/ByPropertyIdArrayTest.php
+++ b/tests/unit/ByPropertyIdArrayTest.php
@@ -117,7 +117,6 @@ class ByPropertyIdArrayTest extends \PHPUnit_Framework_TestCase {
 
 	/**
 	 * @dataProvider listProvider
-	 * @param Snak[] $objects
 	 */
 	public function testGetIds( array $objects ) {
 		$indexedArray = new ByPropertyIdArray( $objects );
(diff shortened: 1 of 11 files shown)


tests/unit/ByPropertyIdArrayTest.php
```diff
@@ -387,10 +382,10 @@ public function addProvider() {
 
 	/**
 	 * @dataProvider addProvider
-	 * @param array $objectsSource
-	 * @param object $object
-	 * @param int $index
-	 * @param array $objectsDestination
+	 * @param Claim[] $objectsSource
```
JeroenDeDauw: This does not need to be `Claim`. Also, be careful to not cause conflicts with 3.0.x that  be avoided.

thiemowmde: Oh. Yea, that was stupid. Will fix.

thiemowmde:
Avoided `Claim` in `ByPropertyIdArrayTest`.

tests/unit/Claim/ClaimGuidTest.php
```diff
@@ -90,10 +87,9 @@ public function provideClaimGuids() {
 
 	/**
 	 * @dataProvider provideClaimGuids
-	 *
 	 * @param ClaimGuid $claimGuid
 	 */
-	public function testEquals( $claimGuid ) {
+	public function testEquals( ClaimGuid $claimGuid ) {
```
JeroenDeDauw: Heh - I totally forgot we had this class

PR #411, opened 2015-03-23
Remove methods from the deprecated Claims class that are no depended on by users

JeroenDeDauw:
Not clear yet we can remove the whole class in 3.x, so this is just a step in that direction.


thiemowmde:
+1

mariushoch:
Verified to be unused, looks good to me.

Issue #157, opened 2014-08-22
Deprecate and remove Claims class


JeroenDeDauw:
The Claims class currently has the highest complexity of this component.

![worst](https://cloud.githubusercontent.com/assets/146040/4007770/0f0384b6-29cb-11e4-9d99-d8167f4bdaa6.jpg)

DataModel 1.0 introduces `ClaimList`, which is a partial alternative. `Entity` and `Item` type hint against this type in several methods.

JeroenDeDauw:
#411 removes a bunch of its methods

JeroenDeDauw:
Looks like the only thing missing before we can ditch the class now is an indexer based on GUIDs, as per #304

JeroenDeDauw:
`Claims::indexOf` still needs a replacement

JeroenDeDauw:
So does `insertClaimAtIndex`

JeroenDeDauw:
Wondering if we can just drop `insertClaimAtIndex` / the second parameter of `addClaim`. Not seeing it used anywhere. Ping @thiemowmde @snaterlicious

JeroenDeDauw:
Looks like everything that was in `Claims` that needed a replacement now has such a replacement. So I'm thinking we can finally ditch the thing.

@thiemowmde anything I'm forgetting about?

thiemowmde:
* The Claims constructor is still called 80 times: 24 times in data-model itself, 18 times in data-model-serialization, 33 times in repo and 4 times in lib.
* `addClaim` is still used in 2 places in repo and data-model-serialization. I would like to prepare a patch[*] for the remaining usage in repo. The serialization component needs a new StatementListDeserializer, which is obvious and could be done later when updating the dependency.
* `hasClaimWithGuid` is still used 2 times in repo and 4 times in repo tests. Replacement is simple: `$statementList->getIndexByGuid( $guid ) !== false`.
* `removeClaimWithGuid` is still used 2 times in repo. There is no simple replacement because StatementList does not support removing statements. This is probably ok but means the 2 remaining usages must be replaced with code that recreates a StatementList or avoids manipulating an existing StatementList at all. I would like to prepare a patch[*] for this before removing the Claims class.
* `getClaimWithGuid` is still used 19 times in repo. Replacement does exist but is not nice:
```php
$statements = $statementList->toArray();
$statement = $statements[$statementList->getIndexByGuid( $guid )];
```
Do you have a better suggestion for that?

[...]

PR #413, opened 2015-03-23
Remove more Claims methods

JeroenDeDauw:
Part of #157

* Claims::getHash is not used by any of our code as far as I can tell
* Claims::isEmpty is used by DataModel Serialization though that can easily be changed. I started updating that component to DM 3.x in https://github.com/wmde/WikibaseDataModelSerialization/pull/115


thiemowmde:
Usages of `getHash` in my local code base:
* In `RemoveClaimsTest` (see https://gerrit.wikimedia.org/r/#/c/198768/)

Usages of `isEmpty` in my local code base:
* `RemoveClaimsTest`, again
* 2 times in wikibase/data-model-serialization

So +2.

Issue #304, opened 2014-11-23
Add indexer for statements based on GUID


JeroenDeDauw:
There is no non-deprecated way to look up statements in a list by guid. Adding this will help with removing `Claims` #157.

We need replacements for

* bool hasClaimWithGuid( $claimGuid )
* Claim|null getClaimWithGuid( $claimGuid )
* function removeClaimWithGuid( $claimGuid )

JeroenDeDauw:
Started work on this in https://github.com/wmde/WikibaseDataModel/pull/419

PR #414, opened 2015-03-23
Add StatementList::indexOf

JeroenDeDauw:
This is needed as replecament for Claims::indexOf, as per
https://github.com/wmde/WikibaseDataModel/issues/157

### DIFF
--- a/src/Statement/StatementList.php
+++ b/src/Statement/StatementList.php
@@ -282,4 +282,24 @@ class StatementList implements IteratorAggregate, Comparable, Countable {
 		return empty( $this->statements );
 	}
 
+	/**
+	 * @since 3.0
+	 *
+	 * @param Statement $statement
+	 *
+	 * @return int|bool
+	 */
+	public function getIndexByGuid( Statement $statement ) {
+		$index = 0;
+
+		foreach ( $this->statements as $s ) {
+			if ( $s->getGuid() === $statement->getGuid() ) {
+				return $index;
+			}
+			$index++;
+		}
+
+		return false;
+	}
+
 }
--- a/tests/unit/Statement/StatementListTest.php
+++ b/tests/unit/Statement/StatementListTest.php
@@ -541,4 +541,50 @@ class StatementListTest extends \PHPUnit_Framework_TestCase {
 		);
 	}
 
+	public function testGivenNotPresentStatement_getIndexByGuidReturnsFalse() {
+		$statements = new StatementList();
+
+		$this->assertFalse( $statements->getIndexByGuid( $this->getStatement( 1, 'kittens' ) ) );
+	}
+
+	public function testGivenPresentStatement_getIndexByGuidReturnsItsIndex() {
+		$statements = new StatementList( array(
+			$this->getStatement( 43, 'kittens43' ),
+			$this->getStatement( 42, 'kittens42' ),
+			$this->getStatement( 41, 'kittens41' ),
+		) );
+
+		$this->assertSame(
+			1,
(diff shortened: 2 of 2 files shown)


src/Statement/StatementList.php
```diff
@@ -282,4 +282,24 @@ public function isEmpty() {
...
+	 *
+	 * @param Statement $statement
+	 *
+	 * @return int|bool
+	 */
+	public function indexOf( Statement $statement ) {
+		$index = 0;
+
+		foreach ( $this->statements as $s ) {
+			if ( $s->equals( $statement ) ) {
```
thiemowmde: The method in Claims compared the GUID only, this compares everything in the object. I don't think this is intended.

thiemowmde:
I suggest to implement a GUID based `remove` along with this, which would resolve https://phabricator.wikimedia.org/T78299 (formerly #175).

thiemowmde:
... and what about `contains`? Having an `indexOf` but no `contains` is a bit odd.

src/Statement/StatementList.php
JeroenDeDauw: I'm aware of the difference. Is this an actual problem for the code using `Claims::indexOf`? Note how there can only be a single Claim with a specific GUID in Claims. While this is not the case with StatementList, in Wikibase Repo it generally does not make sense to end up with a StatementList that has Statements with the same GUID. Which is why I think this change does not matter.

JeroenDeDauw:
The GUID stuff is tracked by #304. No objections against having a `contains` method, though I'm not gonna add it for the sake of having it while we have actual issues to get to.

src/Statement/StatementList.php
thiemowmde: This new `indexOf` will fail if there is a statement with the same GUID but, for example, a reordered list of references.

JeroenDeDauw: Statements with differently ordered references are considered equal if I'm not mistaken

thiemowmde: Nope. Was just an example, can also use different ranks as an example. I still think implementations like `indexOf`, `contains` and `remove` should work with the GUID, not with `equals`.

JeroenDeDauw: `removeStatementWithGuid` and `hasStatementWithGuid` for by GUID and `remove` and `contains` for value based comparison seems to make more sense to me. `indexOf` falling into the later category. Perhaps we need the by GUID index lookup, this has still not been demonstrated to me though. Yes, if you change the value then you won't find it. Do we have a use case where this is a problem?

JeroenDeDauw: I still don't get it though now have updated the method

thiemowmde: Now this will do odd things if the list contains statements with no GUID. Dang, I don't know what's the right thing to do. `indexOfGuid( string $guid )`?

I don't think looking at use cases in Wikibase.git helps. Methods in this component should be self-explaining, not depend on use cases specific for a product.

JeroenDeDauw: The only reason I'm adding this is because we need it in Wikibase.git. So it does make sense to look at what is actually needed there. If we create something else, then we're not solving the lack of having this functionality in non-deprecated code.

thiemowmde: Sorry, my comment was confusing. I suggest to rename this to `indexOfGuid` or `getIndexByGuid` or something like that, so `indexOf` is still free for your original implementation, if ever needed.

JeroenDeDauw: Done

thiemowmde:
+2

PR #417, opened 2015-03-25
Remove Claims::indexOf

JeroenDeDauw:


tests/unit/Claim/ClaimsTest.php
```diff
@@ -237,38 +237,14 @@ public function testAddClaim() {
 		// Insert claim at the beginning:
 		$claim3 = $this->makeClaim( new PropertyNoValueSnak( new PropertyId( 'P17' ) ) );
 		$claims->addClaim( $claim3, 0 );
-		$this->assertEquals( 0, $claims->indexOf( $claim3 ), 'Inserting claim at the beginning failed' );
```
JeroenDeDauw: I'm aware this this test method is not asserting a lot any more now. Since the code is deprecated, I suggest we do not waste time on fixing its bad tests though.

thiemowmde: I added an other assertion that should be a sufficient replacement.

RELEASE-NOTES.md
```diff
@@ -19,6 +19,7 @@
 * Removed `Claims::equals` (and `Claims` no longer implements `Comparable`)
 * Removed `Claims::getHash` (and `Claims` no longer implements `Hashable`)
 * Removed `Claims::isEmpty` (you can use `StatementList::isEmpty` instead)
+* Removed `Claims::indexOf` and added `StatementList::getIndexByGuid`
```
thiemowmde: Having a new method hidden in here is a bit odd.

thiemowmde:
+1.5. See comments, but feel free to merge.

RELEASE-NOTES.md
JeroenDeDauw: Ok, I'll add a dedicated note in a follow up commit

tests/unit/Claim/ClaimsTest.php
JeroenDeDauw: Yeah, that's good

JeroenDeDauw:
+2 to changes made. Merge?

PR #418, opened 2015-03-25
Add floobits config

JeroenDeDauw:


thiemowmde:
Why does this need to be in master?

JeroenDeDauw:
Do you think it should be on another branch?

thiemowmde:
Probably. How is this needed by other people outside of the office?

JeroenDeDauw:
It is useful for people that want to open the project with Floobits

thiemowmde:
This doesn't really answer the question why this needs to be in master to open the project with <whatever service people want to use>.

JeroenDeDauw:
It does not need to be there. However, if it is not, it is created, which is then annoying git wise.

thiemowmde:
Self-merge? O_o Question why this needs to be in master is still unanswered.

PR #419, opened 2015-03-25
Added StatementByGuidMap

JeroenDeDauw:
Pair programmed with @KaiNissen 

Tackles https://github.com/wmde/WikibaseDataModel/issues/304

### DIFF
--- a/src/Statement/StatementByGuidMap.php
+++ b/src/Statement/StatementByGuidMap.php
@@ -10,7 +10,10 @@ use Traversable;
 
 /**
  * Ordered and unique collection of Statement objects.
- * Provides indexed access by Statement GUID. Can only contain Statements that have a non-null GUID.
+ * Can only contain Statements that have a non-null GUID.
+ *
+ * Provides indexed access by Statement GUID.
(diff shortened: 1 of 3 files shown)


src/Statement/StatementByGuidMap.php
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	/**
+	 * If the provided statement has a GUID not yet in the map, it will be appended to the map.
+	 * If the GUID is already in the map, the statement with this guid will be replaced.
+	 *
+	 * @param Statement $statement
+	 */
+	public function addStatement( Statement $statement ) {
+		if ( $statement->getGuid() === null ) {
```
KaiNissen: This seems inconsistent. We allow GUIDs represented by e. g. integers here. If we try to get a statement from the list later, an exception is thrown, because it's not a string.

thiemowmde:
Don't forget to update the release notes.

JeroenDeDauw:
b36a1225cc09736c1891b4671e4536fc4a2472fe

PR #420, opened 2015-03-26
Revert "Add floobits config"

thiemowmde:
Reverts wmde/WikibaseDataModel#418

### DIFF
--- a/.floo
+++ b/.floo
@@ -1,3 +1,0 @@
-{
-    "url": "https://floobits.com/wmde/WikibaseDataModel"
-}
--- a/.flooignore
+++ b/.flooignore
@@ -1,3 +1,0 @@
-vendor
-.idea/workspace.xml
-.idea/misc.xml
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@ composer.lock
 
 vendor/
 build/
+
+.floo
+.flooignore


thiemowmde:
Why are we adding meta data from random tools to a code base? We are not adding PHPStorm's `.idea`, for example. Why should we add this?

JeroenDeDauw:
This tool is just as random as ScrutinizerCI, for which we also have such config. What is the problem with this being in the repo?

Benestar:
I agree with Thiemo that we should reduce the number of config files included into the repository to a minimum. ScrutinizerCI is included as it is part of our deployment and testing system but as far as I understand floobit is not used by all developers of this library.

thiemowmde:
@JeroenDeDauw, you realize that "does not hurt" is not an answer to the question who needs this, why and why it needs to be in the master branch in one of our central repositories? There are already a lot of such config files in our repositories (by the way, what's `.coveralls.yml`?) and each raises the same questions: What does it mean, why is it there, must it be considered when changing something in the repository, is it still needed or an outdated fragment? You are often talking about "maintenance costs". I don't see how this is an exception that doesn't need discussion.

thiemowmde:
#418 was a self-merge, so ...

Benestar:
Sorry, I wanted this to merge as well, forgot about it. So no self-merge but approved by me ;-)

PR #421, opened 2015-03-31
Return null rather than false in StatementList::getIndexByGuid

JeroenDeDauw:
We figured out a long time ago that it's better to return null than false.
Accidentally copied one of the remaining false usages over when reimplementing this.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -20,7 +20,7 @@
 * Removed `Claims::equals` (and `Claims` no longer implements `Comparable`)
 * Removed `Claims::getHash` (and `Claims` no longer implements `Hashable`)
 * Removed `Claims::isEmpty` (you can use `StatementList::isEmpty` instead)
-* Removed `Claims::indexOf` (you can use `StatementList::getIndexByGuid` instead)
+* Removed `Claims::indexOf` (you can use `StatementList::getIndexByGuid` instead. caution: returns null instead of false)
 
 #### Additions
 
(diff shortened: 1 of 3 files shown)


thiemowmde:
-1 for now.
* Who is "we"? All "indexOf" code I can find in our code base returns `false`.
* All "indexOf" functions in PHP (e.g. `strpos` and `array_search`) return `false`.

JeroenDeDauw:
"we" is the Wikidata team from 2 years back

thiemowmde:
I don't have a problem with this being merged, but I will not merge it because of the argument given above and:
* `-1`, `false`, `null`, it doesn't really matter as long as it is documented. Returning `null` is perfect for objects (the "Java" way), returning `-1` is nice in type-safe languages (the "C" way), and `false` seems to be the logical decision in variant-type languages like PHP.
* It seems arbitrary and inconsistent to what PHP does. @brightbyte, can you help me understand the argument why `null` is better?

JeroenDeDauw:
The argumentation was that null is a better representation of something not being there than false.

thiemowmde:
Not a strong argument in my opinion. To me `null` sounds like "I can't help you" while `false` sounds like "I'm sure what you are looking for is not here".

JeroenDeDauw:
Right. I'm thinking that if you had noticed this, the argument would be the other way round :)

thiemowmde:
What do you mean with "consistent"? Consistent to what? As I said, I looked up some other indexOf implementations in our code base and they all return false.

thiemowmde:
I also said that I don't have a problem with this being merged.

PR #422, opened 2015-04-01
Increase test coverage for id classes

thiemowmde:
I wonder why these methods never had tests.

### DIFF
--- a/tests/unit/Entity/ItemIdTest.php
+++ b/tests/unit/Entity/ItemIdTest.php
@@ -2,6 +2,7 @@
 
 namespace Wikibase\DataModel\Tests\Entity;
 
+use PHPUnit_Framework_TestCase;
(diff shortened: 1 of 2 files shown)


tests/unit/Entity/ItemIdTest.php
```diff
@@ -15,7 +16,7 @@
  * @licence GNU GPL v2+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
  */
-class ItemIdTest extends \PHPUnit_Framework_TestCase {
+class ItemIdTest extends PHPUnit_Framework_TestCase {
```
JeroenDeDauw: That's inconsistent with what we have elsewhere btw

thiemowmde: I saw Adrian doing this in more and more places. Personally I don't care that much.

PR #423, opened 2015-04-01
Drop abstract SnakObjectTest base class

thiemowmde:
This patch:
* drops the abstract SnakObjectTest base class
* moves all tests to the three subclasses. Note that this introduces code duplication, but most of this duplicate code is so trivial that it's not worth the pain of moving it back into a base class.
* add `@covers Wikibase\DataModel\Snak\SnakObject` tags so the base class is covered as well, even if it does not have it's own test.
* adds a lot of missing tests, e.g. for the serialize methods. Coverage should be at 100% now, I hope.
* Note that the two givenOtherSnakImplementation tests are now in the notEqualsProvider.


thiemowmde:
@JeroenDeDauw, this increases test coverage for the whole component by 1%. Please add your comments if you disagree with certain details of what I did here (like the `@covers` tags).

JeroenDeDauw:
+0

mariushoch:
+1

JeroenDeDauw:
> this increases test coverage for the whole component by 1%

Just FYI: I think that is a bad usage of much metrics

thiemowmde:
Please be more specific. Bad usage of what and why? I'm very surprised by your reaction here. I expected you to merge this instantly, because:
* This patch increases test coverage of the `Property...ValueSnak` classes to 100%.
* Increases overall test coverage.
* Drops an abstract class.

But instead ...? What's wrong with this patch? I honestly don't have the slightest idea.

JeroenDeDauw:
Never said there is something wrong with the patch. Only that what I quoted is bad usage of metrics.

thiemowmde:
Bad? How? I'm stating a fact. What's wrong with that? What's your intention when you are picking some words from a comment and spending time on discussing wording instead of doing a code review?

addshore:
+1

PR #425, opened 2015-04-02
Drop long deprecated Property::newEmpty

thiemowmde:
The last 2 remaining uses of this method I could find (in my local code base, please double check) are removed in:
* https://gerrit.wikimedia.org/r/#/c/201428/
* https://github.com/wmde/WikibaseQuery/pull/61

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -34,6 +34,7 @@ Other breaking changes:
 
 * Removed previously deprecated `Entity::getAllSnaks`, use `StatementList::getAllSnaks` instead
 * Removed previously deprecated `EntityId::getPrefixedId`, use `EntityId::getSerialization` instead
+* Removed previously deprecated `Property::newEmpty`, use `Property::newFromType` or `new Property()` instead
 * `Reference` and `ReferenceList`s no longer can not be instantiated with `null`
 * Added `setId` method to `EntityDocument`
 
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
+1

JeroenDeDauw:
Needs rebase

thiemowmde:
Rebased.

mariushoch:
Please note that this requires a WikibaseDataModelSerialization release as the fixes there aren't part of 1.3.0 (could be that other changes in there also require that, I didn't follow along everything).

thiemowmde:
Are you sure you aren't confusing something here? This removes a deprecated method. This is a breaking change. This does not depend on anything, but other components will obviously break if they still use the removed method. Which is not the case, I already checked that.

mariushoch:
> This does not depend on anything, but other components will obviously break if they still use the removed method. Which is not the case, I already checked that.

Not on master, yes... but AFAIS the stable version of WikibaseDataModelSerialization still uses the old method, thus a new release should be made there before we want to update Wikibase.

thiemowmde:
> a new release should be made there before we want to update Wikibase.

I'm sorry, but how is this relevant for this pull request? The proposed DataModel 3.0.0 already contains a **lot** of breaking changes. The update path you are describing is not exclusive to this change.

JeroenDeDauw:
@thiemowmde Marius is just pointing out something we should not forget about in the grand scheme of things. It is not an objection to your PR. No need to make a fuzz about it.

PR #427, opened 2015-04-02
Un-deprecate Fingerprint convenience methods

thiemowmde:
These currently cause hundreds and hundreds of deprecation warnings across our code base, but we can't do anything about this. I suggest to remove these tags as long as [the discussion related to Fingerprint](https://phabricator.wikimedia.org/T87237) is not finished.

Bug: [T87237](https://phabricator.wikimedia.org/T87237)

### DIFF
--- a/src/Entity/Entity.php
+++ b/src/Entity/Entity.php
@@ -52,8 +52,6 @@ abstract class Entity implements \Comparable, FingerprintProvider, EntityDocumen
 	/**
 	 * Sets the value for the label in a certain value.
 	 *
-	 * @deprecated since 0.7.3 - use getFingerprint and setFingerprint
-	 *
 	 * @param string $languageCode
 	 * @param string $value
 	 *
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
-2. You can do something about it. We should not change the status quo of the deprecation while the discussion is not finished.

thiemowmde:
I'm sorry? What's your intention when you are closing and deleting this? The discussion is not finished.

JeroenDeDauw:
My intention is not having the queue polluted with things that will not get merged. People generally close their changes themselves.

thiemowmde:
People close their changes when there is agreement or when communication happened. I still think this patch can be a step forward in the ongoing discussion.

JeroenDeDauw:
This patch undeprecates methods in a deprecated class. That makes no sense, regardless of the ongoing discussion.

JeroenDeDauw:
@thiemowmde how about undeprecating the most used ones of these in `Item` and `Property` rather than here in `Entity`? That avoids the main problem here.

thiemowmde:
> how about undeprecating the most used ones of these in Item and Property rather than here in Entity?

Sounds good! :-) Will do.

PR #434, opened 2015-04-12
Change StatementList::getIndexByGuid to take a guid rather than a Statement

JeroenDeDauw:


thiemowmde:
+2. Scrutinizer error can probably be ignored.

PR #436, opened 2015-04-13
Deprecate all public methods in Claims

thiemowmde:
This is part of #157. This makes it easier to see the remaining usages of these methods. In some IDEs the class deprecation is not enough.


JeroenDeDauw:
Needs rebase

thiemowmde:
Rebased.

PR #416, opened 2015-03-23
Rework ...List constructors

thiemowmde:
Resubmission of #376 against master.

PR #439, opened 2015-04-13
Add SnakList constructor

thiemowmde:
This introduces a constructor to the SnakList class that does not call the parent constructor. This is done on purpose to decrease the inheritance and possibly get rid of the `ArrayObject` dependency some day.

This also drops the possibility to call the constructor with null ~~and adds the possibility to call it with variable length arguments (as some similar constructors and add methods already do)~~.

This is split from #416.

### DIFF
--- a/src/Snak/SnakList.php
+++ b/src/Snak/SnakList.php
@@ -2,6 +2,8 @@
 
 namespace Wikibase\DataModel\Snak;
 
+use InvalidArgumentException;
+use Traversable;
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
There are comments by @snaterlicious and me on #376 questioning the merrit of this change.

thiemowmde:
Note that the comment above is not applicable any more because I removed the "variable number of arguments" feature entirely.

PR #440, opened 2015-04-13
Do not allow null in Reference[List] constructors

thiemowmde:
This is split from #416 but focuses on a single change: Calling the two constructors with `null` is not allowed any more.

Release notes will be updated later.


JeroenDeDauw:
Update the rel notes please

thiemowmde:
Done: cc6027e90d5b7fea0d36703e5f558201d9f24d27.

PR #441, opened 2015-04-13
Add test to construct ReferenceList from ReferenceList

thiemowmde:
I can not find a test that checks if a ReferenceList can be constructed from an other Traversable object.

### DIFF
--- a/tests/unit/ReferenceListTest.php
+++ b/tests/unit/ReferenceListTest.php
@@ -48,6 +48,15 @@ class ReferenceListTest extends \PHPUnit_Framework_TestCase {
 		);
 	}
 
+	public function testCanConstructWithReferenceListObject() {
+		$reference = new Reference( array( new PropertyNoValueSnak( 1 ) ) );
+		$original = new ReferenceList( array( $reference ) );
+		$copy = new ReferenceList( $original );
+
+		$this->assertSame( 1, $copy->count() );
+		$this->assertNotNull( $copy->getReference( $reference->getHash() ) );
+	}
+
 	/**
 	 * @dataProvider invalidConstructorArgumentsProvider
 	 * @expectedException InvalidArgumentException


JeroenDeDauw:
+2 if #443 is merged into this branch first

thiemowmde:
Wait, what? You created a pull request that depends on an other pull request without showing each commit twice? How did you do that?

JeroenDeDauw:
My pull request does not depend on your pull request. It is a pull request from a branch based off the branch you created into the branch you created.

JeroenDeDauw:
Now the test relies on 3 production code methods not part of what is under test instead of 1. Still think you ought to just use `assertEquals`

thiemowmde:
I explained why I will not use == to compare objects, especially not list objects.

JeroenDeDauw:
You mentioned dark magic PHP comparison behaviour. Is that problem actually occurring though? In other words, does PHPUnit not properly support comparing objects in its `assertEquals` method?

Benestar:
I don't know about any "dark magic" in object comparison. The [documentation](http://php.net/manual/en/language.oop5.object-comparison.php) clearly says

> When using the comparison operator (==), object variables are compared in a simple manner, namely: Two object instances are equal if they have the same attributes and values, and are instances of the same class.

I'm not sure if the documentation is wrong here but I think the behaviour is quite clear.

thiemowmde:
No, it's not. For example, PHP allows adding properties on runtime. These are not part of the contract of the class and should **not** change the equality of two objects, but do so if you use `==` for comparison.

Benestar:
Adding properties on runtime is afaik a feature we never use in this library so not a strong argument. However, if comparing two objects of this class using `===` is discouraged, we should provide an `equals` method instead which can be used in cases like this one.

thiemowmde:
There is an equals method.

Benestar:
Uhm, so where's the problem?

thiemowmde:
I don't know.

JeroenDeDauw:
The equals method is not actually used here, so how is it relevant?

If there is an issue with using `assertEquals` for the comparison here, then outlining the problem in a mail and sending it to the list would be nice, since we are not avoiding its usage elsewhere.

thiemowmde:
We **are** avoiding `assertEquals` on objects, especially on collection objects. But I'm already repeating myself. Why are we discussing this here?

JeroenDeDauw:
Ok, let's stop discussing it. +0

thiemowmde:
How is this question relevant for this patch? Is it broken? Does it **not** cover what it should cover?

JeroenDeDauw:
You created a test that relies on production code it can avoid relying on, and one that will still pass in cases where assertEquals won't. Which raises the question why you do not use assertEquals.

thiemowmde:
> test that relies on production code it can avoid relying on

Each and every test "relies on production code it can avoid relying on". This can't be an argument.

> will still pass in cases where assertEquals won't.

This argument works both ways, when the test shouldn't pass but assertEquals does. So again: not an argument, especially not an argument **for** the bogus assertEquals.

JeroenDeDauw:
> Why are you asking this question like I never answered it

Because I've not recognized you answering it anywhere and am still wondering what the supposed problem with `assertEquals` is. I've seen you make various claims about it being bad, yet never substantiate them.

thiemowmde:
Sorry, #443.

JeroenDeDauw:
Well, there you say PHPUnit just delegates to PHP. What I'd like to see is a concrete example of where this will go wrong. assertEquals is used in many PHP projects for comparing objects, including our projects. If this method is indeed broken for objects, it's important people in our projects understand why, so they can avoid it where appropriate.

> PHPUnit does not add much magic on top of PHP's.

Can you substantiate that? I know it uses [comparator](https://github.com/sebastianbergmann/comparator).

thiemowmde:
[T92487](https://phabricator.wikimedia.org/T92487)

JeroenDeDauw:
If I'm not mistaken, the issue there was PHPs comparison failing, not that of PHPUnit. Which is an argument against doing `assertTrue( $a == $b )`, not against doing `assertEquals( $a, $b )`. Or am I being confused and was it actually `assertEquals` that was the problem?

thiemowmde:
Quoting myself: How is this relevant for this patch? Is it broken? Does it not cover what it should cover?

Benestar:
Adding a `assertEquals` assertion to that test case can be discussed in another place. Having this test is fine and it assures that the reference is contained in the new created `ReferenceList` which is all we want to make sure. I'd say +2

thiemowmde:
Wait, there is still no test for this? An other pull request just rotting and dying since April? Why?

PR #442, opened 2015-04-13
Throw exception when adding empty reference

thiemowmde:
This reworks #399.

I just run into this problem when working on unrelated tests. I added a reference in a test and it was ignored without a hint what's going on. This is at least confusing and hides possible errors.

### DIFF
--- a/src/ReferenceList.php
+++ b/src/ReferenceList.php
@@ -4,6 +4,7 @@ namespace Wikibase\DataModel;
 
 use Hashable;
 use InvalidArgumentException;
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
I suspect this will cause problems in the Wikibase Repository MediaWiki extension. Especially for installs that have empty references in persistence. Those rely on the empty references simply being ignored.

thiemowmde:
Hm, yea, I think you are right. This will probably cause lots of exceptions in production.

PR #443, opened 2015-04-15
Use assertEquals for comparing reference lists in test

JeroenDeDauw:


tests/unit/ReferenceListTest.php
```diff
@@ -52,7 +52,7 @@ public function getConstructorArg() {
 	public function testCanConstructWithReferenceListObject() {
 		$original = new ReferenceList( array( new Reference( array( new PropertyNoValueSnak( 1 ) ) ) ) );
 		$copy = new ReferenceList( $original );
-		$this->assertSame( 1, $copy->count() );
+		$this->assertEquals( $original, $copy );
```
thiemowmde: This uses PHP's dark magic of comparing objects. I wanted to avoid this when I wrote the test.

JeroenDeDauw: I was under the impression that PHPUnit compares objects in a sane way. We are certainly using assertEquals for this in other places.

thiemowmde: Define "sane". PHPUnit does not add much magic on top of PHP's. It delegates the object comparison to PHP. We already run into horrifying problems when comparing list objects that extend one of PHP's classes. I'm very alarmed whenever I see something like this. I will not merge this PR.

thiemowmde:
I reworked the test in #441, please have a look.

PR #444, opened 2015-04-15
Not all Snaks implementations do have an isEmpty

thiemowmde:
This is currently not a problem because we do have a single class implementing the Snaks interface.

A simple workaround is to use `count()`, but I remember this caused big trouble somewhere else and we avoided it on purpose.

### DIFF
--- a/src/Reference.php
+++ b/src/Reference.php
@@ -70,6 +70,7 @@ class Reference implements \Hashable, \Comparable, \Immutable, \Countable {
 	 * @return bool
 	 */
 	public function isEmpty() {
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
Nice catch. PhpStorm is thinking the field is also of type `SnakList` and thus resolves the method.

Alternatively we can drop the `Snaks` interface (which makes little sense and at this is inconsistent with current practices) and have `Reference` always have a `SnakList`.

thiemowmde:
Since #440 PHPStorm doesn't think this any more. That's why I found this.

I wouldn't touch the Snak vs. SnakList hierarchy at this point. The most simple solution is indeed `count()`. Maybe `empty()` works too?

Issue #282, opened 2014-11-16
Remove deprecated claim methods from Entity


JeroenDeDauw:
This is part of #193

JeroenDeDauw:
@thiemowmde this is in fact not done yet

JeroenDeDauw:
Stuff done by https://github.com/wmde/WikibaseDataModel/pull/429 and https://github.com/wmde/WikibaseDataModel/pull/432

JeroenDeDauw:
Also https://github.com/wmde/WikibaseDataModel/pull/428 which is still open

thiemowmde:
What's left:
* `addClaim` is currently used 31 times in my local code base
* `getClaims` is currently used 55 times in my local code base
* `setClaims` (not in `Entity`, but should be removed too) is currently used 9 + 2 times in my local code base

JeroenDeDauw:
The remaining occurrences probably have a high overlap with the OCP violations that block https://phabricator.wikimedia.org/T75496

thiemowmde:
Patches for review:
* https://github.com/wmde/WikibaseInternalSerialization/issues/48
* https://gerrit.wikimedia.org/r/#/c/202389/

PR #448, opened 2015-04-17
[RFC] Drop getClaims/setClaims from all Entity classes

thiemowmde:
`getClaims` is the last remaining method in `Entity` that refers to claim(s).

This patch removes **all** claim methods from **all** Entity classes. I'm not sure if this is a good idea. This will cause a lot of work in Wikibase.git. Are we sure we want to do that?

[Bug: T78290](https://phabricator.wikimedia.org/T78290) | Issue #282

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -11,7 +11,7 @@
 	* `Statement::setClaim` and `Statement::getClaim` have been removed
 	* Removed `ClaimList`
 	* Removed `ClaimListAccess`
-	* Removed `addClaim`, `hasClaims` and `newClaim` from all entity classes
+	* Removed `addClaim`, `getClaims`, `hasClaims`, `newClaim` and `setClaims` from all entity classes
 * `Claims::addClaim` no longer supports setting an index
 * Removed `Claims::getBestClaims` (you can use `StatementList::getBestStatements` instead)
 * Removed `Claims::getByRank` and `Claims::getByRanks` (you can use `StatementList::getWithRank` instead)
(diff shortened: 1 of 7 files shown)


brightbyte:
I'd recommend to have getClaims in Item for now, for B/C, and mark it as @deprecated.
Removing getClaims from Entity (and thus from Property) should not be as bad, right?

thiemowmde:
Yes, that's a possible intermediate step. @JeroenDeDauw?

Benestar:
`Item::getClaims` &rarr; 8 occurences
`Property::getClaims` &rarr; 2 occurences
`Entity::getClaims` &rarr; 31 occurences

I fear that the Entity usages will be the trickiest ones.

thiemowmde:
@Bene, we could remove the method from Entity but keep it in Item and Property. This way no code will break but we will see scary warnings in our IDEs. I'm not sure if this is worth the trouble. The method is deprecated anyway. We don't win much.

Benestar:
Yes, you're right, it won't help much. However, all those places which currently assume that every Entity has claims (ie. statements) should use a `StatementListProvider` instead and that requires lots of refactoring. Thus I think those cases will be the trickiest ones. :worried:

JeroenDeDauw:
> and mark it as `@deprecated`.

All those methods are either already deprecated or in a deprecated class.

@Benestar thanks for the stats

> I fear that the Entity usages will be the trickiest ones.

Yes, many of them are OCP violations are per https://phabricator.wikimedia.org/T75496

> I'd recommend to have getClaims in Item for now, for B/C

Agree, the compat shims are not a big deal and can be kept until they can be easily removed.

> However, all those places which currently assume that every Entity has claims (ie. statements) should use a StatementListProvider instead and that requires lots of refactoring.

Yeah. It's this code that is the real issue and what we ought to focus on, for it's own sake, rather than being able to remove the compat shims.

PR #449, opened 2015-04-17
Catch exception for invalid EntityId in GUID parser

addshore:
As in PHPdoc the ClaimGuidParser::parse method
should return ClaimGuidParsingExceptions

This catchs previously uncaught EntityIdParsingExceptions

This relates to https://phabricator.wikimedia.org/T89100


thiemowmde:
+2.

Issue #446, opened 2015-04-16
Re-evaluate contract of StatementList::toArray


JeroenDeDauw:
Right now the docs state

    @return Statement[] Numerically indexed (non-sparse) array.

If we add the ability to remove elements from the list, the sparseness will have to be death with somewhere. Which we probably do not want to be in this class, since then we'll need to pay the performance penalty for each `toArray` call, even if most users don't care about the keys.

thiemowmde:
Whats the problem with that? Currently the class can give this guarantee and I think it should  continue to do so.

thiemowmde:
Ok, what we can do:
* Make `toArray` return an array of Statement objects, indexed by GUID.
* Calling code can do `array_values` if required.

JeroenDeDauw:
The idea behind the toArray method is to just return the array the class wraps, so one can as cheaply as possible construct another collection type. It also allows doing

```php
foreach ( $statementList->toArray() as $statement ) {
}
```

Which then makes tools understand that $statement is a `Statement` without adding a specific tag (at the cost of a cheap method call once for the entire loop).

So I think it's indeed better if the caller is responsible for doing the `array_values` operation in case they care.

thiemowmde:
Wait a sec. I wonder:
1. The new `StatementList::getIndexByGuid` is already not very useful, because accessing Statements by index is nothing we do often. And it becomes useless the moment we are dropping this guarantee. Should we remove it?
2. Indexing the internal $statements array by GUID is not really possible because a Statement is not guaranteed to have a GUID set.

JeroenDeDauw:
> The discussion in #453 indicates that the contract of this method should not be changed.

How?

thiemowmde:
#453 answers the question how it should **not** change. The remaining question is how you want to change it and why?

JeroenDeDauw:
The reason why I think the current description is problematic is described in the issue here. You've now stated twice that 453 answers describes you think it should not change. Again, how does it describe this? I read through it and am not seeing what the reason is.

thiemowmde:
I can not see how a sparse array can make it into the class, so I wonder why we are discussing this? Why should we **drop** the guarantee and not replace it with an other useful one? Indexing by GUID is out, which is fine. Is there an other option that makes sense?

JeroenDeDauw:
> I can not see how a sparse array can make it into the class

It cannot now. I'm thinking about the future. Please read my first comment once again.

[...]

PR #450, opened 2015-04-20
Replace StatementList::getIndexByGuid with getFirstStatementByGuid

thiemowmde:
This reverts #414 and #434.

Yes, we need a convenient method to get a Statement by GUID from a StatementList (or an array of Statements if multiple Statements do not have a GUID or do have the same GUID, which is currently possible).

But: a method that returns an integer index is not very useful. StatementList is not an ArrayObject and does not have a getter that accepts an integer index. The only way to make use of the index is via `toArray`, a guarantee that may change (see the discussion in #446) and is odd and dangerous.

```php
$array = $list->toArray();
$index = $list->getIndexByGuid( $guid );
$statement = $array[$index];
```

Such code looks suspicious to me. The two calls are unconnected. I'm accessing an array with an index returned by an other method? Are the two methods guaranteed to use the same indexes in the same order? I can't really tell without looking at the implementations of the two methods.

Instead I suggest to have a `getStatementsByGuid` method that returns an array.

TODO: Tests.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -23,7 +23,7 @@
 * Removed `Claims::getHash` (and `Claims` no longer implements `Hashable`)
 * Removed `Claims::hasClaim`
 * Removed `Claims::isEmpty` (you can use `StatementList::isEmpty` instead)
-* Removed `Claims::indexOf` (you can use `StatementList::getIndexByGuid` instead)
+* Removed `Claims::indexOf`, use `StatementList::getFirstStatementByGuid` or `StatementByGuidMap` instead
 * Removed `Claims::removeClaim`
 * Removed `Entity::getAllSnaks`, use `StatementList::getAllSnaks` instead
 * Removed `EntityId::getPrefixedId`, use `EntityId::getSerialization` instead
@@ -32,7 +32,7 @@
 #### Additions
 
 * Added `StatementByGuidMap`
-* Added `StatementList::getIndexByGuid`
+* Added `StatementList::getFirstStatementByGuid`
 * `ReferenceList::addNewReference` and `Statement::addNewReference` support an array of Snaks now
 
 ## Version 2.6.0 (2015-03-08)
--- a/src/Claim/Claims.php
+++ b/src/Claim/Claims.php
@@ -92,7 +92,7 @@ class Claims extends ArrayObject {
 
 	/**
 	 * @since 0.3
-	 * @deprecated since 1.0, use StatementList::getIndexByGuid() instead.
+	 * @deprecated since 1.0, use StatementList::getFirstStatementByGuid() instead.
 	 *
 	 * @param string $claimGuid
 	 *
@@ -116,7 +116,7 @@ class Claims extends ArrayObject {
 
 	/**
 	 * @since 0.3
-	 * @deprecated since 1.0, use StatementList::getIndexByGuid() instead.
+	 * @deprecated since 1.0, use StatementList::getFirstStatementByGuid() instead.
 	 *
 	 * @param string $claimGuid
 	 *
--- a/src/Statement/StatementList.php
+++ b/src/Statement/StatementList.php
@@ -285,27 +285,20 @@ class StatementList implements IteratorAggregate, Comparable, Countable {
 
 	/**
 	 * @since 3.0
+	 * @see StatementByGuidMap
 	 *
 	 * @param string|null $statementGuid
 	 *
-	 * @throws InvalidArgumentException
-	 * @return int|bool Zero-based index or false if not found.
+	 * @return Statement|null The first statement with the given GUID or null if not found.
 	 */
-	public function getIndexByGuid( $statementGuid ) {
-		if ( !is_string( $statementGuid ) && $statementGuid !== null ) {
-			throw new InvalidArgumentException( '$statementGuid needs to be string or null' );
-		}
-
-		$index = 0;
-
-		foreach ( $this->statements as $s ) {
-			if ( $s->getGuid() === $statementGuid ) {
-				return $index;
+	public function getFirstStatementByGuid( $statementGuid ) {
+		foreach ( $this->statements as $statement ) {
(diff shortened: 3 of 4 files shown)


Benestar:
I agree that having a `getStatemenByGuid` method would be more useful. However, why should the method return an array? The guid of each statement is unique per definition so we can return a Statement object here.

thiemowmde:
```php
public function testConflictingGuids() {
	$list = new StatementList();
	$snak = new PropertyNoValueSnak( 42 );

	$statement0 = new Statement( $snak );
	$list->addStatement( $statement0 );

	$statement1 = new Statement( $snak );
	$list->addStatement( $statement1 );

	$statement2 = new Statement( $snak );
	$statement2->setGuid( 'conflict' );
	$list->addStatement( $statement2 );

	$statement3 = new Statement( $snak );
	$statement3->setGuid( 'conflict' );
	$list->addStatement( $statement3 );

	$expected = array(
		$statement0,
		$statement1,
		$statement2,
		$statement3,
	);

	$this->assertSame( 4, $list->count() );
	$this->assertSame( $expected, array_values( $list->toArray() ) );
	$this->assertSame( 0, $list->getIndexByGuid( null ) );
	$this->assertSame( 2, $list->getIndexByGuid( 'conflict' ) );
}
```

Benestar:
What does this test case tell me? Having a test testing unexpected behaviour like supporting conflicts isn't an argument that this behaviour is correct or good.

thiemowmde:
I'm sorry? There is nothing unexpected here. This simply is how the class currently works. It's an ordered list of statements and currently does not care about GUIDs at all.

Benestar:
Again, if the class doesn't care about GUIDs at all, it shouldn't have methods dealing with GUIDs or pretend indexing by them.

thiemowmde:
It doesn't "pretend indexing by" GUID.

I strongly disagree that a class that manages statements should not have methods dealing with the properties of statements (e.g. PropertyIds or GUIDs). This doesn't makes sense.

Benestar:
I would propose to tackle this problem in a more pragmatic manner: Do we actually need a list of statements which supports statements without GUIDs? As far as I can see all places which use StatementList already assigned a GUID to the added statements so I'm not sure if we actually need a support for null GUIDs in this class.

thiemowmde:
I think this is unrelated and should not be discussed here.

Benestar:
Ok, +1 for this commit though :+1:

JeroenDeDauw:
I agree that the code in the first comment is suspicious, and is not something you want to do.

I'm also in favour of dropping these methods if we can. I added them since I figured we needed them to retain feature parity when dropping `Claims`. Perhaps I was to quick when looking at existing usage though.

So +1 from me at this point. Need verification of being able to drop this (ie usage of the 3.0 rel not being seriously blocked on other needed changes in Wikibase.git) for +2.

thiemowmde:
`StatementList::getIndexByGuid` is currently not used, if that's what you are asking for.

My statistics in #157 still apply:
* `Claims::getClaimWithGuid` is still used and not easy to replace. The proposed `StatementList::getStatementByGuid` in this pull request is an easy replacement.
* `Claims::hasClaimWithGuid` is almost unused and can easily be replaced with the proposed `StatementList::getStatementByGuid`.
* `Claims::removeClaimWithGuid` is unused the moment https://gerrit.wikimedia.org/r/#/c/204511/is merged.

The open question I want to discuss is: What's better, a `getStatementByGuid` or a `getStatementsByGuid` (note the "s") that returns an array of statements? This is relevant because the list can contain multiple statements with the same GUID (or multiple where the GUID is null).

JeroenDeDauw:
> StatementList::getIndexByGuid is currently not used, if that's what you are asking for.

Well sure, since this is the new one. What about the old one in `Claims`? I can't find any usages, which makes me wonder what made me reimplement this...

> What's better, a getStatementByGuid or a getStatementsByGuid

For this we need at the users of the code again. What would be most convenient for them? If all else is the same, then just returning the first match, has my preference, since it is simpler.

thiemowmde:
> What about the old one in Claims?

What are you missing in my list above? Some method to search for a GUID is very helpful. Not required, but not heaving such a method means we have to reimplement loops all over the place.

> returning the first match, has my preference [...]
> I really find getFirstStatementWithGuid a more natural name.

Great, thanks. Will make it so.

thiemowmde:
Renamed to `getFirstStatementByGuid` and added tests.

Benestar:
Hmm, `getFirstStatementByGuid` sounds a bit strange when used in repo. As I said somewhere else, on Wikibase Repo and other usages of this class, only unique GUIDs will be inserted into this class so the name might be confusing in this places.

thiemowmde:
As I said, you are talking about an other class. It exists. See `StatementByGuidMap`. I picked this name on purpose because I consider an identically named `getStatementByGuid` method in two classes with different contracts way more confusing.

Benestar:
Then I may ask the question why this class is actually needed. Why have two almost ambigous classes? The only difference is that this class allows storage of items without GUID or with duplicate GUIDs. The use cases of these features are a bit suspicious to me, maybe not even existing.

JeroenDeDauw:
@Benestar suppose you want to represent the intersection between the statements of two items. Or you want to have a list of statements before having an item (and thus do not yet have an entity id, which you need to construct the GUID). You're right that in Wikibase Repo, most of the time we deal with statements that have GUIDs and don't have the same statement twice. That does not mean we should force this limitation upon every user of the DataModel. Lists and Maps are two very common collection types. There is nothing strange about using both in one application to solve different problems.

JeroenDeDauw:
Merging, as @Benestar his concerns are about having StatementList and StatementByGuidMap rather than this change.

Benestar:
Sure, your argument is valid. I rather asked if we actually need a `StatementList` then but this question isn't part of this patch obviously. Furthermore, I already gave my +1 in a comment above. :wink:

PR #453, opened 2015-04-20
[RFC] StatementList internally indexed by GUID

thiemowmde:
See the discussion in #446.

TODO: Tests.
* Add two statements with no GUID. Check if the indexes are `0` and `1`.
* Add two statements with the **same** GUID. Check if the indexes are the GUID for the first and `0` for the second statement.
* Maybe it's better to index the **last** statement with the same GUID by it's GUID? Clarify the contract of the relevant methods accordingly.
* Add a test that mixes statements with conflicting and without GUIDs.
* Add a test that **changes** the GUID of a statement after it was added to the list. The index can not change and will desync.

### DIFF
--- a/src/Statement/StatementList.php
+++ b/src/Statement/StatementList.php
@@ -54,7 +54,7 @@ class StatementList implements IteratorAggregate, Comparable, Countable {
 				throw new InvalidArgumentException( 'Every element in $statements must be an instance of Statement' );
 			}
 
-			$this->statements[] = $statement;
+			$this->addStatement( $statement );
 		}
 	}
 
@@ -87,7 +87,13 @@ class StatementList implements IteratorAggregate, Comparable, Countable {
 	}
 
 	public function addStatement( Statement $statement ) {
-		$this->statements[] = $statement;
+		if ( $statement->getGuid() !== null
+			&& !array_key_exists( $statement->getGuid(), $this->statements )
+		) {
+			$this->statements[$statement->getGuid()] = $statement;
+		} else {
+			$this->statements[] = $statement;
+		}
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


src/Statement/StatementList.php
```diff
@@ -87,7 +87,13 @@ public function getPropertyIds() {
 	}
 
 	public function addStatement( Statement $statement ) {
-		$this->statements[] = $statement;
+		if ( $statement->getGuid() !== null
+			&& !array_key_exists( $statement->getGuid(), $this->statements )
+		) {
+			$this->statements[$statement->getGuid()] = $statement;
+		} else {
+			$this->statements[] = $statement;
```
Benestar: Uh, having string and integer keys in the same array seems really ugly and indicates that there is a problem in the design of this class. If we want indexing by GUID, we have to enforce all statements in this class to have a GUID.

thiemowmde: That would be an other class. The main purpose of **this** class is an ordered list of Statement objects, no matter what their GUID is.

That's why this is an RFC.

Benestar: So if this class does not care about the statements' GUIDs is shouldn't provide methods dealing with that. If however this class has methods dealing with GUIDs, it obviously cares about GUIDs and therefore has to deal with them in a meaningful manner.

thiemowmde: This sentence doesn't make much sense, sorry. "Shouldn't" according to which standard?

Benestar: You can't say that a class which does indexing by GUID does not care about GUIDs, or uses statements "o matter what their GUID is".

thiemowmde: I don't understand what direction this discussion is going and why. This is an RFC. You are free to disagree if you think this is changing the **current** behavior of the class in an unexpected way. That's why this is an RFC.

Benestar: Indeed, I commented that this implementation does not coincide with the current idea of the class and as this is an RFC I think such a comment is completely valid.

thiemowmde: I understand and accept your opposition, even if I'm missing actual arguments in what you write. I'm totally fine with closing this RFC unmerged, along with #446.

JeroenDeDauw:
-2. This class is explicitly defined as not being intended to do such indexing, as per its description. I can certainly see how it's tempting to do this as it is exactly what I did in `Claims`.

thiemowmde:
That's the result I was hoping for. ;-) Thanks for the input guys.

PR #454, opened 2015-04-20
More StatementList tests

thiemowmde:
This is the opposite of what we discussed in #453.

I also moved the private getters to the top.

### DIFF
--- a/tests/unit/Statement/StatementListTest.php
+++ b/tests/unit/Statement/StatementListTest.php
@@ -17,8 +17,65 @@ use Wikibase\DataModel\Statement\StatementList;
  *
  * @licence GNU GPL v2+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ * @author Thiemo Mättig
  */
 class StatementListTest extends \PHPUnit_Framework_TestCase {
+
+	/**
+	 * @param int $propertyId
+	 * @param string|null $guid
+	 * @param int $rank
+	 *
+	 * @return Statement
+	 */
+	private function getStatement( $propertyId, $guid, $rank = Statement::RANK_NORMAL ) {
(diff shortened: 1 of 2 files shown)


tests/unit/Statement/StatementListTest.php
```diff
@@ -17,9 +17,66 @@
...
 class StatementListTest extends \PHPUnit_Framework_TestCase {
 
+	/**
+	 * @param int $propertyId
+	 * @param string|null $guid
+	 * @param int $rank
+	 *
+	 * @return Statement
+	 */
+	private function getStatement( $propertyId, $guid, $rank = Statement::RANK_NORMAL ) {
```
Benestar: `getStatementMock` perhaps?

Benestar:
+1 Yey more tests

tests/unit/Statement/StatementListTest.php
thiemowmde: Personally I really dislike the `get...Mock` naming scheme. Code that calls this should not care if it gets a mock or a real object. As the `@return` tag states.

JeroenDeDauw: Success. Brainwashing of @thiemowmde complete :D

JeroenDeDauw:
All fine. Found this misleading though:

> I also moved the private getters to the top.

This made me think about actual getters in the production code. The things moved in the test are not getters.

thiemowmde:
I disagree. `get` is not a reserved word that can only have a single, well defined meaning. Also: unrelated to this pull request.

JeroenDeDauw:
`getter` has a well defined meaning

thiemowmde:
It doesn't say where the object the caller gets is coming from, so no, calling a get method a getter is not "misleading". Still unrelated to this pull request.

PR #455, opened 2015-04-20
Add `setStatements` to `StatementListProvider`

Benestar:
This method is required since there is no other way to remove or rebuild the list of statements of a `StatementListProvider` without knowing it is an `Item` or `Property` instance.

*Edit:* For an alternative proposal, see #470

### DIFF
--- a/src/StatementListProvider.php
+++ b/src/StatementListProvider.php
@@ -7,7 +7,7 @@ use Wikibase\DataModel\Statement\StatementList;
 /**
  * Interface for classes that contain a StatementList.
  *
- * @since 2.2.0
+ * @since 2.2.0, modified in 3.0
  *
  * @license GNU GPL v2+
  * @author Marius Hoch < hoo@online.de >
@@ -20,6 +20,7 @@ interface StatementListProvider {
 	public function getStatements();
 
 	/**
+	 * @since 3.0
 	 * @param StatementList $statements
 	 */
 	public function setStatements( StatementList $statements );
--- a/src/StatementListProvider.php
+++ b/src/StatementListProvider.php
@@ -19,4 +19,9 @@ interface StatementListProvider {
 	 */
 	public function getStatements();
 
+	/**
+	 * @param StatementList $statements
+	 */
+	public function setStatements( StatementList $statements );
+
 }


JeroenDeDauw:
I have some concerns here:

* Current users of the interface clearly just need the getter. Adding the setter decreases the suitability of the interface for them due to less nice segregation. It might not matter in this case though. Can anyone think of a case where we'd have an implementation of this thing where we'd implement the getter though would not want to implement the setter?
* This adds mutability of the thing holding the collection itself. There might be cases where this is not desired.
* Generally `Provider` interfaces have no setters, so the name becomes less good

This is not a -2 or even a -1

JeroenDeDauw:
We have other `Providers` in DM, such as `PropertyIdProvider`. `PropertyIdProvider` actually has a derivative which is immutable: `Snak`. So adding such a setter to all is not an option.

Benestar:
Other options would be

1. Create another interface which inherits `StatementListProvider` and includes the setter. Possible but maybe too much overhead
2. Add a method to `StatementList` to clear that list or remove particular statements from that list. However, this would ignore "Does not provide complex modification functionality."

From the three options we have I assumed given the current implementations it would be best to add the method directly to this interface. However, if you can think of any use case for a situation @JeroenDeDauw described this might become a problem. If there is no such usecase in our minds, lets just do this solution.

JeroenDeDauw:
> However, this would ignore "Does not provide complex modification functionality."

Not really. Clearing the list is not complex. Neither is removing a statement based on equality or equality of an attribute. What should not be included are things such as removing duplicates or filtering based on best rank. If you want to do such a filter you can put it in there, though not as a mutator.

JeroenDeDauw:
Even if this ends up going wrong, I strongly suspect it'll be significantly easier to fix than the OCP violation that motivated this change. So +1.

thiemowmde:
I would like to solve this "the right way"<sup>TM</sup>. I do not want to block this approach, it makes sense, but count my vote as a -1 please.

thiemowmde:
After very closely related discussions in #446 I think this is not actually needed and should not be done. The gained benefit is minimal. Better solve this the right way.

Benestar:
@thiemowmde Please take a look at #470 which tries to solve this in a different (and now imo more clean) way. Perhaps this is "the right way"<sup>TM</sup> so this PR could be closed then.

thiemowmde:
I still would like to know where this setter is needed.

The "right" way we are proposing for two years now is an entity type registry and/or strategy as described above.

PR #456, opened 2015-04-22
[RFC] Drop IteratorAggregate from StatementList

thiemowmde:
As discussed in https://gerrit.wikimedia.org/r/#/c/204511/5/repo/includes/ChangeOp/ChangeOpClaim.php.

If having an iterable StatementList is not useful and we are calling `toArray` anyway, why have the class iterable in the first place? This will save us a lot of `/* @var Statement $statement */`.

### DIFF
--- a/src/Statement/BestStatementsFinder.php
+++ b/src/Statement/BestStatementsFinder.php
@@ -29,10 +29,6 @@ class BestStatementsFinder {
 	 * @throws InvalidArgumentException
 	 */
 	public function __construct( $statements ) {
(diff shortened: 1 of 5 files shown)


JeroenDeDauw:
Dropping this interface implementation will not save us the annotations. It's using the `toArray` function that will. You're right that the interface implementation is a lot less useful than if you where to always use it (which was never the intention). That however does not mean we should remove it. Having it here is very cheap and gives us the ability to directly pass the class to something expecting an `Iterator`.

PR #458, opened 2015-04-23
ClaimGuid -> StatementGuid

JeroenDeDauw:
As per https://github.com/wmde/WikibaseDataModel/issues/412


JeroenDeDauw:
Will update rell notes in follow up change as usual.

src/Statement/StatementGuidParsingException.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Wikibase\DataModel\Statement;
+
+/**
+ * @since 0.5
```
Benestar: Maybe add a note that this was renamed in 3.0?

Benestar:
+1 not sure if the inline `@since` tags should mention the rename though.

JeroenDeDauw:
Good catch, will amend

mariushoch:
We really need to make sure we don't forget about replacing all the new class aliases in Wikibase before this hits production. (cc @filbertkm )

PR #459, opened 2015-04-26
Alter guarantees provided by StatementList::toArray

JeroenDeDauw:
The idea of this method always has been that it is cheap and can be used to construct different statement collections or to get type hint support in loops

```php
foreach ( $statementList->toArray() as $statement ) { /* ... */ }
```

This means we cannot guarantee a non-spare array if we want to be able to add cheap removal methods (ie `removeStatement`).

### DIFF
--- a/src/Statement/StatementList.php
+++ b/src/Statement/StatementList.php
@@ -226,7 +226,10 @@ class StatementList implements IteratorAggregate, Comparable, Countable {
 	}
 
 	/**
-	 * @return Statement[] Numerically indexed (non-sparse) array.
+	 * Returns the wrapped array of statements. This retrieval operation is cheap.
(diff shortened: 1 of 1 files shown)


thiemowmde:
-2 as discussed in #446. What's the benefit of dropping this guarantee? We can give it and do so. Why is the discussion not mentioned? Why is this merged?

src/Statement/StatementList.php
```diff
@@ -226,7 +226,10 @@ public function getIterator() {
 	}
 
 	/**
-	 * @return Statement[] Numerically indexed (non-sparse) array.
+	 * Returns the wrapped array of statements. This retrieval operation is cheap.
+	 * No guarantees are given about the keys of the returned array.
```
thiemowmde: This is not true.

Issue #240, opened 2014-10-14
Remove Snaks interface


Benestar:
The `Snaks` interface adds no value and should be replaced by the `SnakList` class everywhere it is used. This is basically the same as for the `References` interface #204

PR #461, opened 2015-04-26
Deprecate Snaks interface

JeroenDeDauw:
First step of https://github.com/wmde/WikibaseDataModel/issues/240


JeroenDeDauw:
@Benestar

thiemowmde:
My PHPStorm counts 14 usages of `Snaks` in Wikibase.git and 9 usages in DataModel Serialization. Shouldn't be that hard to replace them. I will try to prepare patches. Maybe we can drop the interface right away in 3.0.0 instead of deprecating it.

Don't forget the release notes.

PR #464, opened 2015-04-26
DataValues 2.x is also supported

JeroenDeDauw:


thiemowmde:
DataValues 2.0 is not released and doesn't need to, in my opinion, see my comment at https://github.com/DataValues/DataValues/pull/23.

PR #467, opened 2015-04-27
Remove Snaks interface

### LINKED ISSUE #240: Remove Snaks interface

thiemowmde:
This fixes #240. Usage of this interface is minimal and very, very easy to fix (simply replace "Snaks" with "SnakList" everywhere, done).


Benestar:
Needs rebase

Issue #469, opened 2015-05-04
Add setters to `FingerprintProvider` and `StatementListProvider`


Benestar:
#455 is already an approach to do this for the `StatementListProvider`. When we want to get rid of `Entity` we have to provide a way to set a `Fingerprint` or a `StatementList` on those providers. Currently on many places typehints refer against an `Entity` object which is apparently bad.

Instead of adding those setters to the interfaces, we could also add setters to the classes themselves (already done on `Fingerprint`, disputed for `StatementList`) or create dedicated setter interfaces which however might be too much overhead we certainly should avoid if possible.

JeroenDeDauw:
As on the pull request:

I have some concerns here:

Current users of the interface clearly just need the getter. Adding the setter decreases the suitability of the interface for them due to less nice segregation. It might not matter in this case though. Can anyone think of a case where we'd have an implementation of this thing where we'd implement the getter though would not want to implement the setter?

* This adds mutability of the thing holding the collection itself. There might be cases where this is not desired.
* Generally Provider interfaces have no setters, so the name becomes less good

This is not a -2 or even a -1

--

I'm thinking we can solve the OCP violation that motivated this change by introducing a strategy either in this class or doing some refactoring to the changeop system and introducing one on a slightly higher level. Checking against this interface is definitely simpler for the thing we looked at. I'd not be surprised if we are forced to go with the strategy approach regardless if we want to fully fix the entity type related OCP violations in the changeop system.

--

> disputed for StatementList

Well not really - no one has stated they think this usecase is not valid or relevant enough.

JeroenDeDauw:
This is an interesting problem though, and I wonder what the existing understanding on the subject is. In a single case where one runs into dealing with subtypes that require such different handling, using the strategy pattern is the thing to generally do. In case of Wikibase Repository, we run into this dozens of times for the same subtypes though. And having dozens of different strategies that at least partially mirror the subtypes is not nice in various ways. Doing what is proposed here allows for a simpler solution in some, definitely not all, of these cases. Does this provide sufficient worth for the risks, and is there not a better approach to begin with? I really don't know, apart from suspecting that this probably is part of the most pragmatic solution.

Benestar:
Given the fact that a `Fingerprint` object itself is also mutable, I think we could make the `StatementList` class mutable and let the interface remain with the getter only. However, I think @thiemowmde opposed adding a remove method to that class.

[...]

PR #470, opened 2015-05-04
Add removeStatementsByGuid and hasStatementWithGuid 	to StatementList

Benestar:
This is another option to make `StatementList` objects mutable. See #455 #446 #459 and #469 for reference.

### DIFF
--- a/src/Statement/StatementList.php
+++ b/src/Statement/StatementList.php
@@ -25,6 +25,7 @@ use Wikibase\DataModel\Snak\SnakList;
  *
  * @licence GNU GPL v2+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ * @author Bene* < benestar.wikimedia@gmail.com >
  */
 class StatementList implements IteratorAggregate, Comparable, Countable {
 
@@ -103,6 +104,39 @@ class StatementList implements IteratorAggregate, Comparable, Countable {
 		$statement->setGuid( $guid );
 
 		$this->addStatement( $statement );
+	}
+
+	/**
+	 * @param string $guid
+	 * @return Statement[]
+	 */
+	public function removeStatementsByGuid( $guid ) {
+		$removedStatements = array();
+
+		foreach ( $this->statements as $index => $statement ) {
+			if ( $statement->getGuid() === $guid ) {
+				$removedStatements[] = $statement;
+				unset( $this->statements[$index] );
+			}
+		}
+
+		$this->statements = array_values( $this->statements );
+
+		return $removedStatements;
+	}
+
+	/**
+	 * @param string $guid
+	 * @return bool
+	 */
+	public function hasStatementWithGuid( $guid ) {
+		foreach ( $this->statements as $statement ) {
+			if ( $statement->getGuid() === $guid ) {
+				return true;
+			}
+		}
+
+		return false;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


src/Statement/StatementList.php
```diff
@@ -109,6 +110,34 @@ public function addNewStatement( Snak $mainSnak, $qualifiers = null, $references
...
 	}
 
+	public function removeStatement( Statement $statement ) {
+		foreach ( $this->statements as $index => $value ) {
+			if ( $statement->equals( $value ) ) {
+				unset( $this->statements[$index] );
+			}
+		}
+
+		$this->statements = array_values( $this->statements );
```
Benestar: Given that this method will not be called very often, we can do `array_values` here without a huge performance loss. This lets us guarantee a non-sparse array in this class all the time.

thiemowmde:
-2. I don't see why this is needed. It violates the current contract of the class. See the referenced discussions.

Benestar:
Imo this does not violate the current contract of the class as a simple `remove` method is no "complex modification functionality". This is clearly needed in all places we want not only to add but also remove `Statement` objects from a `StatementList`.

thiemowmde:
Which places? I worked on the `Claims` related code a lot and have never seen the need for such code outside of the RemoveClaims change op. A single place in the whole code base that can easily construct a new StatementList and pass it to Item::setStatements. None of the currently proposed changes solve the problem we want to solve: Support other entity types.

Benestar:
"pass it to Item::setStatements" &ndash; this is exactly the problem. `Item` has a `setStatements` method but there is no guarantee that every other entity class will have such a method. `StatementListProvider` will not guarantee that as you gave a +2 to my change. Entity doesn't provide such a method either (and we are perhaps dropping that class anyways).

So no, just using `Item::setStatements` is neither a solution nor an argument.

thiemowmde:
The issue you are describing is not solved by stuffing a setter into a provider interface. The issue is also not new. This is the wrong place to discuss this. We are wasting enormous amounts of time on things that are not on any sprint board. Please close this or mark it as [RFC] or something and add your arguments to a Phabricator ticket and wait until we picket that ticket for a sprint.

src/Statement/StatementList.php
```diff
@@ -105,6 +106,34 @@ public function addNewStatement( Snak $mainSnak, $qualifiers = null, $references
 		$this->addStatement( $statement );
 	}
 
+	public function removeStatement( Statement $statement ) {
```
JeroenDeDauw: As far as I can tell, the equivalent method in `Claims` is actually never used. Removing a Statement by GUID is however something currently done, with this commit proving we have use for such a method: https://gerrit.wikimedia.org/r/#/c/204511/

Benestar:
Task on phabricator: https://phabricator.wikimedia.org/T98143

src/Statement/StatementList.php
Benestar: Ok, so this should rather be a `removeByGuid` method I see now. I will update the PR accordingly. Thanks for the notice Jeroen :-)

JeroenDeDauw: Actually I think it is best to create a dedicated PR for adding that method and one for the `clear` one. So a -2 for one method does not block the other

Benestar: Agreed.

PR #471, opened 2015-05-05
Add StatementList::removeStatementsWithGuid

Benestar:


src/Statement/StatementList.php
```diff
@@ -106,6 +107,25 @@ public function addNewStatement( Snak $mainSnak, $qualifiers = null, $references
 	}
 
 	/**
+	 * @param string $guid
+	 * @return Statement[]
```
thiemowmde: This is not a getter, why does it return anything?

src/Statement/StatementList.php
```diff
@@ -106,6 +107,25 @@ public function addNewStatement( Snak $mainSnak, $qualifiers = null, $references
 	}
 
 	/**
+	 * @param string $guid
```
thiemowmde: `string|null`

src/Statement/StatementList.php
```diff
@@ -106,6 +107,25 @@ public function addNewStatement( Snak $mainSnak, $qualifiers = null, $references
...
+	 * @return Statement[]
+	 */
+	public function removeStatementsWithGuid( $guid ) {
```
thiemowmde: Inconsistent naming, see `StatementByGuidMap`.

src/Statement/StatementList.php
JeroenDeDauw: Yep, should not do the return

https://en.wikipedia.org/wiki/Command%E2%80%93query_separation

tests/unit/Statement/StatementListTest.php
```diff
@@ -260,6 +260,36 @@ public function testAddStatementWithGuid() {
 		$this->assertEquals( new StatementList( $statement ), $list );
 	}
 
+	public function testRemoveStatementsWithGuid() {
```
JeroenDeDauw: Please create dedicated test method for different behaviours.

http://c2.com/cgi/wiki?ArrangeActAssert

Doing non-assertion stuff after doing asserts is the thing to avoid.

src/Statement/StatementList.php
```diff
@@ -106,6 +107,17 @@ public function addNewStatement( Snak $mainSnak, $qualifiers = null, $references
...
 
 	/**
+	 * @param string|null $guid
+	 */
+	public function removeStatementsWithGuid( $guid ) {
+		foreach ( $this->statements as $index => $statement ) {
+			if ( $statement->getGuid() === $guid ) {
+				unset( $this->statements[$index] );
+			}
+		}
```
thiemowmde: Why did you removed the `array_values`? Now this messes with the indexes.

PR #472, opened 2015-05-05
Use With instead of By for consistency in this class

Benestar:


thiemowmde:
-2. What consistency?

Benestar:
consistency **in this class** &ndash; please read the commit message

Benestar:
**By**Guid suggests that all statements will be returned grouped **by** their GUID. However, this method returns only one statement **with** the specified GUID. It's English grammar, it has to be `With`.

JeroenDeDauw:
The naming of `StatementByGuidMap` is a different topic. I don't think anyone is saying that `StatementWithGuidMap` would be a sane name.

JeroenDeDauw:
I'll update the readme

thiemowmde:
Still -1. I did not picked the method name by a dice roll or something. I picked it **for consistency** with `StatementByGuidMap`. Please fix both or nothing.

thiemowmde:
@JeroenDeDauw Why the merge? My -1 above was pretty specific. This is not how decisions in a team should be made. What now? Should I submit an other revert patch?

Benestar:
@thiemowmde There is an obvious difference between both use cases. The `StatementByGuidMap` does **grouping by** guid while the method returns a single statement **with** guid.

Using by for grouping and with for selecting is just the correct use of English prepositions.

JeroenDeDauw:
@thiemowmde you told me in person that it really does not matter all that much. Promoting such a thing to whole team discussion is as bikeshed as you can go. We have more important things to direct our attention at

thiemowmde:
Tracked in https://phabricator.wikimedia.org/T98180.

thiemowmde:
> you told me in person that it really does not matter

I found this patch, questioned our naming decision (you introduced it in #414, I reviewed it, I sticked with it in #450, you merged it) as ridiculous as you did and expected you to ignore it.

> removed your -2

And left a -1.

> promoting such a thing to whole team discussion is as bikeshed as you can go.

What I'm promoting as a discussion is how decisions are made in a team.

PR #474, opened 2015-05-06
New methods to add and remove aliases from an AliasGroupList

Benestar:
These methods should act as a replacement for `Entity::addAliases` and `Entity::removeAliases` as those methods are deprecated and there exists no simple replacement yet.

### DIFF
--- a/src/Term/AliasGroupList.php
+++ b/src/Term/AliasGroupList.php
@@ -199,6 +199,39 @@ class AliasGroupList implements Countable, IteratorAggregate {
 	}
 
 	/**
+	 * @since 3.0
+	 *
+	 * @param string $languageCode
+	 * @param string[] $aliases
+	 */
+	public function addAliasesForLanguage( $languageCode, array $aliases ) {
+		if ( !$this->hasGroupForLanguage( $languageCode ) ) {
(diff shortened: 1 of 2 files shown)


thiemowmde:
-2 for the moment. Where is this needed and why?

Benestar:
Please read my commit and PR message:

> These methods should act as a replacement for Entity::addAliases and Entity::removeAliases as those methods are deprecated and there exists no simple replacement yet.

I mean really, it is written there quite clearly, isn't it?

thiemowmde:
There is literally not a single use of `Entity::addAliases` and `Entity::removeAliases` outside of tests. So again, where is this needed and why?

Benestar:
There might be other users of this library who need that functionality for adding and removing aliases. However, you are right that this is not needed for Wikibase itself so I will leave this PR open but make the other patches in the chain not depend on this one. This patch then has low priority.

Thanks for the notice, @thiemowmde :-)

JeroenDeDauw:
I agree that we should have usecases for such methods outside of the tests for deprecated code.

Benestar:
Closing this for now. We can still add those methods when there is a need for them.

PR #475, opened 2015-05-06
Dont use deprecated methods

Benestar:
This removes usage of lots of deprecated b/c methods in this package and replaces them by the new methods.

<s>This actually depends on #474 </s>

[Bug: T98290](https://phabricator.wikimedia.org/T98290)

### DIFF
--- a/src/Entity/Diff/PropertyDiffer.php
+++ b/src/Entity/Diff/PropertyDiffer.php
@@ -79,9 +79,9 @@ class PropertyDiffer implements EntityDifferStrategy {
 	private function toDiffArray( Property $property ) {
 		$array = array();
 
-		$array['aliases'] = $property->getAllAliases();
-		$array['label'] = $property->getLabels();
-		$array['description'] = $property->getDescriptions();
+		$array['aliases'] = $property->getFingerprint()->getAliasGroups()->toTextArray();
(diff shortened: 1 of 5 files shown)


thiemowmde:
-2. As long as https://phabricator.wikimedia.org/T87237 is not solved no new instances of Fingerprint should be introduced.

JeroenDeDauw:
CR+2

src/Entity/Diff/PropertyDiffer.php
```diff
@@ -79,9 +79,9 @@ private function diffPropertyArrays( array $from, array $to ) {
 	private function toDiffArray( Property $property ) {
```
thiemowmde: The full property is not needed inside of this method.

Benestar: Indeed, a `Fingerprint` is sufficient  here. I already thought about creating a separate `FingerprintDiffer` but I think  these 5 lines of code duplication are not worth a new class.

JeroenDeDauw: I concur a new class is not warranted at this point. And I agree a Fingerprint could be passed to this method, though find this an odd observation given the -2.

PR #477, opened 2015-05-06
Drop support for numeric ids in EntityDocument::setId and add proper typehinting

Benestar:
Only `EntityId` objects are accepted in `EntityDocment::setId` per its documentation. This patch introduces proper typehinting in `EntityDocument` which also assures that programming wise. It also drops support for numeric ids (which was only used in tests).

Tracked in https://phabricator.wikimedia.org/T98343

### DIFF
--- a/src/Entity/EntityDocument.php
+++ b/src/Entity/EntityDocument.php
@@ -38,8 +38,10 @@ interface EntityDocument {
 	 *
 	 * @since 3.0
 	 *
+	 * @param EntityId|null $id
+	 *
 	 * @throws InvalidArgumentException if the id is not of the correct type.
 	 */
-	public function setId( $id );
+	public function setId( EntityId $id = null );
 
 }
--- a/src/Entity/Item.php
+++ b/src/Entity/Item.php
@@ -56,24 +56,19 @@ class Item extends Entity implements StatementListProvider {
 	}
 
 	/**
-	 * Can be integer since 0.1.
 	 * Can be ItemId since 0.5.
 	 * Can be null since 1.0.
 	 *
-	 * @param ItemId|int|null $id
+	 * @param ItemId|null $id
 	 *
 	 * @throws InvalidArgumentException
 	 */
-	public function setId( $id ) {
-		if ( $id === null || $id instanceof ItemId ) {
-			$this->id = $id;
+	public function setId( EntityId $id = null ) {
+		if ( !( $id instanceof ItemId ) && $id !== null ) {
+			throw new InvalidArgumentException( '$id must be an instance of ItemId or null' );
 		}
-		elseif ( is_integer( $id ) ) {
-			$this->id = ItemId::newFromNumber( $id );
-		}
-		else {
-			throw new InvalidArgumentException( '$id must be an instance of ItemId, an integer, or null' );
-		}
+
+		$this->id = $id;
 	}
 
 	/**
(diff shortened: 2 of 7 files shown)


thiemowmde:
I disagree. What do we win if we drop that useful convenience feature?

Benestar:
The use of numeric ids is discouraged all over the place and so we should remove as many places as possible where they appear. Also I don't think this is really a useful feature, in tests we can simply create an ItemId object (and tests should actually be exemplary in how they use the code) and I see no other place as well where we should ever deal with numeric ids.

thiemowmde:
The situations where numeric IDs are a problem are not caused by this particular feature, so still -2. This is useful and does not cause harm.

Benestar:
However, having such a feature creates the impression that it is ok to have numeric ids which isn't. Can you tell me cases where this is useful (besides tests which don't count)?

thiemowmde:
A 3.0 release can not be blocked by a draft that has a -2. All I could do at this point is repeating myself. Killing a useful feature that doesn't hurt anybody has nothing to do with "quality".

Benestar:
@thiemowmde I also don't go ahead and remove the milestone from #466 just because I think that in my very personal opinion it cannot block a 3.0 release.

Which things block a 3.0 release is the decision of PM and should indeed be discussed on phabricator. I will create a task there as this isn't as obvious as I thought it to be.

However, please don't remove labels/milestones from a patch just because of your very own opinion.

All in all, the milestons set here on Github have no value as the relevant tasks are all on phabricator.

thiemowmde:
Please stop adding misleading labels and blocking milestones.

Benestar:
As written, those labels and milestones have no actual meaning and thus don't hurt, or are you personally offended by them?

thiemowmde:
I want to use the milestone as a reminder for the minimal set of patches that must be reviewed before a release. I just removed some non-critical, nice-to-have test-only changes from the milestone.

If there is no agreement, a change can't be in the milestone. Since I disagree that #459 can be part of 3.0, and there is no easy way to branch without that breaking change, #466 must be in the 3.0 milestone.

JeroenDeDauw:
> The use of numeric ids is discouraged all over the place

I'm sorry if you got that impression, since that is probably bad communication on my end. The thing we want to avoid is code that assumes all types of entity ids have a numeric part. Doing this is bad:

```php
foo( EntityDocument $e ) {
    $e->setId( 42 );
}
```

Doing this is fine:

```php
foo( Item $i ) {
    $i->setId( 42 );
}
```

The later does not introduce any bad assumptions, and as far as I can see, has no issues that the more verbose form does not have:

```php
foo( Item $i ) {
    $i->setId( new ItemId( 'Q42' ) );
}
```

EntityDocument::setId does not specify it supports numeric ids. However, due to constraints in the PHP language, we cannot type hint against `EntityId` there, since this cannot be nullified in the derivatives.

Benestar:
Not having this constraint in the PHP language however may be a source for errors or bad assumptions as you pointed out above. Given the very few use cases we have with integer parameters, I still think it is better for the cost of maintenance and security to remove this functionality.

JeroenDeDauw:
Note that code actually doing this will gets static analysis tools shouting at it already, since only `EntityId` and `null` are in the type hint docs:

```php
foo( EntityDocument $e ) {
    $e->setId( 42 );
}
```

There was discussion about how to type hint this when we introduced the method. I think what was settled upon is the most reasonable trade-off. What you suggest is certainly more clean and all, though personally I think it is not the most pragmatic approach.

Benestar:
Closing this as I'm a fan of pragmatic solutions. ^^ I hope there will not be any confusion on this though.

PR #481, opened 2015-05-07
Add setStatements to StatementList

Benestar:
As @thiemowmde [has pointed out correctly](https://phabricator.wikimedia.org/T98143#anchor-1265330), `StatementList::removeStatementsWithGuid` alone does not suffice all our needs. We also want to have a method to set a whole list of statements to the `StatementList`.

This patch just takes code from the constructor and makes it a publicly available method, so that everyone dealing with a `StatementList` no has full access and can also set the whole list. This is especially useful for classes like `ByPropertyIdGrouper` or (proposed) `ByPropertyIdMap` which return a list of statements that can be passed to this method directly.

### DIFF
--- a/src/Statement/StatementList.php
+++ b/src/Statement/StatementList.php
@@ -49,6 +49,8 @@ class StatementList implements IteratorAggregate, Comparable, Countable {
 	}
 
 	/**
+	 * @since 3.0
+	 * 
 	 * @param Statement[]|Traversable
 	 *
 	 * @throws InvalidArgumentException
--- a/src/Statement/StatementList.php
+++ b/src/Statement/StatementList.php
@@ -45,10 +45,20 @@ class StatementList implements IteratorAggregate, Comparable, Countable {
 			$statements = func_get_args();
 		}
 
+		$this->setStatements( $statements );
+	}
+
+	/**
+	 * @param Statement[]|Traversable
+	 *
+	 * @throws InvalidArgumentException
+	 */
+	public function setStatements( $statements ) {
 		if ( !is_array( $statements ) && !( $statements instanceof Traversable ) ) {
 			throw new InvalidArgumentException( '$statements must be an array or an instance of Traversable' );
 		}
 
+		$this->statements = array();
 		foreach ( $statements as $statement ) {
 			if ( !( $statement instanceof Statement ) ) {
 				throw new InvalidArgumentException( 'Every element in $statements must be an instance of Statement' );
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -44,6 +44,7 @@ Other breaking changes:
 * Added `StatementByGuidMap`
 * Added `StatementList::getFirstStatementWithGuid`
 * Added `StatementList::removeStatementsWithGuid`
+* Added `StatementList::setStatements`
 * `ReferenceList::addNewReference` and `Statement::addNewReference` support an array of Snaks now
 * Added PHPMD support
 
--- a/tests/unit/Statement/StatementListTest.php
+++ b/tests/unit/Statement/StatementListTest.php
@@ -666,4 +666,40 @@ class StatementListTest extends \PHPUnit_Framework_TestCase {
 		$this->assertNull( $statements->getFirstStatementWithGuid( false ) );
 	}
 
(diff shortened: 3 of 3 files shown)


src/Statement/StatementList.php
```diff
@@ -45,10 +45,20 @@ public function __construct( $statements = array() /*...*/ ) {
...
+	 * @param Statement[]|Traversable
+	 *
+	 * @throws InvalidArgumentException
+	 */
+	public function setStatements( $statements ) {
 		if ( !is_array( $statements ) && !( $statements instanceof Traversable ) ) {
 			throw new InvalidArgumentException( '$statements must be an array or an instance of Traversable' );
 		}
 
+		$this->statements = array();
```
JeroenDeDauw: I'm wondering about the position of this relative to the throwing of extensions. If the method now gets say, a list of ints, the statement list will be cleared and then an exception will be thrown. I can't think of a concrete case we are likely to run into where this will be a problem, though do wonder if this is something to best avoid.

Benestar: That is a good point indeed. Perhaps having a local variable here in the first place and setting it to the field only at the end of the method would be better.

JeroenDeDauw: Seems like you'll either need more memory or do the loop twice. Not sure which very minor point is more important :)

thiemowmde:
-2. Guys, what is going on here? Did we gave up reviewing patches? Where is this needed and why? What's wrong with `Item:setStatements`?

JeroenDeDauw:
@thiemowmde I reviewed this. It's not because you did not review it that no review happened.

Benestar:
For clarification, `Item::setStatements` is only defined in `Item`, neither in `StatementListProvider` nor in any other interface. However, api modules or change ops like `ChangeOpClaim` should also be able to handle other entity types then `Item` and `Property`. One part of the refactoring for that is to remove the bindings to those two entity types and this is only possible when there are proper setters in more generel classes like `StatementList`.

thiemowmde:
I still have a problem with this approach. What the added method allows is:
```php
if ( $entityDocument instanceof StatementListProvider ) {
	$statementList = $entityDocument->getStatementList();
	$statementList->setStatements( $newStatements );
}
```
This code looks suspicious and can break any time because there is no guarantee that modifying the StatementList will also modify the Entity. I really don't think this is going in the right direction and I still suggest to revert this and have an actual discussion at the right place at the right time with the right people.

Benestar:
That's actually a good point and we should discuss this. I thought that it was guaranteed that the `StatementList` returned by a `StatementListProvider` always was a reference to the `StatementList` object in side the `StatementListProvider`. However, this guarantee isn't given by the interface unfortunately. When keeping that, my methods of course make no sense as you pointed out correctly. However, I think then also the `addStatement` method doesn't make sense and we should make `StatementList` readonly completely.

There are now several options to fix this issue:

* Add `setStatements` to `StatementListProvider`
* Create another interface which contains `setStatements`
* Add a guarantee to `StatementListProvider` that the `StatementList` returned always is a reference to the object within the provider and that any modifications are reflected in the actual `Item`/`Property`

I'm not sure if this list is complete or if there are also other possible solutions I didn't think of. The easiest solution which has least costs is perhaps the third one which I'd prefer.

What are your opinions? Should I move this comment to phabricator? Can you tell what option you suggest or do you have an option to add?

thiemowmde:
You are right that all methods that modify the returned StatementList do have the same problem. But that's not the only argument why I have a problem with `StatementList::setStatements`. The code example I posted above looks suspicious to me because it attempts to replace the collection of statements stored in an entity with a completely different collection of statements. But instead of simply calling `$item->setStatements( $newStatementList )` it needs to call `$item->getStatementList()->setStatements( $newStatementList() ) )`. That's just an odd level of indirection that doesn't match the nice `StatementListProvider` solution.

The setter makes it possible to repurpose an object and turn it into a completely different object with a single call. It's basically a `$a->repurpose( $b )` call. But this is nothing that should be done and therefor should not be in a public interface. If somebody needs a new object he should create a new object and use that.

Add and remove methods don't have that problem.

I suggest to have an
```php
interface StatementListHolder extends StatementListProvider {
	public function setStatements( $statements );
}
```
on `Item` and `Property` instead.

PR #482, opened 2015-05-07
Move StatementListProvider into the Statement namespace

Benestar:
When having `FingerprintProvider` in the Term namespace, `StatementListProvider` should also be in the Statement namespace.

Furthermore, this makes the src/ folder less crowded.

### DIFF
--- a/Aliases.php
+++ b/Aliases.php
@@ -29,3 +29,12 @@ namespace Wikibase\DataModel\Claim {
 	class ClaimGuidParsingException extends \Wikibase\DataModel\Statement\StatementGuidParsingException {}
 
 }
+
+namespace Wikibase\DataModel {
+
+	/**
+	 * @deprecated since 3.0.0, use the base class instead.
+	 */
+	class StatementListProvider extends \Wikibase\DataModel\Statement\StatementListProvider {}
+
+}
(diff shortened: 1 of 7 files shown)


JeroenDeDauw:
Huh. Bad me for having merged the addition of this interface while it was not in the proper NS...

Good catch

JeroenDeDauw:
You do need to add compat aliases. See `WikibaseDataModel.php` and `Aliases.php`.

thiemowmde:
+2 when the tests pass.

Benestar:
Scrutinizer seems to be broken anyways, travis approved this just now. I think this can be merged then.

thiemowmde:
Rebased.

Aliases.php
```diff
@@ -29,3 +29,12 @@ class ClaimGuidParser extends \Wikibase\DataModel\Statement\StatementGuidParser
 	class ClaimGuidParsingException extends \Wikibase\DataModel\Statement\StatementGuidParsingException {}
 
 }
+
+namespace Wikibase\DataModel {
+
+	/**
+	 * @deprecated since 3.0.0, use the base class instead.
+	 */
+	class StatementListProvider extends \Wikibase\DataModel\Statement\StatementListProvider {}
```
thiemowmde: Wait a second, this is not a class. o_O

Benestar: Argh, you are right :astonished: Does `class_alias` actually work for interfaces? If this is the only issue, I will change this quickly on master.

thiemowmde: #489.

PR #483, opened 2015-05-07
Dont modifiy StatementList when invalid arguments passed to setStatements

Benestar:
@JeroenDeDauw Addresses your comment and adds tests for those cases.

*Edit:* Added a missing `@since 3.0` tag as well.

### DIFF
--- a/src/Statement/StatementList.php
+++ b/src/Statement/StatementList.php
@@ -50,7 +50,7 @@ class StatementList implements IteratorAggregate, Comparable, Countable {
 
 	/**
 	 * @since 3.0
-	 * 
+	 *
 	 * @param Statement[]|Traversable
 	 *
 	 * @throws InvalidArgumentException
@@ -60,14 +60,17 @@ class StatementList implements IteratorAggregate, Comparable, Countable {
 			throw new InvalidArgumentException( '$statements must be an array or an instance of Traversable' );
 		}
 
-		$this->statements = array();
+		$statementsToSet = array();
+
 		foreach ( $statements as $statement ) {
 			if ( !( $statement instanceof Statement ) ) {
 				throw new InvalidArgumentException( 'Every element in $statements must be an instance of Statement' );
 			}
 
-			$this->addStatement( $statement );
+			$statementsToSet[] = $statement;
 		}
+
+		$this->statements = $statementsToSet;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


thiemowmde:
What is `setStatements`? Who reviewed this? Who needs this and why?

tests/unit/Statement/StatementListTest.php
```diff
@@ -702,4 +703,48 @@ public function testGivenNoTraversable_setStatementsThrowsException() {
 		$statements->setStatements( $statement1 );
 	}
 
+	public function testGivenNoStatement_setStatementsThrowsException() {
```
JeroenDeDauw: Did you mean `FivenNonStatement`?

tests/unit/Statement/StatementListTest.php
```diff
@@ -702,4 +703,48 @@ public function testGivenNoTraversable_setStatementsThrowsException() {
...
+		$statement1 = $this->getStatement( 1, 'guid1' );
+		$statements = new StatementList();
+
+		$this->setExpectedException( 'InvalidArgumentException' );
+		$statements->setStatements( array( $statement1, false ) );
+	}
+
+	public function testGivenNoTraversable_setStatementsDoesNotEditList() {
+		$statement1 = $this->getStatement( 1, 'guid1' );
```
JeroenDeDauw: For this test you do not need all these statements

src/Statement/StatementList.php
```diff
@@ -50,7 +50,7 @@ public function __construct( $statements = array() /*...*/ ) {
 
 	/**
 	 * @since 3.0
-	 *
```
thiemowmde: I think this is why #482 fails.

JeroenDeDauw:
@thiemowmde you can easily find that out yourself

tests/unit/Statement/StatementListTest.php
Benestar: Perhaps that is the right way to describe the test in English.

tests/unit/Statement/StatementListTest.php
Benestar: One statement is enough, indeed. However, it does not hurt to have more of them. Is this an issue?

src/Statement/StatementList.php
Benestar: I fixed this directly in master as this change does not need any discussion and makes all the tests fail.

thiemowmde: Sure. Thanks for documenting it.

PR #484, opened 2015-05-07
Same format of alternatives in release notes

Benestar:
I don't know who introduced the brackets, but everywhere else, we use "Removed `bla`, use `blubb` instead" so this should be done here as well.


JeroenDeDauw:
> I don't know who introduced the brackets

Which is fine, since that is really not relevant

JeroenDeDauw:
And feel free to self-merge changes to the release notes. We had them on the wiki once, and they are not code, so no CR is needed.

PR #487, opened 2015-05-13
Partial revert of the With vs. By inconsistency

### LINKED ISSUE #486: Rename WithGuid to ByGuid
The GUID is an identifier. Selecting stuff "with" an identifier sounds odd, it should be selected "by it's identifier".

Discussion at https://phabricator.wikimedia.org/T98180 please.

thiemowmde:
Existing "byId" methods:
* `ByPropertyIdArray::getByPropertyId`
* `ByPropertyIdGrouper::getByPropertyId`
* `SiteLinkList::getBySiteId`

Existing "withId" methods:
* `SiteLinkList::hasLinkWithSiteId`
* `SiteLinkList::removeLinkWithSiteId`
* `StatementByGuidMap::hasStatementWithGuid`
* `StatementByGuidMap::getStatementWithGuid`
* `StatementByGuidMap::removeStatementWithGuid`
* `StatementList::removeStatementsWithGuid`
* `StatementList::getWithPropertyId`
* `StatementList::getFirstStatementWithGuid`
* 3 more in the deprecated `Claims`

Suggested naming scheme:
* The object type should not be repeated if a method returns an instance of `self`, e.g. not ~~`Foo::getFooByBar`~~ but `Foo::getByBar`.
* "has*Object*With" and "remove*Object*With" no matter if the parameter is an unique identifier or not.
* But "removeBy" when the object type is not repeated.
* "getBy" for methods that are guaranteed to return all matching elements, either as a list, or as a single element when the selection criteria is an unique identifier.
* "getWith" otherwise, e.g. `StatementList::getFirstStatementWithGuid`.

This patch tries to apply this naming scheme without introducing more breaking changes. This fixes #486.

[Bug: T98180](https://phabricator.wikimedia.org/T98180)

### DIFF
--- a/src/Statement/StatementByGuidMap.php
+++ b/src/Statement/StatementByGuidMap.php
@@ -72,7 +72,7 @@ class StatementByGuidMap implements IteratorAggregate, Countable {
 	 *
 	 * @return Statement|null
 	 */
-	public function getStatementWithGuid( $statementGuid ) {
+	public function getStatementByGuid( $statementGuid ) {
 		$this->assertIsStatementGuid( $statementGuid );
 
 		if ( array_key_exists( $statementGuid, $this->statements ) ) {
--- a/src/Statement/StatementList.php
+++ b/src/Statement/StatementList.php
@@ -138,6 +138,8 @@ class StatementList implements IteratorAggregate, Comparable, Countable {
 	 * Statements that have a main snak already in the list are filtered out.
 	 * The last occurrences are retained.
 	 *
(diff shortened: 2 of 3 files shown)


JeroenDeDauw:
Looks good to me. I'd merge except that at this point I'm really confused about who is getting mad about which naming. @Benestar any objections to this change from your side?

Benestar:
No objections, though I think this is getting a bit complicated ^^ As long as there is some consistency I'm happy

PR #489, opened 2015-05-13
StatementListProvider is not a class

thiemowmde:
I may confuse something here but I think this is an actual bug.

The test is unrelated, just a mismatch I found, obviously caused by replacing `new EntityId` with `new PropertyId` without renaming the test.

### DIFF
--- a/Aliases.php
+++ b/Aliases.php
@@ -35,6 +35,6 @@ namespace Wikibase\DataModel {
 	/**
 	 * @deprecated since 3.0.0, use the base class instead.
 	 */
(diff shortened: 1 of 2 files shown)


Benestar:
Sorry, somehow missed this PR and did a change on master already. My fault -.-

JeroenDeDauw:
> I may confuse something here but I think this is an actual bug.

The `Aliases.php` file is just an IDE helper which never gets executed, so it's certainly not a bug in the actual library code.

PR #492, opened 2015-05-31
Undeprecate term setters in Item and Property

JeroenDeDauw:
Right now code as follows is deprecated:

```php
$item = new Item();
$item->setLabel( 'en', 'foo' );
$item->setDescription( 'de', 'bar' );
```

This is caused by these setters being deprecated in Entity, which
was done to deprecate code assuming all entities have these terms,
and to discourage bad interface segregation. The above code suffers
from neither these problems, and is a style often used in tests,
and has no clean alternative due to PHPs lack of named parameters.
Hence this undeprecation in Item and Property, keeping the one in
Entity.

### DIFF
--- a/src/Entity/Item.php
+++ b/src/Entity/Item.php
@@ -77,6 +77,36 @@ class Item extends Entity implements StatementListHolder {
 	}
 
 	/**
+	 * @param string $languageCode
+	 * @param string $value
+	 *
+	 * @throws InvalidArgumentException
+	 */
+	public function setLabel( $languageCode, $value ) {
+		$this->fingerprint->setLabel( $languageCode, $value );
+	}
+
+	/**
+	 * @param string $languageCode
+	 * @param string $value
+	 *
+	 * @throws InvalidArgumentException
+	 */
+	public function setDescription( $languageCode, $value ) {
+		$this->fingerprint->setDescription( $languageCode, $value );
+	}
+
+	/**
+	 * @param string $languageCode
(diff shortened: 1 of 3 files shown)


Benestar:
Can we remove the deprecated methods then from `Entity`?

JeroenDeDauw:
We *can* yes. Things will still work. Unless you are mocking an Entity object, and perhaps some other such stuff I'm not thinking about. We will however still have code as follows:

```
function( Entity e ) {
    e->getLabel()
}
```

Those usages will then change from "deprecated method" to "unknown method". This is why I think we should not remove them yet.

src/Entity/Item.php
```diff
@@ -77,6 +77,38 @@ public function setId( $id ) {
 	}
 
 	/**
+	 * @param string $languageCode
+	 * @param string $value
+	 *
+	 * @return string
```
thiemowmde: Uh, I never realized the setters in Entity returned strings. Good you did not copied the `return`code lines. Please remove the `@return`` tags too.

JeroenDeDauw: You want to go back to 2012 in the Wikibase.git version history and see the first versions of Entity ;p We had some crazy stuff, such as putting Entity and EntityDocument in one class, and having a database query in the constructor of SiteLink.

thiemowmde:
+2 from my side. Who else should be asked if she or he does have an opinion on this?

PR #494, opened 2015-05-31
Remove unused MapHasher interface

JeroenDeDauw:
Note: release notes do not need updating due to interface being
in the internal namespace already


Benestar:
https://github.com/wmde/WikibaseDataModel/pull/494/files#diff-1d692efc07d4f828202c2ade7e528a6dL26 still contains an annotation to this interface. The `@see` should be replaced by the comment of the method in the interface.

thiemowmde:
@Benestar, done. +2 from my side. Note to self: I want to do a full grep over our whole code base just to be sure (but can't because I'm on the wrong machine at the moment).

PR #495, opened 2015-06-03
Also allow a Snak parameter in Statement

Benestar:
This makes the transition to DataModel 3.0 much easier on other components, because this allows both the use of the old and the new constructor in `Statement`

This will be a <s>2.6.2</s> **2.7** release.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,9 @@
 # Wikibase DataModel release notes
 
+## Version 2.6.2 (dev)
+
+* The `Statement` constructor now also accepts a `Snak` parameter
+
 ## Version 2.6.1 (2015-04-25)
 
 * Allow installation together with Diff 2.x.
--- a/WikibaseDataModel.php
+++ b/WikibaseDataModel.php
@@ -12,7 +12,7 @@ if ( defined( 'WIKIBASE_DATAMODEL_VERSION' ) ) {
 	return 1;
 }
 
-define( 'WIKIBASE_DATAMODEL_VERSION', '2.6.1' );
+define( 'WIKIBASE_DATAMODEL_VERSION', '2.6.2' );
 
 if ( defined( 'MEDIAWIKI' ) ) {
 	call_user_func( function() {
--- a/src/Statement/Statement.php
+++ b/src/Statement/Statement.php
@@ -45,10 +45,38 @@ class Statement extends Claim {
 	 *
 	 * @param Claim $claim
 	 * @param ReferenceList|null $references
+	 * or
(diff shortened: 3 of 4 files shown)


JeroenDeDauw:
This cannot be a patch release since the change is not a bugfix. http://semver.org/

tests/unit/Statement/StatementTest.php
```diff
@@ -32,6 +32,13 @@ public function testConstructorTakesGuidFromClaim() {
 		$this->assertEquals( 'meh', $statement->getGuid() );
 	}
 
+	public function testConstructorWithtoutClaim() {
```
thiemowmde: Without

WikibaseDataModel.php
```diff
@@ -12,7 +12,7 @@
 	return 1;
 }
 
-define( 'WIKIBASE_DATAMODEL_VERSION', '2.6.1' );
+define( 'WIKIBASE_DATAMODEL_VERSION', '2.6.2' );
```
thiemowmde: 2.7.0

RELEASE-NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Wikibase DataModel release notes
 
+## Version 2.6.2 (dev)
```
thiemowmde: 2.7.0

src/Statement/Statement.php
```diff
@@ -45,12 +45,40 @@ class Statement extends Claim {
 	 *
 	 * @param Claim $claim
 	 * @param ReferenceList|null $references
+	 * or
+	 * @param Snak $mainSnak
+	 * @param Snaks|null $qualifiers
+	 * @param ReferenceList|null $references
+	 * @param string|null $guid
```
thiemowmde: I'm afraid no tool is able to understand this, but I also don't see a better way. This is meant to be temporary, right?

src/Statement/Statement.php
```diff
@@ -45,12 +45,40 @@ class Statement extends Claim {
...
 	 */
-	public function __construct( Claim $claim, ReferenceList $references = null ) {
+	public function __construct( $claim /* , $args */ ) {
```
thiemowmde: Suggestion:
```php
/**
 * @since 2.0
 *
 * @param Snak|Claim $mainSnakOrClaim
 * @param Snaks|ReferenceList|null $qualifiersOrReferences
 * @param ReferenceList|null $references
 * @param string|null $guid
 */
public function __construct(
	$mainSnakOrClaim,
	$qualifiersOrReferences = null,
	ReferenceList $references = null,
	$guid = null
) {
```

PR #496, opened 2015-06-03
Do not return anything from Entity setters

thiemowmde:
I realized this in https://github.com/wmde/WikibaseDataModel/pull/492#discussion_r31474648.

~~I consider this a mistake that's not worth any documentation.~~ I did a quick check in our code base and could not find a user relying on these return values.

The renamed method is unrelated, I just found this setter that's not a setter while searching for setters that may return something.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -38,11 +38,12 @@ Other breaking changes:
 * Removed previously deprecated `Property::newEmpty`, use `Property::newFromType` or `new Property()` instead
 * Renamed `StatementList::getWithPropertyId` to `StatementList::getByPropertyId`
 * Renamed `StatementList::getWithRank` to `StatementList::getByRank`
-* `Reference` and `ReferenceList`s no longer can not be instantiated with `null`
-* Added `setId` method to `EntityDocument`
+* `Entity::setLabel` and `Entity::setDescription` no longer return anything
+* `Reference` and `ReferenceList`s no longer can be instantiated with `null`
 
 #### Additions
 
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
Rel notes do need updating. This bevahour has been there since 0.1 and was replied on back then.

Benestar:
Can somebody analyze our code base and check if anything still relies on this methods returning stuff? Otherwise it would be nice to have this in 3.0 as well.

thiemowmde:
Release notes updated.

RELEASE-NOTES.md
```diff
@@ -38,11 +38,12 @@ Other breaking changes:
...
 * Renamed `StatementList::getWithPropertyId` to `StatementList::getByPropertyId`
 * Renamed `StatementList::getWithRank` to `StatementList::getByRank`
-* `Reference` and `ReferenceList`s no longer can not be instantiated with `null`
-* Added `setId` method to `EntityDocument`
+* `Entity::setLabel` and `Entity::setDescription` no longer return anything
+* `Reference` and `ReferenceList`s no longer can be instantiated with `null`
 
 #### Additions
 
+* Added `EntityDocument::setId`
```
JeroenDeDauw: This is a breaking change

thiemowmde: How is a method added to an interface a breaking change? I can't think of a way how does this could break existing code.

JeroenDeDauw: Imagine you implement the interface.

PR #497, opened 2015-06-10
Restore Property::newEmpty

thiemowmde:
This reverts #425.

When phasing out this method we missed code in Wikibase.git that relies on the fact that all `Entity` classes do have a `newEmpty` method. Most notably https://github.com/wikimedia/mediawiki-extensions-Wikibase/blob/398af20d17c119a37350e5ce3df89940945d7f0f/lib/includes/EntityFactory.php#L114. This is a single place in our code base and we already worked around it. The actual reason why I'm suggesting to restore the function is that other code may also depend on the same assumption. Such code breaks when switching to 3.0 with no easy way forward. People can not look into Wikibase.git and replicate what we did. We currently do not have a good alternative in our own code base.

These are the reasons why I'm suggesting to release this as a bugfix release, even if you could argue that the addition of a method requires a 3.1.0.

It's sad that `newEmpty` is not backed by an interface or something but that's how it is at the moment.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,9 @@
 # Wikibase DataModel release notes
 
+## Version 3.0.1 (2015-06-10)
+
+* Restored removed `Property::newEmpty` for compatibility reasons
+
 ## Version 3.0.0 (2015-06-06)
 
 #### Breaking changes
--- a/WikibaseDataModel.php
+++ b/WikibaseDataModel.php
@@ -12,7 +12,7 @@ if ( defined( 'WIKIBASE_DATAMODEL_VERSION' ) ) {
 	return 1;
 }
 
-define( 'WIKIBASE_DATAMODEL_VERSION', '3.0.0' );
+define( 'WIKIBASE_DATAMODEL_VERSION', '3.0.1' );
 
 if ( defined( 'MEDIAWIKI' ) ) {
 	call_user_func( function() {
--- a/src/Entity/Property.php
+++ b/src/Entity/Property.php
@@ -207,6 +207,15 @@ class Property extends Entity implements StatementListHolder {
 	}
 
 	/**
+	 * @deprecated since 0.7.3, use Property::newFromType() or new Property() instead.
+	 *
+	 * @return Property
+	 */
+	public static function newEmpty() {
+		return self::newFromType( '' );
+	}
+
+	/**
 	 * @since 1.1
 	 *
 	 * @return StatementList


RELEASE-NOTES.md
```diff
@@ -183,7 +187,6 @@ Changes in the `Entity` hierarchy:
 * Changed the constructor signature of `Property`
 * Removed `Entity::setClaims` (`Item::setClaims` has been retained)
 * Removed `Entity::stub`
-* Removed `Property::newEmpty`
```
JeroenDeDauw: This should not be removed, even if it was undone in a later release

JeroenDeDauw:
> other code may also depend on the same assumption

You can find it by searching for `newEmtpy` no?

RELEASE-NOTES.md
thiemowmde: I checked and this line is just not true. The method moved, that's why it may look like it's removed in diffs.

thiemowmde:
With "other code" I mean code I can not search for, e.g. on other peoples computers.

JeroenDeDauw:
Well, those other people that have their own code that uses Wikibase DataModel are responsible themselves to look at the release notes and make sure their code still works with the breaking changes after upgrading. That Wikibase.git had some hard to find usage of this method does not change that. For all we know some other user of Wikibase DataModel has hard to find usages of something else.

thiemowmde:
@JeroenDeDauw, what I'm basically suggesting is: Let's remove all newEmpty methods at the same time in a later major release. Not only one of them while the other in Item is still there. This is confusing. I wish we had not missed the reason why these methods existed when removing the one in Property.

This pull request is still a suggestion. You may disagree.

Benestar:
I agree with Thiemo that we should remove all `newEmpty` methods at once and that this should not have happend in 3.0. My reasoning is not that other components may have a problem but that we in Wikibase itself struggle to find a nice replacement for `newEmpty` at the moment.

Btw, very interesting is also the method `PropertyContent::newEmpty` which just uses a random datatype (`string`) to construct the property: https://github.com/wikimedia/mediawiki-extensions-Wikibase/blob/398af20d17c119a37350e5ce3df89940945d7f0f/repo/includes/content/PropertyContent.php#L73

So +1 to this patch

RELEASE-NOTES.md
JeroenDeDauw: I don't get it. Are you talking about the method here in the README or in the code? Looking at it again, this change in the readme still seems wrong

Benestar: Did you click on the expand link on the left? I also was confused by this removal but it is part of the release notes of our Version 1.0 and github just makes it look somewhat strange.

thiemowmde: `Property:newEmpty` was **not** removed in version 1.0, see https://github.com/wmde/WikibaseDataModel/blob/1.0.0/src/Entity/Property.php#L172.

JeroenDeDauw: The relevant release here is 3.0, not 1.0. It was removed in 3.0 as far as I can tell: https://github.com/wmde/WikibaseDataModel/blob/3.0.0/src/Entity/Property.php

JeroenDeDauw:
I'm for keeping the thing removed now we did that and released it, with the option of re-adding it if the removal is really causing a concrete problem somewhere. So yeah, I disagree with that part. Perhaps not with any discussion about if what was done was correct or not, though then I think "what should have happened" is not the same as "what should happen" due to changed circumstances (the release being made).

addshore:
I do not believe we should be adding things back that we have previously removed and also kind of don't want..

thiemowmde:
@addshore, we try to avoid that, but are reverting patches if they turn out to be an issue. Same here. So I wonder what your comment means? You probably don't mean "reverts shouldn't be done any more". What do you mean?

Benestar:
Haven't the issues be resolved in Wikibase.git (in a more or less ugly way but not worse than before)?

thiemowmde:
@Benestar, yes. My initial commit message already answered this question.

Benestar:
I read your commit message again and I think if any third-party user of this library existed and had problems with that change, they would've already told us so please don't re-add this evil stuff.

JeroenDeDauw:
We now have 3 "we should not do this". Time to close?

thiemowmde:
I applied the two readme changes that where in this pull request, see 772ce82be12a050d1481709555725fd488e4a76a and bbed7bbd6681c8308ca7aba75696c367df5b9ba8.

Issue #212, opened 2014-09-25
Do not inherit from SPL array classes


Benestar:
Inheriting from those classes invites issues like objects of the wrong type being used in some List classes and generally misuse inheritance. We should rather use arrays as private variables of those classes and make them implement `IteratorAggregate`, `Countable` and `Comparable`. The addition of `addObject` and `removeObject` methods gives us more control which types of objects can be added to the lists.

Benestar:
@JeroenDeDauw @thiemowmde I think we can just remove `HashArray` and `HashableObjectStorage` completely as they are used only once in the datamodel (`SnakList` extends `HashArray` and `ReferenceList` extends `HashableObjectStorage`). Those base classes are basically an implementation detail and afaik should not be used outside of the datamodel. So just removing them and adding the relevant code to the dedicated subclasses is the cleanest way to solve this problem imo. Any thoughts?

JeroenDeDauw:
You can indeed move the code these abstract classes contain into their derivatives. And while that will be a breaking change, it is, like you noted, one we can easily make. That step alone does not get rid of deriving from the SPL classes though. In any case, sounds good to me

PR #498, opened 2015-06-11
Remove HashableObjectStorage

Benestar:
HashableObjectStorage was a package-private base class which was
only used in ReferenceList. This patch just removes this inheritance,
adds all methods to ReferenceList and removes HashableObjectStorage.
More refactoring in ReferenceList will be done in a follow-up patch.

See #212


src/ReferenceList.php
```diff
@@ -23,7 +27,7 @@
  * @author H. Snater < mediawiki@snater.com >
  * @author Thiemo Mättig
  */
-class ReferenceList extends HashableObjectStorage {
+class ReferenceList extends SplObjectStorage implements Comparable {
```
thiemowmde: I'm not sure if this is an improvement. I like it more if that bogus SPL stuff is in an other class.

src/ReferenceList.php
```diff
@@ -241,4 +245,59 @@ public function unserialize( $data ) {
 		$this->__construct( unserialize( $data ) );
 	}
 
+	/**
+	 * Removes duplicates bases on hash value.
+	 *
+	 * @since 0.2
+	 */
+	public function removeDuplicates() {
```
thiemowmde: Is this even need in any code base?

src/ReferenceList.php
```diff
@@ -241,4 +245,59 @@ public function unserialize( $data ) {
...
+	}
+
+	/**
+	 * The hash is purely valuer based. Order of the elements in the array is not held into account.
+	 *
+	 * @since 0.3
+	 *
+	 * @return string
+	 */
+	public function getValueHash() {
```
thiemowmde: Does this need to be public? Where is this used outside of this class?

src/ReferenceList.php
Benestar: I decided to do it this way round (first remove `HashableObjectStorage`, then refactoring the class). If you prefer it doing the other way round (first removing the SPL stuff) and have some arguments I can also do it that way.

src/ReferenceList.php
Benestar: No, we should remove this method (and many others). However, I'm not sure if our next release should be a breaking release again. :S

src/ReferenceList.php
Benestar: This is called once in `Statement`. However, it should call the `MapValueHasher` there and this method should be removed as well.

thiemowmde:
I understand that this patch is meant to be as simple as possible, to just move stuff without changing the public interface of the ReferenceList class. But I'm not sure if this is worth the trouble and our combined developers time. This is not in a sprint at the moment.

I don't want to block this patch, but I will not +1 it at the moment.

Benestar:
@thiemowmde You don't have to review this patch in the current sprint if you don't have time. I'm just preparing some stuff, maybe for the next sprint. These classes are imo very unlikely to be modified next time so there won't be any merge conflicts.

thiemowmde:
I know, everything is fine. :-) Go on.

src/ReferenceList.php
thiemowmde: Good point.

src/ReferenceList.php
thiemowmde: Go on. You can add an other commit (in an other pull request) on top of this.

JeroenDeDauw:
@Benestar How do you figure this thing is package private?

Benestar:
I assumed they were but apperently I just hoped that :P So this is going 
to be a breaking change anyways, right? I'm not a fan of doing another 
major release after we just did one though. :S What are your thoughts?

Am 11.06.2015 um 16:28 schrieb Jeroen De Dauw:

JeroenDeDauw:
Just keep this open for the next breaking release. Doing two close together is not an issue, but at this point we simple do not have anything that warrants one.

Benestar:
Sounds good, so I will just continue working on this branch and when we are ready for the next major release we merge them. This issue isn't urgent, just nice to have. ;-)

Benestar:
Anything still blocking this? @JeroenDeDauw @thiemowmde @addshore

JeroenDeDauw:
Meh, looks like we forgot to include this in 4.0? >_> Well, same argumentation still applies, let's wait till we actually have sufficient stuff to warrant a new breaking release. Though let's try to not forget it this time...

JeroenDeDauw:
PHPMD agrees we should remove this class

> The class ReferenceList has 11 public methods. Consider refactoring ReferenceList to keep number of public methods under 10.

:)

JeroenDeDauw:
Needs rebase

src/ReferenceList.php
```diff
@@ -55,31 +62,23 @@ public function __construct( $references = array() ) {
 	 * @throws InvalidArgumentException
 	 */
 	public function addReference( Reference $reference, $index = null ) {
-		if ( !is_int( $index ) && $index !== null ) {
-			throw new InvalidArgumentException( '$index must be an integer or null' );
+		if ( $index !== null && ( !is_int( $index ) || $index < 0 ) ) {
+			throw new InvalidArgumentException( '$index must be a non-negative integer or null' );
```
thiemowmde: The non-negative check is a breaking change.

src/ReferenceList.php
```diff
@@ -46,6 +52,7 @@ public function __construct( $references = array() ) {
 
 	/**
 	 * Adds the provided reference to the list.
+	 * Empty references are ignored.
```
thiemowmde: This is a breaking change.

src/ReferenceList.php
```diff
@@ -220,25 +177,94 @@ public function getReference( $referenceHash ) {
...
-	 * @see Serializable::serialize
+	 * Returns the index of a reference or false if the reference could not be found.
+	 *
+	 * @since 0.5
+	 *
+	 * @param Reference $reference
+	 *
+	 * @return int|boolean
+	 */
+	public function indexOf( Reference $reference ) {
```
thiemowmde: This should be moved up so the diff is more minimal.

src/ReferenceList.php
```diff
@@ -220,25 +177,94 @@ public function getReference( $referenceHash ) {
...
+	public function getHash() {
+		$hashes = array();
+
+		foreach ( $this->references as $reference ) {
+			$hashes[] = $reference->getHash();
+		}
+
+		sort( $hashes );
+
+		return implode( '|', $hashes );
```
thiemowmde: This is a breaking change. The old algorithm used MapValueHasher, which used sha1. This one does not, possibly creating strings that are thousands of characters long.

src/Statement/Statement.php
```diff
@@ -231,7 +230,7 @@ public function getHash() {
 			array(
 				sha1( $this->mainSnak->getHash() . $this->qualifiers->getHash() ),
 				$this->rank,
-				$this->references->getValueHash(),
+				$this->references->getHash(),
```
thiemowmde: This must also be announced as a breaking change.

thiemowmde:
I just tried to rebase this patch, but am afraid of the conflicts and possible issues I see. Why are there two commits in here, that do have their own pull requests (#500 and #501) that are already merged? Does this get lost when I rebase and squish this?

Why are there so many unrelated breaking changes in this patch? I tried to mark them all with comments. I think this should be split and discussed separately.

I'm mostly pro this change and most of the breaking changes it includes. I would love to see this moving forward, but I do not know exactly how we could do that. At the moment it seems easier to resubmit a series of much smaller changes instead of continuing this one.

Benestar:
Hah, I tried to resolve this branching and merging weirdness but apparently only messed around even more -.- Perhaps it's the easiest way to create clean new branches with the changes applied?

PR #499, opened 2015-06-12
Remove unused methods

Benestar:
Methods which were never used in our code base or have an easy replacement got removed


thiemowmde:
I don't think you can remove the serialize methods. They are used when calling PHP's `serialize()` function.

Benestar:
We never implemented the `Serializable` interface actually so I think these methods weren't of much use. Furthermore, in the next patches, this class will no longer extend `SplObjectStorage` anyways so one can just serialize the array variable.

thiemowmde:
I was curious and checked and yes, PHP only calls the methods if the class does implement the `Serializable` interface.

Benestar:
Resubmitted against the correct branch #501

PR #506, opened 2015-06-18
[RFC] Add EntityTerms and provider interface

thiemowmde:
* Should the interface also contain some convenience methods, e.g. getLabel, which just returns a string?
* Should the setters be in a separate "holder" interface?
* Tests missing.

[Bug: T87237](https://phabricator.wikimedia.org/T87237)

### DIFF
--- a/src/Term/EntityTerms.php
+++ b/src/Term/EntityTerms.php
@@ -1,0 +1,109 @@
+<?php
+
+namespace Wikibase\DataModel\Term;
+
+use Comparable;
+use InvalidArgumentException;
+
+/**
+ * Imutable value object.
+ *
+ * @since 3.1.0
+ *
+ * @licence GNU GPL v2+
+ * @author Thiemo Mättig
+ */
+class EntityTerms implements Comparable {
+
+	/**
+	 * @var string
+	 */
+	private $languageCode;
+
+	/**
+	 * @var Term
+	 */
+	private $label;
+
+	/**
+	 * @var Term
+	 */
+	private $description;
+
+	/**
+	 * @var AliasGroup
+	 */
+	private $aliases;
+
+	/**
+	 * @param string $languageCode
+	 * @param Term|null $label
+	 * @param Term|null $description
+	 * @param AliasGroup|null $aliases
+	 *
+	 * @throws InvalidArgumentException
+	 */
+	public function __construct(
+		$languageCode,
+		Term $label = null,
+		Term $description = null,
+		AliasGroup $aliases = null
+	) {
+		if ( !is_string( $languageCode ) || $languageCode === '' ) {
+			throw new InvalidArgumentException( '$languageCode must be a non-empty string' );
+		}
+
+		$this->languageCode = $languageCode;
+		$this->label = $label ?: new Term( $languageCode, '' );
+		$this->description = $description ?: new Term( $languageCode, '' );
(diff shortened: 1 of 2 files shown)


Benestar:
I think we should a separate `Holder` interface for setters. Also we should at least have an `hasEntityTerms( $languageCode )` method.

Also, what is returned if no `EntityTerms` are set for the given language code in `EntityTermsProvider::getEntityTerms`? Either return an empty object or throw an `OutOfBoundsException`.

How can we get a list of all `EntityTerms` available (ie. all languages set)?

src/Term/EntityTerms.php
```diff
@@ -0,0 +1,109 @@
...
+		Term $label = null,
+		Term $description = null,
+		AliasGroup $aliases = null
+	) {
+		if ( !is_string( $languageCode ) || $languageCode === '' ) {
+			throw new InvalidArgumentException( '$languageCode must be a non-empty string' );
+		}
+
+		$this->languageCode = $languageCode;
+		$this->label = $label ?: new Term( $languageCode, '' );
```
JeroenDeDauw: I'm concerned about assigning a term with an empty string when there is none. This clashes with how we have been using Term so far. `hasTermForLanguage` will return true for one of these, which is not intended.

thiemowmde: There is no concept of "has" in this proposal.

src/Term/EntityTerms.php
```diff
@@ -0,0 +1,109 @@
...
+
+	/**
+	 * @param string $languageCode
+	 * @param Term|null $label
+	 * @param Term|null $description
+	 * @param AliasGroup|null $aliases
+	 *
+	 * @throws InvalidArgumentException
+	 */
+	public function __construct(
```
thiemowmde: EntityTerms::__construct "has an NPath complexity of 375". WTF, PHPMD? Four lines of straightforward code? No. Just no.

src/Term/EntityTerms.php
Benestar: I'd suggest to just leave $label, $description and $aliases null if they don't exist. Currently, it is impossible in Wikibase to have empty labels because they mean there is no label and get removed. We rely on many places on the fact that a Term can exist or can not exist in a language so there is a  need for `hasTermForLanguage`.

thiemowmde: I understand that some kind of `hasLabelForLanguage` must be possible. But this is unrelated to the current state of this proposal.

src/Term/EntityTerms.php
JeroenDeDauw: PHPMD is correct, and this demonstrates quite nicely how a metric such as NPath is merely a useful tool, not a replacement for understanding the context.

src/Term/EntityTerms.php
JeroenDeDauw: I'm pointing this out because I think it is very relevant. If we organize the terms like this, we'll need to be able to deal with not sets terms in some way.

Is there something else we should consider first?

thiemowmde: There is no "objection". This is a very early work in progress. I'm very thankful for the comments so far and will continue working on this.

thiemowmde:
With #513 going the opposite direction this became pointless.

PR #508, opened 2015-06-25
Implement ExtendedPropertyValueSnak

addshore:


addshore:
https://phabricator.wikimedia.org/T102791

src/Snak/ExtendedPropertyValueSnak.php
```diff
@@ -0,0 +1,87 @@
...
+	 * Support for passing in an EntityId instance that is not a PropertyId instance has
+	 * been deprecated since 0.5.
+	 *
+	 * @param PropertyId|EntityId|integer $propertyId
+	 * @param DataValue $dataValue
+	 * @param DataValue[] $extendedDataValues
+	 */
+	public function __construct( $propertyId, DataValue $dataValue, array $extendedDataValues ) {
+		parent::__construct( $propertyId, $dataValue );
+		foreach ( $extendedDataValues as $key => $extraDataValue ) {
```
Benestar: Readability of this method benefits from new lines before and after the loop.

tests/unit/Snak/ExtendedPropertyValueSnakTest.php
```diff
@@ -0,0 +1,85 @@
...
+
+namespace Wikibase\DataModel\Tests\Snak;
+
+use DataValues\StringValue;
+use Wikibase\DataModel\Entity\PropertyId;
+use Wikibase\DataModel\Snak\ExtendedPropertyValueSnak;
+
+/**
+ * @covers Wikibase\DataModel\Snak\ExtendedPropertyValueSnak
+ *
```
Benestar: `@license` and `@group` tags missing

Benestar:
Code looks fine although I don't know the actual usecases of such a class. Maybe https://www.mediawiki.org/wiki/Wikibase/DataModel#PropertyValueSnak should be extended by this new snak type?

src/Snak/ExtendedPropertyValueSnak.php
```diff
@@ -0,0 +1,87 @@
...
+			if ( !$extraDataValue instanceof DataValue ) {
+				throw new InvalidArgumentException( '$extraDataValues must be an array of DataValue objects' );
+			}
+			if ( !is_string( $key ) ) {
+				throw new InvalidArgumentException( '$extraDataValues keys must be strings' );
+			}
```
Benestar: Not sure what others think, imo we can just put this into one `if` for readability.

src/Snak/ExtendedPropertyValueSnak.php
```diff
@@ -0,0 +1,87 @@
...
+		return $this->extendedDataValues;
+	}
+
+	/**
+	 * @param string $key
+	 *
+	 * @return DataValue|null
+	 */
+	public function getExtendedDataValue( $key ) {
+		if ( array_key_exists( $key, $this->extendedDataValues ) ) {
```
Benestar: I think @thiemowmde did some benchmarks that `isset` is remarkable faster than `array_key_exists` and should be generally preferred.

src/Snak/ExtendedPropertyValueSnak.php
```diff
@@ -0,0 +1,87 @@
...
+			return $this->extendedDataValues[$key];
+		}
+		return null;
```
Benestar: We usually throw an exception in this case afaik.

src/Snak/ExtendedPropertyValueSnak.php
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	/**
+	 * @see Serializable::serialize
+	 *
+	 * @return string
+	 */
+	public function serialize() {
+		return serialize( array(
```
Benestar: There's nothing like `@override` in PHP :S Maybe add a comment that this overrides the methods defined in `PropertyValueSnak` so a) we remember to change this when `PropertyValueSnak` changes and b) nobody gets confused where there is the `implements Serializable` annotation. ;-)

src/Snak/ExtendedPropertyValueSnak.php
```diff
@@ -0,0 +1,87 @@
...
+	 * @param PropertyId|EntityId|integer $propertyId
+	 * @param DataValue $dataValue
+	 * @param DataValue[] $extendedDataValues
+	 */
+	public function __construct( $propertyId, DataValue $dataValue, array $extendedDataValues ) {
+		parent::__construct( $propertyId, $dataValue );
+		foreach ( $extendedDataValues as $key => $extraDataValue ) {
+			if ( !$extraDataValue instanceof DataValue ) {
+				throw new InvalidArgumentException( '$extraDataValues must be an array of DataValue objects' );
+			}
```
brightbyte: ...before and after every block, actually :)

src/Snak/ExtendedPropertyValueSnak.php
addshore: I cant seem to find an example of that in DM. I do however see lots of returning null and false.

src/Snak/ExtendedPropertyValueSnak.php
addshore: doing

src/Snak/ExtendedPropertyValueSnak.php
addshore: done

src/Snak/ExtendedPropertyValueSnak.php
addshore: done

src/Snak/ExtendedPropertyValueSnak.php
Benestar: Oh, you're right. I apparently have a wrong impression of our DM o.O

src/Snak/ExtendedPropertyValueSnak.php
```diff
@@ -0,0 +1,88 @@
...
+	 *
+	 * @param PropertyId|EntityId|integer $propertyId
+	 * @param DataValue $dataValue
+	 * @param DataValue[] $extendedDataValues
+	 */
+	public function __construct( $propertyId, DataValue $dataValue, array $extendedDataValues ) {
+		parent::__construct( $propertyId, $dataValue );
+
+		foreach ( $extendedDataValues as $key => $extraDataValue ) {
+			if ( !$extraDataValue instanceof DataValue || !is_string( $key ) ) {
```
thiemowmde: `!( $extraDataValue instanceof DataValue )` please.

src/Snak/ExtendedPropertyValueSnak.php
```diff
@@ -0,0 +1,88 @@
...
+					'$extraDataValues must be an array of DataValue objects with string keys'
+				);
+			}
+		}
+
+		$this->extendedDataValues = $extendedDataValues;
+	}
+
+	/**
+	 * @return DataValue[]
```
thiemowmde: Since this is guaranteed you can add something like "with string keys" here.

src/Snak/ExtendedPropertyValueSnak.php
```diff
@@ -0,0 +1,88 @@
...
+	public function getExtendedDataValue( $key ) {
+		if ( isset( $this->extendedDataValues[$key] ) ) {
+			return $this->extendedDataValues[$key];
+		}
+		return null;
+	}
+
+	/**
+	 * @see Serializable::serialize
+	 * @note This overrides PropertyValueSnak::serialize
```
thiemowmde: I find this odd. Why not simply `@see PropertyValueSnak::serialize`?

src/Snak/ExtendedPropertyValueSnak.php
```diff
@@ -0,0 +1,88 @@
...
+
+namespace Wikibase\DataModel\Snak;
+
+use DataValues\DataValue;
+use InvalidArgumentException;
+use Wikibase\DataModel\Entity\EntityId;
+use Wikibase\DataModel\Entity\PropertyId;
+
+/**
+ * @since 3.1
```
thiemowmde: Please add a description.

For example, I see that this class does not override `getType`, therefore will be of type "value" and not distinguishable from `PropertyValueSnak`. As far as I know this is intended but should be documented.

src/Snak/PropertyValueSnak.php
```diff
@@ -18,7 +18,10 @@
  */
 class PropertyValueSnak extends SnakObject {
 
-	private $dataValue;
+	/**
+	 * @since 3.1
```
thiemowmde: This was protected before 1.0. Just something I find funny, not worth documentation.

tests/unit/Snak/ExtendedPropertyValueSnakTest.php
```diff
@@ -0,0 +1,90 @@
...
+				'InvalidArgumentException',
+				new PropertyId( 'P9001' ),
+				new StringValue( 'bc' ),
+				array( 'foo' => 'bar' ),
+			),
+		);
+	}
+
+	public function getClass() {
+		return 'Wikibase\DataModel\Snak\ExtendedPropertyValueSnak';
```
thiemowmde: Ugly. Why nobody review #423?

tests/unit/Snak/ExtendedPropertyValueSnakTest.php
```diff
@@ -0,0 +1,90 @@
...
+	public function testHashStability() {
+		$snak = new ExtendedPropertyValueSnak(
+			new PropertyId( 'P1' ),
+			new StringValue( 'a' ),
+			array( 'foo' => new StringValue( 'foo' ) )
+		);
+		$hash = $snak->getHash();
+
+		// @codingStandardsIgnoreStart
+		$expected = sha1( 'C:49:"Wikibase\DataModel\Snak\ExtendedPropertyValueSnak":110:{a:3:{i:0;i:1;i:1;C:22:"DataValues\StringValue":1:{a}i:2;a:1:{s:3:"foo";C:22:"DataValues\StringValue":3:{foo}}}}' );
```
thiemowmde: Can we use this chance to come up with a new serialization (at least for this class) that does *not* include the class name?

src/Snak/ExtendedPropertyValueSnak.php
```diff
@@ -0,0 +1,88 @@
...
+
+	/**
+	 * Support for passing in an EntityId instance that is not a PropertyId instance has
+	 * been deprecated since 0.5.
+	 *
+	 * @param PropertyId|EntityId|integer $propertyId
+	 * @param DataValue $dataValue
+	 * @param DataValue[] $extendedDataValues
+	 */
+	public function __construct( $propertyId, DataValue $dataValue, array $extendedDataValues ) {
```
JeroenDeDauw: I suspect it is not needed to support the deprecated behaviour.

tests/unit/Snak/ExtendedPropertyValueSnakTest.php
JeroenDeDauw: It would be nicer to not follow this not so nice testing pattern for this new code here yes.

src/Snak/ExtendedPropertyValueSnak.php
addshore: Done in next commit, is this part of our CS now?

src/Snak/ExtendedPropertyValueSnak.php
addshore: Done in next commit

src/Snak/ExtendedPropertyValueSnak.php
addshore: Well, bene did make a good point, if at any point that if PropertyValueSnak::serialize changes then this method will also need to be changed.

Perhaps SnakObject could provide a getArrayToSerialize method or something similar which would be implemented everywhere?
This way this new call could call parent::getArrayToSerialize and add the data we need to, meaning changes in the parent would be less likely to break things.

JeroenDeDauw:
If this is being added purely for the serialization component, then how about adding it there initially? If we put it here, we first need to try get it perfect, and then might find that we actually want it somewhat different once we start using it. At which point making the needed changes might force a major release. OTOH, if its in the serialization component, we can work on it more iteratively, and always move it once it reaches stability, if we actually have reason to do so then. Not an objection, just something we should consider.

src/Snak/ExtendedPropertyValueSnak.php
addshore: done in next commits

tests/unit/Snak/ExtendedPropertyValueSnakTest.php
JeroenDeDauw: I suggest to tackle that together with the other snaks. Which will be a breaking change anyway. Since this blocks getting rid of the whole pile of old serialization code in Wikibase Lib, let's not block the introduction of this class on such a change.

src/Snak/PropertyValueSnak.php
addshore: done in next commits

tests/unit/Snak/ExtendedPropertyValueSnakTest.php
addshore: +1, how about a hash of the return value of the serialize method?

src/Snak/ExtendedPropertyValueSnak.php
addshore: done in next patch

addshore:
I would have no problem with it being in the serialization component first.

JeroenDeDauw:
`$extendedDataValues` and associated method names suggest they are dealing with "extended data values", while instead they are dealing with normal data values. `$extensionDataValues` seems more correct in this context. This lacks the deceptive element, even if the name, like the one of the class, is not the most obvious ever. (But who are we kidding, this thing is called `Snak` :)

addshore:
Changed to $extensionDataValues in next commit

src/Snak/ExtendedPropertyValueSnak.php
```diff
@@ -0,0 +1,92 @@
...
+			}
+		}
+
+		$this->extensionDataValues = $extensionDataValues;
+	}
+
+	/**
+	 * @return DataValue[] with string keys
+	 */
+	public function getExtensionDataValues() {
```
brightbyte: "Extension" is odd... How about "extra data values", "secondary data values", or "derived data values"?

src/Snak/ExtendedPropertyValueSnak.php
brightbyte: Yes, getArrayToSerialize, or toArray, would be nice. Could be protected.
NOTE: SnakSerializer doesn't use this at all. We will have to make SnakSerializer aware of ExtendedPVS.

addshore:
Closing, Will move to serialization component and address all comment I have not yet here!

addshore:
After another short discussion I,ll try and leave this in here..!

src/Snak/DerivedPropertyValueSnak.php
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * PropertyValueSnak with with derived values attached.
+ * Derived DataValues can be used to provide infomation, for example, normalized values.
+ *
+ * Calls to the getType method will indicate that this is a PropertyValueSnak
+ *
+ * Direct serialization of this object will not include the extra derived values
+ *
+ * The hash of this object is not changed by the extra Derived values
+ * thus this object will ->equal a PropertyValueSnak of the same value.
```
thiemowmde: Please make sure this is tested.

thiemowmde:
Sooo many comments. And all tackled. :-) +1.999!

addshore:
-73, due to something else that has been merged I need to fix these tests!

JeroenDeDauw:
Why merge? Looks like addshore still wanted to fix something

JeroenDeDauw:
This merge actually broke master, will revert.

addshore:
Yus, the -73 was due to the fact the base class the tests I wrote depends on is no longer there, thus I need to amend the patch! :)
And this is the 1 issue with not having gate submit jobs ;)

PR #511, opened 2015-07-02
Implement DeivedPropertyValueSnak

JeroenDeDauw:
Resubmission of #508 by @addshore which got merged prematurely and broke master


addshore:
I will amend these tests in the morning

JeroenDeDauw:
"Fix phpmd" is a bad commit message, since clearly it should have been "make addshore compatible with phpmd"

PR #512, opened 2015-07-03
Deprecate methods in Entity

Benestar:
Deprecate methods which are in `Entity` but not in `EntityDocument`

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -3,6 +3,11 @@
 ## Version 3.1 (dev)
 
 * Added `DerivedPropertyValueSnak`
+* Deprecated `Entity::copy`
+* Deprecated `Entity::getFingerprint`
+* Deprecated `Entity::setFingerprint`
+* Deprecated `Entity::isEmpty`
+* Deprecated `Entity::clear`
 
 ## Version 3.0.1 (2015-07-01)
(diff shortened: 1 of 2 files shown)


RELEASE-NOTES.md
```diff
@@ -3,6 +3,11 @@
 ## Version 3.1 (dev)
 
 * Added `DerivedPropertyValueSnak`
+* Deprecated `Entity::copy`
+* Deprecated `Entity::getFingerprint`
+* Deprecated `Entity::setFingerprint`
+* Deprecated `Entity::isEmpty`
+* Deprecated `Entity::clear`
```
thiemowmde: Please add what to use instead, if possible. For example "use `new Item` and `new Property` instead".

src/Entity/Entity.php
```diff
@@ -380,6 +381,7 @@ public function getClaims() {
 
 	/**
 	 * @since 0.7.3
+	 * @deprecated since 3.1, use FingerprintProvider instead
```
thiemowmde: This does not make much sense. This method implements the interface, how can one use the interface instead of it's implementation?

JeroenDeDauw:
Closing as Entity is already deprecated, discussed with @Benestar

PR #513, opened 2015-07-03
Introduce provider interfaces for labels, descriptions and aliases

Benestar:
The method names currently conflict with the ones defined in `Entity`. <s>We should consider renaming them to `getLabelList`, `getDescriptionList` and `getAliasList` and rename them again in the next breaking release (4.0).</s>

### DIFF
--- a/src/Term/LabelsProvider.php
+++ b/src/Term/LabelsProvider.php
@@ -3,7 +3,7 @@
 namespace Wikibase\DataModel\Term;
 
 /**
- * @since 4.0
+ * @since 4.1
  *
  * @licence GNU GPL v2+
  * @author Bene* < benestar.wikimedia@gmail.com >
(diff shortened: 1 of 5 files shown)


src/Term/LabelsProvider.php
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * @since 3.1
+ *
+ * @licence GNU GPL v2+
+ * @author Bene* < benestar.wikimedia@gmail.com >
+ */
+interface LabelsProvider {
+
+	/**
+	 * @return TermList
```
thiemowmde: As discussed, please add a short sentence explaining this is not guaranteed to return the original object.

JeroenDeDauw:
Discussed this commit and #514 with @Benestar. He will see if its doable to remove the last usages of the equivalent deprecated methods defined in Entity. If so, we can go with #514.

Benestar:
@addshore @JeroenDeDauw this is no longer a breaking change and should be done in 4.1. There are use cases for those interfaces now as `Fingerprint` already implements them. `Item` and `Property` can also implement those interfaces if the conflicting methods get removed in 5.0.

JeroenDeDauw:
+1

PR #514, opened 2015-07-03
Alternative provider interfaces for labels, descriptions and aliases

Benestar:
This is an alternative to #513 without creating breaking changes

If we decide to do it this way, we should rename the methods to `getLabels`, `getDescriptions` and `getAliases` in 4.0

### DIFF
--- a/src/Term/LabelsProvider.php
+++ b/src/Term/LabelsProvider.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace Wikibase\DataModel\Term;
+
+/**
+ * @since 3.1
+ *
+ * @licence GNU GPL v2+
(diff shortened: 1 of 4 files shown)


src/Term/LabelsProvider.php
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * @since 3.1
+ *
+ * @licence GNU GPL v2+
+ * @author Bene* < benestar.wikimedia@gmail.com >
+ */
+interface LabelsProvider {
+
+	/**
+	 * @return TermList
```
thiemowmde: Same as in https://github.com/wmde/WikibaseDataModel/pull/513#discussion_r33853410.

PR #515, opened 2015-07-03
Add deprecated tags to all Entity methods

thiemowmde:
This is an alternative "resubmission" of #512. Technically the additional deprecation tags are not needed because the class is already deprecated. But the individual tags make the deprecation much more obvious in IDEs, because the methods are highlighted as deprecated. They are not without these tags.

[Bug: T98290](https://phabricator.wikimedia.org/T98290)


JeroenDeDauw:
Damn, should not have merged this. Now you'll get a deprecation notice when you do

```php
$item = new Item();
$item->copy();
```

Ping @Benestar

PR #516, opened 2015-07-24
Remove services that have been moved to DataModel Services

JeroenDeDauw:
Note that this removes `StatementList::getBestStatementPerProperty` as it depends on `BestStatementsFinder`. I did not have that in mind when I started this commit. We could also just keep `BestStatementsFinder` in this component rather than removing it and the method it uses. Ping @Benestar


Benestar:
I'm OK with removing that method from StatementList as it has not much additional value. I also wonder why we introduced BestStatementsFinder because in all cases I know we can simply use StatementList::getByProperty::getBestStatenents. Perhaps we should kill that service completely, I don't know if it is used anywhere.

Benestar:
I think we should merge #514 first and do a 3.1 release before merging this. What's your opinion?

JeroenDeDauw:
Well, you should know why it was added, since you did that in f695626c8063a018ca4f80599116767fdb00c64c :)

I don't see any usages of the thing in WB.git. So perhaps we can just remove it altogether. In any case, that supports removing the associated method as done in this PR already.

> I think we should merge #514 first and do a 3.1 release before merging this. What's your opinion?

That makes little sense to me, as it can be done just as well after merging this, and seems rather unrelated in general. That being said, what about my comment on #513? I was expecting you to do something there but nothing has happened yet.

Benestar:
Well, we'd need to work on another branch for 3.1 and rebase the pull request on top of that.

I already made some commits in Wikibase.git to get rid of methods like copy or isEmpty but didn't finish that work yet. Maybe you can check how many usages are still left as I'm on vacation currently and work with my phone ;-) I think I fixed most usages in WB.git but we have to look into the other components as well.

JeroenDeDauw:
> I also wonder why we introduced BestStatementsFinder because in all cases I know we can simply use StatementList::getByProperty::getBestStatenents.

Well, now one needs to do this:

```php
private function getBestStatementsPerProperty( StatementList $statements ) {
		$bestStatements = new StatementList();

		foreach ( $statements->getPropertyIds() as $propertyId ) {
			foreach ( $statements->getByPropertyId( $propertyId )->getBestStatements() as $statement ) {
				$bestStatements->addStatement( $statement );
			}
		}

		return $bestStatements;
	}
```

Perhaps we ought to have that in StatementList

thiemowmde:
This probably never was a good approach because it requires the full list of possibly thousands of statements fully deserialized and loaded at once. Look at what we recently did in the StatementDataUpdater classes instead.

PR #518, opened 2015-07-28
Remove Entity::copy

Benestar:
Not used in Wikibase.git anymore, some usages in Wikibase DataModel Services left, which I'm going to remove as well.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,11 @@
 # Wikibase DataModel release notes
 
+## Version 5.0 (dev)
+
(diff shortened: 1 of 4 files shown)


JeroenDeDauw:
Er... I suggest to revert this as it forced master into 5.0 alpha, and to just merge this once we want to make 5.0. We don't win anything by merging it before anyway.

Benestar:
Can I somehow reopen this?

JeroenDeDauw:
Yeah well, it's a bit silly you have to do the double revert thing yes...

PR #519, opened 2015-07-29
Make constructor signature of DerivedPropertyValueSnak match parent

Benestar:
This allows using the short constructor in tests etc.


Benestar:
Should this be a 4.0.1 release?

addshore:
could go into a 4.0.1 if you want i guess!

JeroenDeDauw:
I'm wondering what the rationale is for also supporting `EntityId`. It's in the derivative as well, though is there for legacy reasons that do not affect the new class as far as I can tell. Keep in mind that constructors, unlike other methods, are not constrained by the LSP.

JeroenDeDauw:
Since this is not a bugfix, it should be 4.1.0, not 4.0.1.

PR #520, opened 2015-07-29
Remove abstract methods from Entity

Benestar:


JeroenDeDauw:
Both these methods are sill used in Wikibase.git on something of type `Entity`. Let's not remove those without at least having a viable alternative proposed.

Fun fact: I actually already removed these two at some point, though then found out when we tried to make the release that they needed to be re-added.

Benestar:
Redundant with #539

Issue #283, opened 2014-11-16
Remove patch and getDiff from Entity


JeroenDeDauw:
This is part of #193

JeroenDeDauw:
Still some usages left, mainly in `EditEntity`, though also in `EntityChangeFactory` and `EntityContent`

JeroenDeDauw:
Done with #516

Issue #281, opened 2014-11-16
Remove deprecated term methods from Entity


JeroenDeDauw:
This is part of #193

PR #521, opened 2015-07-29
Move id and fingerprint to Item and Property

Benestar:


addshore:
Seems very odd to have all of the calls to getFingerprint() in entity, when getFingerprint() is not in entity...

Benestar:
Perhaps I should just remove the whole Entity class in one PR? :O

JeroenDeDauw:
Agree with @addshore. You'd need to leave an abstract method behind.

It's unclear to me that this actually helps us much. I think it's better to just continue removing the methods as soon as we can, and once they are all gone, remove Entity. There still are #282 and #281. And apparently we closed #283 yesterday.

addshore:
Also needs a rebase ;)

Benestar:
> Seems very odd to have all of the calls to getFingerprint() in entity, when getFingerprint() is not in entity...

I just realized that this isn't an issue. As `Entity` implements `FingerprintProvider` we have the guarantee that the `getFingerprint` method exists. Therefore, this isn't even a breaking change

JeroenDeDauw:
Ah indeed. Well, then I'm not opposed to making this change, even though I'm not convinced it helps us much.

src/Entity/Item.php
```diff
@@ -77,6 +98,24 @@ public function setId( $id ) {
 	}
 
 	/**
+	 * @since 5.0
```
JeroenDeDauw: This is deceptive. While the method might not have been defined in this file, it was already in the class by means of inheritance.

JeroenDeDauw:
You can remove the `Fingerprint` and `Term` imports from `Entity` with this change

src/Entity/Item.php
```diff
@@ -56,6 +66,17 @@ public function __construct(
 	}
 
 	/**
+	 * Returns the id of the entity or null if it does not have one.
+	 *
+	 * @since 5.0
```
JeroenDeDauw: This is also deceptive, even though in this case the type hint changed. However, since it was made more specific, and since the more specific version would also have worked for the old code, there is no reason for users of this method to note that any change was made here.

src/Entity/Item.php
Benestar: Will update the `@since` tags

Benestar:
I just saw that `setFingerprint` isn't defined in any interfaces. That method should therefore be declared abstract or we create an interface called `FingerprintHolder` which adds the setter method.

PR #522, opened 2015-07-29
Remove methods from Entity undeprecated in subclasses

Benestar:
These methods can be safely removed as they are reimplemented in all subclasses


JeroenDeDauw:
Hold on... what about all the `$entity->setLabel()` and similar calls? Those will no longer be resolved now...

PR #525, opened 2015-07-29
Add filter method to StatementList

### LINKED ISSUE #505: RFC: filter support for collections
We could implement filter support for our collections. I got this idea by looking at [Ardent](https://github.com/morrisonlevi/Ardent).

This means adding a filter method to StatementList and other collections where we find need for it.

You can then do things such as these without needing to write a foreach:

```php
$referencedStatements = $statements->filter( function( Statement $statement ) {
    return !$statement->getReferences->isEmpty();
} );
```

This approach is more OOP / high level than the more imperative foreach.

Implementation is simple:

```php
public function filter( callable $filter ) {
		$statementList = new self();

		foreach ( $this->statements as $statement ) {
			if ( call_user_func( $filter, $statement ) ) {
				$statementList->addStatement( $statement );
			}
		}
		
		return $statementList;
	}
```

Example usecase for this in Wikibase.git is explained in this commit message: https://gerrit.wikimedia.org/r/#/c/219041/

JeroenDeDauw:
This fixes #505

### DIFF
--- a/src/Statement/ReferencedStatementFilter.php
+++ b/src/Statement/ReferencedStatementFilter.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Wikibase\DataModel\Statement;
+
+/**
+ * @since 4.1
+ *
+ * @licence GNU GPL v2+
+ * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ */
+class ReferencedStatementFilter implements StatementFilter {
+
+	/**
+	 * @param Statement $statement
+	 *
+	 * @return boolean
+	 */
+	public function statementMatches( Statement $statement ) {
+		return count( $statement->getReferences() ) !== 0;
+	}
+
+}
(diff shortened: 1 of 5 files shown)


src/Statement/ReferencedStatementFilter.php
```diff
@@ -0,0 +1,22 @@
...
+ */
+class ReferencedStatementFilter implements StatementFilter {
+
+	/**
+	 * @param Statement $statement
+	 *
+	 * @return boolean
+	 */
+	public function statementMatches( Statement $statement ) {
+		return count( $statement->getReferences() ) !== 0;
```
Benestar: I'd suggest `> 0` but that's a question of taste

JeroenDeDauw: Go for it!

Benestar:
Is this used already anywhere? Otherwise I wonder why we introduced this functionality...

thiemowmde:
I talked about this in https://github.com/wmde/WikibaseDataModel/issues/505#issuecomment-112961746 and I think we will barely use this. Possibly in 1 place, possibly never. But it's so simple and obvious that it can't hurt.

PR #529, opened 2015-07-30
Add FingerprintHolder interface

Benestar:
This interface is the same as `StatementListHolder` for `StatementListProvider`. We might also consider to add the setter to `FingerprintProvider` itself but I guess the same argumentation as for `StatementListHolder` will happen and we should stay consistent here.


Benestar:
We should also discuss if this pattern should be applied for all interfaces introduced in #513

JeroenDeDauw:
Still not entirely convinced this approach is good. +1

addshore:
needs rebase

Benestar:
@addshore rebased ;-)

PR #531, opened 2015-08-03
Remove misleading deprecation tag

JeroenDeDauw:
While Entity is deprecated, usage of this method is not.
This is entirely valid, yet is now seen as deprecated:

```php
$item = new Item();
$item->copy();
```

### DIFF
--- a/src/Entity/Entity.php
+++ b/src/Entity/Entity.php
@@ -306,7 +306,6 @@ abstract class Entity implements \Comparable, FingerprintProvider, EntityDocumen
 	 * Returns a deep copy of the entity.
 	 *
 	 * @since 0.1
-	 * @deprecated since 1.0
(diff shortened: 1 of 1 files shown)


Benestar:
What's the benefit of using this over

```php
$item = new Item();
unserialize( serialize( $item ) );
```

I think that piece of code makes it more obvious that we do a deep copy and rely on PHP's serialization functions (which might be a problem as you pointed out once).

JeroenDeDauw:
Well, perhaps we ought to deprecate it, though we have not done that so far. So the tag is wrong. I'm not entirely convinced we should remove this, as it's not causing any issues, though this discussion really belongs in a PR that does actual deprecation, rather than here, where a mistake in the docs is corrected.

PR #535, opened 2015-08-07
Run coverage generation as part of the test command

JeroenDeDauw:
This is to detect incorrect @covers tags before the code gets to ScutinizerCI. We are already doing this at https://github.com/wmde/WikibaseDataModelServices/blob/master/composer.json

A better approach would be nice though, considering this slows down the execution of the test command by doing a lot of not needed work. Apparently this is not going to come from the PHPUnit side though, as per https://github.com/sebastianbergmann/phpunit/issues/1760

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,7 +47,7 @@
 	"scripts": {
 		"test": [
 			"composer validate --no-interaction",
-			"phpunit"
+			"phpunit --coverage-text=/dev/null"
 		],
 		"cs": [
 			"composer phpcs",


thiemowmde:
-2, sorry. The slowdown on my machine is **significant**, Instead of ~700 ms it takes 7 seconds.

JeroenDeDauw:
Is 7 second thats unreasonable for the things you need to run before pushing a commit? (If so, how can you stand working on Wikibase.git?) You can still run the PHPUnit tests themselves while coding.

thiemowmde:
I can barely "stand working on Wikibase.git". I do not run the repo, client and lib tests on my machine on a regular basis. Usually I run only the single test I'm touching and then count on Jenkins to complain if I broke something. I do not like this situation. Over the past months I spend several hours benchmarking the Wikibase.git tests and already found and fixed some bottlenecks.

addshore:
> I can barely "stand working on Wikibase.git"

Maybe we should split it into the 3 repos it belongs in ? ;)

JeroenDeDauw:
Alternative provided by #572

thiemowmde:
* `phpunit`: 0m1.195s
* `phpunit --coverage-text=/dev/null`: 0m8.521s
* `phpunit --coverage-php /dev/null`: 0m6.983s

PR #539, opened 2015-08-10
Remove Entity and EntityTest

### LINKED ISSUE #193: Remove deprecated Entity
~~This should be done in DataModel 2.0~~

Benestar:
This removes the whole `Entity` class and makes it an alias of `EntityDocument`

Fixes #193


src/Entity/Item.php
```diff
@@ -21,7 +23,7 @@
  * @licence GNU GPL v2+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
  */
-class Item extends Entity implements StatementListHolder {
+class Item implements Comparable, EntityDocument, FingerprintHolder, StatementListHolder {
```
JeroenDeDauw: That is very... alphabetical? :) Think EntityDocument should be first

Benestar: Indeed :) where should Comparable go then? To the end? Honestly this shouldn't matter a lot ;)

JeroenDeDauw: Does not matter a lot yes. However putting the least important one before the by far most important one is kinda odd.

JeroenDeDauw:
Somewhat surprised by you submitting this, given my comment on #520. As mentioned there, I think we should keep these deprecated methods as long as they are still used. Even if no code breaks with a change such as this, we will still end up with having lots of unresolved method calls. That also leaves us with no easy way of finding usages of deprecated methods, as there are existing unresolved method calls that have nothing to do with deprecation.

thiemowmde:
Breaking changes, breaking changes everywhere

Benestar:
@JeroenDeDauw wrote
> Somewhat surprised by you submitting this, given my comment on #520. As mentioned there, I think we should keep these deprecated methods as long as they are still used. Even if no code breaks with a change such as this, we will still end up with having lots of unresolved method calls. That also leaves us with no easy way of finding usages of deprecated methods, as there are existing unresolved method calls that have nothing to do with deprecation.

Sorry that I didn't answer your comment on the other pull request. As far as I can see both methods have only one usage left in Wikibase.git so we should be able to get rid of that quite easily. I wouldn't consider that a blocker tbh.

JeroenDeDauw:
This PR is removing more than those two methods though, including those still used more than once...

Benestar:
I will have a look again on the still remaining usages though since @addshore removed the serialization stuff I think all usages have gone.

JeroenDeDauw:
Usage search including WB.git, WikibasQuality and WikibaseQualityConstraints and their dependencies (except DM itself)

* removeLabel: 1
* removeDescription: 1
* getAliases: 3
* getAllAliases: 3
* addAliases: 11
* removeAliases: 2
* getDescriptions: 5
* getLabels: 11
* getDescription: 6
* getLabel: 7
* setLabels: 2
* setDescriptions: 1
* setAllAliases: 0
* copy: 20
* getClaims: 4
* isEmpty: 9
* clear: 2
* setLabel & setDescription: bleh, hard to find out since derivative is undeprecated

thiemowmde:
I think I used this argument before in an other pull request: The fact that this deprecated class does exist does not block us from implementing better solutions that do not need it. Just do not use it.

thiemowmde:
I tried again to review this patch, but it does way to many things in my opinion. Please split or close this.

PR #542, opened 2015-08-11
Stop using the Claim alias

thiemowmde:
In production Claim is an alias, so there is no real difference between Claim and Statement. But the IDE and certain tools think Claim is a subclass of Statement because of Aliases.php. All code using the deprecated Claims class calls it with Statement objects, not with Claim objects. The tools think we are passing a superclass when only a subclass is allowed.

This is the most minimal fix to avoid these warnings.

### DIFF
--- a/src/Claim/Claims.php
+++ b/src/Claim/Claims.php
@@ -5,6 +5,7 @@ namespace Wikibase\DataModel\Claim;
 use ArrayObject;
 use InvalidArgumentException;
 use Traversable;
+use Wikibase\DataModel\Statement\Statement;
 
 /**
  * A claim (identified using it's GUID) can only be added once.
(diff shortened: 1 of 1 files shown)


src/Claim/Claims.php
```diff
@@ -58,12 +59,12 @@ private function getGuidKey( $guid ) {
 	}
 
 	/**
-	 * @param Claim $claim
+	 * @param Statement $claim
```
JeroenDeDauw: You do realize you've -1'd stuff several times because it changed the type and not the var name?

thiemowmde: What's the purpose of this comment other than making me upset? Do you want me to change the variable names? I will happily do this. Just tell me. But please do not play games.

PR #544, opened 2015-08-18
Avoid calling Statement's $claim

thiemowmde:
I did a fulltext search for `Statement $claim`. See https://gerrit.wikimedia.org/r/#/c/231249/ for the same patch in the Wikibase.git master.


JeroenDeDauw:
I suppose these where not changed because the class is called Claims and it has methods such as `addClaim`

JeroenDeDauw:
+1

PR #545, opened 2015-08-21
More specific @param ArrayIterator PHPDocs

thiemowmde:
`getIterator` comes from the `IteratorAggregate` interface. It is documented as `@return Traversable` there. The most simple thing is to copy-paste this line. *But* why not be honest and use the *actual* type? These classes are *guaranteed* to return this type.

### DIFF
--- a/src/Entity/ItemIdSet.php
+++ b/src/Entity/ItemIdSet.php
@@ -7,7 +7,6 @@ use Comparable;
 use Countable;
 use InvalidArgumentException;
 use IteratorAggregate;
-use Traversable;
 
 /**
  * Immutable set of ItemId objects. Unordered and unique.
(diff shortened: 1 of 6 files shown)


JeroenDeDauw:
Huh - did you not do the reverse someone else yesterday or so?

Iterator is used as type hint to indicate users should not rely on it being a specific implementation.

thiemowmde:
Not sure how this is relevant for this pull request. `getIterator` is in the `IteratorAggregate` interface. It's marked as `@return Traversable`. The reason for this is what you just explained.

This patch touches concrete implementations, not abstract interfaces. All these implementations are guaranteed to return these types. Why should we hide that?

This fixes an actual issue: The return values are used in `count` calls, for example. This is not possible with the `Iterator` interface. It's wrong to call `count` on an object that implements `Iterator`. But it works on `Callable`, which is in `ArrayIterator`.

JeroenDeDauw:
If you document it as ArrayIterator, then users can start relying on that by doing things such as passing it to count. Which means you no longer can change the iterator type. What if you want to change the thing internally to be lazy and return a generator? That's one example of something that would no longer be possible since internal details (the array) leaked out.

Of course, if there is more benefit in exposing ArrayIterator, then that can certainly outweigh the encapsulation difference. I'm not seeing any clear such benefit. Doing `count( $list->getIterator() )` instead of `count( $list )` is odd. I suspect you can come up with cases where it'd be handy to have less encapsulation. The same goes for method visibility. YAGNI

PR #546, opened 2015-08-25
Add workaround type hints to getIterator implementations

thiemowmde:
Note that we already did this in https://github.com/wmde/WikibaseDataModel/blob/master/src/SiteLinkList.php#L102. The "perfect" type hint would be `Traversable<ItemId>`, but PHP does not support this and IDEs don't understand it and get confused. The combination of `Traversable|ItemId[]` is a workaround to help IDEs understand what type the elements in a foreach loop are.

### DIFF
--- a/src/Entity/ItemIdSet.php
+++ b/src/Entity/ItemIdSet.php
@@ -2,6 +2,7 @@
 
 namespace Wikibase\DataModel\Entity;
 
+use ArrayIterator;
 use Comparable;
 use Countable;
 use InvalidArgumentException;
@@ -51,10 +52,10 @@ class ItemIdSet implements IteratorAggregate, Countable, Comparable {
 	/**
 	 * @see IteratorAggregate::getIterator
 	 *
-	 * @return Traversable
+	 * @return Traversable|ItemId[]
 	 */
 	public function getIterator() {
-		return new \ArrayIterator( $this->ids );
+		return new ArrayIterator( $this->ids );
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


JeroenDeDauw:
I agree with this.

Though before merging I'd like to know if you are aware of that I have done this in the past and got strong objection from someone on the basis that these type hints are wrong (as `foo[]` means `array` and not `Traversable`)? That's a big part of the reason the `toArray` methods where added. And in case you are aware of this, what do you think about that argument?

thiemowmde:
Personally I do not like `foreach ( $statementList->toArray() ... )`. I mean, what's the point in making classes `Traversable` if this feature is not used?

Code like `foreach ( $statementList as $statement )` does need `@var Statement[] $statementList` or `@var Statement $statement` as a workaround. There is no way around this in PHP 5, not even the type hints in this patch help. But they give a hint and don't hurt much. Usually `getIterator` is not called explicitly. If it is, the worst thing that can happen is that somebody tries to use the iterator in array functions without calling `iterator_to_array` first. I hope the name `getIterator` and the dual `Traversable|foo[]` do give strong hints why this must fail.

I would object a change that *removes* the `Traversable` (or `Iterator` or `ArrayIterator`) type hint and leaves an exclusive `foo[]`.

*Side note:* Out of curiosity I did a micro benchmark and found that `foreach ( $statementList->toArray() ... )` is about 20% faster than `foreach ( $statementList as ... )`.

PR #547, opened 2015-08-26
Add isEmpty to EntityDocument

Benestar:
This is part of the RFC on phabricator: https://phabricator.wikimedia.org/T110177

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,9 @@
 # Wikibase DataModel release notes
 
+## Version 4.3 (dev)
+
+* Added `isEmpty` to `EntityDocument`
+
 ## Version 4.2 (2015-08-26)
 
 * Added `EntityRedirect`
--- a/src/Entity/Entity.php
+++ b/src/Entity/Entity.php
@@ -323,17 +323,6 @@ abstract class Entity implements \Comparable, FingerprintHolder, EntityDocument 
 	}
 
 	/**
-	 * Returns if the Entity has no content.
-	 * Having an id set does not count as having content.
-	 *
-	 * @since 0.1
-	 * @deprecated since 1.0
-	 *
-	 * @return bool
-	 */
-	public abstract function isEmpty();
-
-	/**
 	 * Removes all content from the Entity.
 	 * The id is not part of the content.
 	 *
--- a/src/Entity/EntityDocument.php
+++ b/src/Entity/EntityDocument.php
@@ -43,7 +43,7 @@ interface EntityDocument {
 	public function setId( $id );
 
 	/**
-	 * Returns if the entity has no content.
+	 * Returns true if the entity has no content.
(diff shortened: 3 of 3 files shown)


thiemowmde:
I can not think of a reason why an entity type should *not* have an isEmpty method. Basically: Can you think of an entity type where it's not possible to tell if it's empty or not? In a worst case scenario it can simply return true all the time. So +2 from my side, but I want to wait for opinions of other team members.

JeroenDeDauw:
Y u create new issue? ;p There already was one https://phabricator.wikimedia.org/T109581

Benestar:
@JeroenDeDauw that is only about isEmpty and not about the other methods. Should've linked it though.

JeroenDeDauw:
@Benestar so is this PR right? ;p

JeroenDeDauw:
Strictly speaking this is a breaking change. However, this is one of those cases where it might be more pragmatic to deviate a bit.

Benestar:
I think via the comments here and on [phabricator](https://phabricator.wikimedia.org/T110177) this can be merged safely.

JeroenDeDauw:
As this is a breaking change for derivatives of EntityDocument, can someone check nothing will blow up when putting this in a minor release? I think we might have test doubles in some other component that implement the interface.

Benestar:
I also don't object if we wait until other changes and do a 5.0. This really isn't an urgent request.

JeroenDeDauw:
If nothing in our codebase will actually break, I think for this case it's fine to release as 4.x. Even if something breaks, and we'll be able to deal with it quickly. Can someone check? I'm not able to ATM

JeroenDeDauw:
@brightbyte did you do the verification?

thiemowmde:
Subclasss of EntityDocument in my local code base:
* Entity, Item and Property.
* [Wikibase\DataModel\Fixtures\EntityOfUnknownType](https://github.com/wmde/WikibaseDataModel/blob/master/tests/fixtures/EntityOfUnknownType.php)
* [Wikibase\DataModel\Services\Fixtures\EntityOfUnknownType](https://github.com/wmde/WikibaseDataModelServices/blob/master/tests/fixtures/EntityOfUnknownType.php)
* [Wikibase\DataModel\Services\Fixtures\FakeEntityDocument](https://github.com/wmde/WikibaseDataModelServices/blob/master/tests/fixtures/FakeEntityDocument.php)

JeroenDeDauw:
Thanks @thiemowmde.

That means DMS's tests will blow up when we tag this. Actually, they already did as we where testing with a dev version of DM as well. @Benestar fixed this with https://github.com/wmde/WikibaseDataModelServices/commit/66df0d70eac52bca51272ec96964f643af38227b, though that is ignoring the issue rather than addressing it properly. Based on that I suggest we tag ASAP and fix the DMS tests.

PR #550, opened 2015-09-01
Remove unused fixture class

Benestar:
This class is unused in this package and can be easily re-added if needed. (Also re-added in DM Services)


thiemowmde:
+2. As far as I can see this became unused the moment EntityDiffer and -Patcher got removed from this component.

JeroenDeDauw:
Was also broken after #547

PR #551, opened 2015-09-01
4.3.0 release

JeroenDeDauw:

PR #553, opened 2015-09-02
4.3.0 release

JeroenDeDauw:
Resubmission of #551

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,6 +1,6 @@
 # Wikibase DataModel release notes
 
-## Version 4.3.0 (dev)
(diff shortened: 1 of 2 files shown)


thiemowmde:
What's wrong with #551?

JeroenDeDauw:
I included stuff that should not be in there and there was a merge conflict

thiemowmde:
I think I'm trying to understand this workflow for some months now. A rebase takes a minute. Resolving the 1-line conflict takes another minute. A rebase does not create the impression that something went wrong and an other approach is needed. Does not create the impression that a new full review is needed. Does not send more emails to reviewers.

PR #555, opened 2015-09-02
Use tarballs rather than clones

JeroenDeDauw:
They are faster, esp now we are using the cache. Originally we switched to clones since the tarballs failed sometimes, though this should already be mitigated by us now using `travis_retry`


addshore:
Is the cache working then? ;)

Benestar:
Yay for speediness

PR #556, opened 2015-09-04
Remove deprecated Claims class

### LINKED ISSUE #157: Deprecate and remove Claims class

Benestar:
Fixes #157 
See https://phabricator.wikimedia.org/T78281
See https://gerrit.wikimedia.org/r/#/c/235684/


thiemowmde:
Please split this patch into two: One that removes all remaining usages of this class, the second removes the class.

JeroenDeDauw:
No need to split if we're fine with none of this being merged before we switch master to 5.x-dev. Given there is only a single small change in a test method that gets delayed like this, I suspect that is the case.

thiemowmde:
Rebased. Please note that this is the same as/part of #557.

thiemowmde:
I did a rebase with no other change. @Benestar, feel free to merge this. +2 from my side.

JeroenDeDauw:
Finally fully rid of this class \o/

PR #557, opened 2015-09-04
Remove getClaims/setClaims from all Entity classes

Benestar:
This depends on #556 and #539

### DIFF
--- a/src/Entity/Item.php
+++ b/src/Entity/Item.php
@@ -6,7 +6,6 @@ use InvalidArgumentException;
 use OutOfBoundsException;
 use Wikibase\DataModel\SiteLink;
 use Wikibase\DataModel\SiteLinkList;
-use Wikibase\DataModel\Statement\Statement;
 use Wikibase\DataModel\Statement\StatementList;
 use Wikibase\DataModel\Statement\StatementListHolder;
 use Wikibase\DataModel\Term\Fingerprint;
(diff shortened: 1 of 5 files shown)


thiemowmde:
How does this depend on `Entity` being removed? This shouldn't matter.

thiemowmde:
As already asked in #556, please makes this patch independent from the other so the removal of the `setClaims` methods can be reviewed and merged without being forced to remove the class at the same time. I'm still not sure if it's worth the trouble to remove it. As I already said many times: The fact that something exists is not a problem in itself. It's deprecated. The problem is when it's used, which can be avoided. Not using something is not blocked on removing it. Just don't use it.

thiemowmde:
Rebased.

src/Entity/Item.php
```diff
@@ -303,24 +302,6 @@ public function setStatements( StatementList $statements ) {
 	}
 
 	/**
-	 * @deprecated since 1.0, use getStatements()->toArray() instead.
-	 *
-	 * @return Statement[]
-	 */
-	public function getClaims() {
```
thiemowmde: This is unrelated to both commits in this pull request. Is this breaking change really necessary?

src/Entity/Property.php
```diff
@@ -254,22 +253,4 @@ public function setStatements( StatementList $statements ) {
 		$this->statements = $statements;
 	}
 
-	/**
-	 * @deprecated since 1.0, use getStatements()->toArray() instead.
-	 *
-	 * @return Statement[]
-	 */
-	public function getClaims() {
```
thiemowmde: As above.

thiemowmde:
+2 from my side. @Benestar, feel free to merge this if it's ok for you.

During rebase I removed the first commit (it's still in #556). I change this patch so it only touches the Item and Property classes, and removes both getClaims and setClaims. The later is also removed in #556, so whatever patch is merged first, the other needs to be rebased. I suggest to merge this first.

JeroenDeDauw:
Needs rebase

Benestar:
Rebased and merged per Thiemo's +2

PR #558, opened 2015-09-10
Replace Claims with ArrayObject in tests

thiemowmde:
These are the only remaining usages of the Claims class (besides `Item::setClaims` and `Property::setClaims`). I'm aware that these classes are also touched in other patches. The difference is that this patch here is non-breaking. I will rebase the other patches when this is merged.

[Bug: T78281](https://phabricator.wikimedia.org/T78281)


Benestar:
Not sure what the benefit of this is as there are already patches basically doing the sameis but certainly an improvement.

PR #559, opened 2015-09-29
Add ReferenceList::isEmpty

thiemowmde:
Motivated by https://gerrit.wikimedia.org/r/#/c/242084/1/view/src/ClaimHtmlGenerator.php.

The second test with the iterator stuff is because an other valid implementation of an isEmpty method could use `!$this->valid()` (this is done in the HashArray class) but would reset the iterator.


JeroenDeDauw:
hah, guess that CI failure is my fault :) will fix it later today

PR #560, opened 2015-10-01
Fix throws tags with multiple exceptions

thiemowmde:


mariushoch:
Since when are we doing it like that? I see by far more places where we have two `@throws` tags than I see places where we do this.

thiemowmde:
I'm pretty sure you are confusing the two styles.

PR #562, opened 2015-10-08
Add a StatementRank class

thiemowmde:


Jonas-WMDE:
+1

JeroenDeDauw:
Did you consider having an actual Rank object, and if so, what made you go with the static approach?

src/Statement/Rank.php
```diff
@@ -0,0 +1,93 @@
...
+	const DEPRECATED = 0;
+	const NORMAL = 1;
+	const PREFERRED = 2;
+
+	/**
+	 * TODO: Add doc.
+	 */
+	const RANK_0 = self::DEPRECATED;
+	const RANK_1 = self::NORMAL;
+	const RANK_2 = self::PREFERRED;
```
Benestar: What do we need this for?

Benestar:
Thanks for sharing your proposal, we definitely need an improvement how we handle ranks in our data model. However, I think we could avoid introducing so many methods by just relying on the fact that the constant integers of the rank values replicate the order of the ranks (which is in fact what the class does).

What I think would be more appropriate would be a static method `Statement::getAvailableRanks` which lets us check if some value is a valid rank. All the comparison stuff can be done using the integer values of the constants.

JeroenDeDauw:
> we definitely need an improvement how we handle ranks in our data model.

Can you elaborate on what you currently find inadequate?

thiemowmde:
> Did you consider having an actual Rank object

Yes, and I find the idea of an object wrapping nothing but an integer pointless and scary. It also doesn't help solving the issue I want to solve here: The semantics of how ranks are compared is currently duplicated across our code base. This would only be solved if we ditch the idea of ranks being integers completely, across all components. This would be a major breaking change that's not worth it, in my opinion. Therefore: Ranks stay integers and a helper class with some comparison methods is good enough.

> just relying on the fact that the constant integers of the rank values replicate the order of the ranks

That's exactly the issue we have: This knowledge is not in a single place but duplicated across the code base. This caused heavy confusion yesterday, but nobody was able to point to that single place that says "ranks are always, now and in the future, guaranteed to be in increasing order from least to most preferred".

> which is in fact what the class does

Yes, it does create an explicit place for this knowledge. That's the point.

> getAvailableRanks

That's a possible enhancement, but doesn't solve any issue.

> All the comparison stuff can be done using the integer values of the constants.

As said above, yes, it could, but no, it shouldn't.

src/Statement/Rank.php
thiemowmde: This is just a suggestion for a possible enhancement. You can use these alternative constants in a context where it is critical to know that ranks are integers.

JeroenDeDauw:
> This caused heavy confusion yesterday, but nobody was able to point to that single place that says "ranks are always, now and in the future, guaranteed to be in increasing order from least to most preferred".

That is somewhat surprising to me. Ranks are parts of statements. So if there is no Rank enum, then that is where I'd first look. And if that is too unclear, I'd just add a Rank enum.

thiemowmde:
How would you do that? There are no enums in PHP.

JeroenDeDauw:
I meant having a Dedicated class to hold the constants rather than putting them into Statement.

JeroenDeDauw:
Thinking about it, if we create a dedicated class for ranks, regardless of what we put in there, Rank(s) seems like a bad name. The concept here is not disjoint from statements, so the name ought to reflect that, by being like StatementRank(s).

thiemowmde:
`StatementRank` is indeed better, will change.

src/Statement/Rank.php
Benestar: In those cases just use `0` instead of `Rank::RANK_0` as you hardcode the zero anyways.

Benestar:
> This caused heavy confusion yesterday, but nobody was able to point to that single place that says "ranks are always, now and in the future, guaranteed to be in increasing order from least to most preferred".

I wasn't part of that discussion or meeting when this arised so could you summarize what happened? The rank to integer mapping seems to be quite clear to me as there is only one place where the integer values are defined (inside Statement). The only thing that is missing imo is a static place where one can get the list of all available ranks, which can easily be solved by adding a method to Statement. That should btw also answer Jeroen's question.

> Can you elaborate on what you currently find inadequate?

We have several places where we hardcode `array(deprecated, normal, preferred)` in our code and it would be nice if that could be centralized. Beyond that, I don't see the need for another class and therefore more complexity.

src/Statement/Rank.php
thiemowmde: No, what? What is "hardcoded anyway"? I'm not sure you get the point of all this. Constants are for easy tracking and refactoring. A rank should never be expresses via an integer constant.

thiemowmde:
> could you summarize what happened?

What I just explained: There is no single place that encodes the knowledge what "best" means, when a rank is "better" and so on.

> The only thing that is missing imo is a static place where one can get the list of all available ranks

I'm not sure how often I should repeat myself. Please read https://github.com/wmde/WikibaseDataModel/pull/562#issuecomment-146830690 again.

thiemowmde:
> We have several places where we hardcode array(deprecated, normal, preferred)

I can not find this array anywhere outside the Statement class. Some tests use arrays of ranks in data providers, but I do not understand how this is a mistake. Tests should be independent from implementations they are not testing.

src/Statement/StatementRank.php
```diff
@@ -0,0 +1,161 @@
...
+	 * @throws InvalidArgumentException
+	 * @return bool True if the given ranks are equal.
+	 */
+	public static function isEqual( $rank1, $rank2 ) {
+		return self::compare( $rank1, $rank2 ) === 0;
+	}
+
+	/**
+	 * @param int|null $rank1
+	 * @param int $rank2
```
thiemowmde: int|null

src/Statement/StatementRank.php
```diff
@@ -0,0 +1,161 @@
...
+	 *
+	 * @throws InvalidArgumentException
+	 * @return bool True if the first rank is less preferred than the second.
+	 */
+	public static function isLower( $rank1, $rank2 ) {
+		return self::compare( $rank1, $rank2 ) === -1;
+	}
+
+	/**
+	 * @param int $rank1
```
thiemowmde: int|null

src/Statement/StatementRank.php
```diff
@@ -0,0 +1,161 @@
...
+		}
+	}
+
+	/**
+	 * @param int[]|int $ranks
+	 * @param int [$rank2,...]
+	 *
+	 * @return int|null Best rank in the array or list of arguments, or null if none given.
+	 */
+	public static function findBestRank( $ranks = array() /*...*/ ) {
```
thiemowmde: I'm afraid a method with this signature is of not much use. How would you call it when you have a StatementList?

src/Statement/StatementRank.php
```diff
@@ -0,0 +1,161 @@
...
+		if ( !is_array( $ranks ) ) {
+			$ranks = func_get_args();
+		}
+
+		$best = null;
+
+		foreach ( $ranks as $rank ) {
+			if ( self::isHigher( $rank, $best ) ) {
+				$best = $rank;
```
thiemowmde: This should `break` when it reaches an optimum, which currently is `self::PREFERRED`.

JeroenDeDauw:
No changes since 2015. Please reopen if needed for current goals

JeroenDeDauw:
Also beware that the @since tags need updating

PR #563, opened 2015-10-13
Update CS rules

JeroenDeDauw:
This makes them match the already updated Wikibase DataModel Services ones.


thiemowmde:
Succeeds, so obviously good enough for now. But please note the issue I run into in https://github.com/wmde/WikibaseDataModelServices/pull/82.

PR #565, opened 2015-11-03
Switch to MediaWiki code sniffer 0.5.0

thiemowmde:
- I switched from PHP code sniffer to the most recent MediaWiki code sniffer 0.7.0.
- I removed all rules that are already enabled by default in the MediaWiki rule set.

Note that this is not a cosmetic patch. This enables a series of additional sniffs that only exist in the MediaWiki rule set and are currently not tested in this code repository.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,10 +28,10 @@
 		"wikimedia/assert": "~0.2.2"
 	},
 	"require-dev": {
-		"phpunit/phpunit": "~4.8",
+		"mediawiki/mediawiki-codesniffer": "~0.7.0",
 		"ockcyp/covers-validator": "~0.4.0",
-		"squizlabs/php_codesniffer": "~2.3",
-		"phpmd/phpmd": "~2.3"
+		"phpmd/phpmd": "~2.3",
+		"phpunit/phpunit": "~4.8"
 	},
 	"autoload": {
 		"files" : [
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -1,19 +1,14 @@
 <?xml version="1.0"?>
-<!--
-	- https://github.com/squizlabs/PHP_CodeSniffer/wiki/Annotated-ruleset.xml
-	- https://github.com/squizlabs/PHP_CodeSniffer/tree/master/CodeSniffer/Standards
--->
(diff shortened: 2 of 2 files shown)


JeroenDeDauw:
I'm not a big fan of the additional dependency. Same goes for the changed order of the visibility and abstract modifiers, as well as the `else` and `else if` stuff being put on a single line. I prefer to have it stay as is (though with the other style fixes applied of course).

thiemowmde:
@JeroenDeDauw, I reworked this patch and disabled the two rules responsible "for the changed order of the visibility and abstract modifiers, as well as the else and else if stuff being put on a single line."

Benestar:
+1 would like to merge the style fixes independently :)

JeroenDeDauw:
Now the style changes have been merged separately the only thing that is left is something I think should not be done. I don't like the dependence on the MW style thing, esp in non-MW PHP libraries.

thiemowmde:
The mediawiki/mediawiki-codesniffer library is as small as a library can be:
* It does have a single dependency: squizlabs/php_codesniffer.
* It contains 9 custom MediaWiki sniffs, all well tested.
* It contains a predefined, well maintained ruleset.

The advantage is that we do not have to maintain our own independent ruleset, when it is so close to the MediaWiki ruleset anyway. Instead we are maintaining the differences only. [We did the same in Wikibase](https://github.com/wikimedia/mediawiki-extensions-Wikibase/blob/master/phpcs.xml). We are disabling a few sniffs we do not like or do not care about and adding a ton more. But we do not maintain a duplicate any more.

Which of these aspects is it you do not like?

composer.json
```diff
@@ -27,8 +27,8 @@
 		"data-values/data-values": "~0.1|~1.0"
 	},
 	"require-dev": {
+		"mediawiki/mediawiki-codesniffer": "0.4.0",
```
JeroenDeDauw: ~

thiemowmde:
Will rebase after lunch.

JeroenDeDauw:
If we use specific sniffs (or groups thereof), then we don't have to comply with all the MW style things, so then I'm happy with this change.

PR #572, opened 2015-11-04
Validate covers tags

JeroenDeDauw:
Alternative to #535

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,6 +27,7 @@
 		"data-values/data-values": "~0.1|~1.0"
 	},
 	"require-dev": {
+		"ockcyp/covers-validator": "~0.2.0",
 		"squizlabs/php_codesniffer": "~2.3",
 		"phpmd/phpmd": "~2.3"
 	},
@@ -46,7 +47,8 @@
 	"scripts": {
 		"test": [
 			"composer validate --no-interaction",
-			"phpunit"
+			"phpunit",
+			"vendor/bin/covers-validator"
 		],
 		"cs": [
 			"composer phpcs",


thiemowmde:
Holy ...! I tried and found that this pulls an insane chain of dependencies:
```
  - Installing symfony/console (v2.7.6)
  - Installing sebastian/version (1.0.6)
  - Installing sebastian/global-state (1.1.1)
  - Installing sebastian/recursion-context (1.0.1)
  - Installing sebastian/exporter (1.2.1)
  - Installing sebastian/environment (1.3.2)
  - Installing sebastian/diff (1.3.0)
  - Installing sebastian/comparator (1.2.0)
  - Installing symfony/yaml (v2.7.6)
  - Installing doctrine/instantiator (1.0.5)
  - Installing phpdocumentor/reflection-docblock (2.0.4)
  - Installing phpspec/prophecy (v1.5.0)
  - Installing phpunit/php-text-template (1.2.1)
  - Installing phpunit/php-timer (1.0.7)
  - Installing phpunit/php-token-stream (1.4.8)
  - Installing phpunit/php-file-iterator (1.4.1)
  - Installing phpunit/phpunit-mock-objects (2.3.8)
  - Installing phpunit/php-code-coverage (2.2.4)
  - Installing phpunit/phpunit (4.8.16)
  - Installing ockcyp/covers-validator (v0.2.0)
```
Are you sure we want this? In #565 you used dependencies as one of your arguments. Therefore I would -1 this for now.

JeroenDeDauw:
I'm not happy with the dependencies and have filed an issue for this upstream. We only have a single direct dependency, so this is very different than us depending on all of them directly. A difference with 565 is that we do not already have this functionality.

thiemowmde:
Still an issue in my opinion because it slows down all composer updates, but doesn't give us much. It would be much better to have an actual coverage report. Broken `@covers` tag will be very visible there.

JeroenDeDauw:
Coverage report generation is very slow compared to just running the tests. (Unless you have horrible tests that is.)

> slows down all composer updates

Composer has a cache, so not really

thiemowmde:
I'm not sure how this is an argument. A proper coverage report is needed no matter what. It may be generated once a day. But it needs to be generated and it needs to be visible, just as the actual test results. With such a coverage report being way more helpful than this or #535, neither this nor #535 is needed.

JeroenDeDauw:
We have code coverage reports generated for each commit after merge. Of course this is helpful, though that is entirely besides the point of this PR. What it does is add a way to detect broken coverage tags before merge, so the coverage report generation which happens afterwards does not get broken.

JeroenDeDauw:
Using this in a dozen or so repos now. Including the WMDE fundraising one. Not had any issues with it and one is complaining about this.

thiemowmde:
* `phpunit`: 0m1.177s
* `phpunit --coverage-text=/dev/null`: 0m8.528s
* `phpunit --coverage-php /dev/null`: 0m7.009s
* `vendor/bin/covers-validator`: 0m0.946s

PR #576, opened 2015-11-27
Code experiment: apply facet pattern.

brightbyte:
The idea of "facets" as implemented here is also known as the "role object" or "extension" pattern.

This implementation uses a helper object, FacetManager, that is used via composition to provide facet support to other objects that implement FacetContainer.

Facet support is added to the following classes: Term, SiteLink, Statement, and SnakObject.

The following classes are converted to facets: DerivedPropertyValueSnak and TermFallback.

Performance implications were not considered for this experiment. The idea is to provide sample code as a basis for discussion.

[Bug: T118860](http://phabricator.wikimedia.org/T118860)


src/SiteLink.php
```diff
@@ -38,6 +42,11 @@ class SiteLink implements Comparable {
 	private $badges;
 
 	/**
+	 * @var FacetManager
+	 */
+	private $facetManager;
```
Benestar: Where would $facetManager be set to some value? As far as I see it will always be null and not usable in the current implementation of the methods. When would this object be instantiated?

RELEASE-NOTES.md
```diff
@@ -1,7 +1,11 @@
 # Wikibase DataModel release notes
 
-## Version 4.4.0 (dev)
-
+## Version 5 (dev)
```
JeroenDeDauw: No need to make breaking changes. `TermFallback` and `DerivedPropertyValueSnak` can be deprecated and scheduled for removal in the next breaking release.

src/Facet/FacetContainer.php
```diff
@@ -0,0 +1,55 @@
...
+	public function getFacet( $name, $type = null );
+
+	/**
+	 * Adds a facet object. Any facet object previously registered under the same name
+	 * is discarded.
+	 *
+	 * @param string $name
+	 * @param object $facetObject The facet object
+	 */
+	public function addFacet( $name, $facetObject );
```
Jonas-WMDE: Why not name it setFacet as there can only be one facet under the same name?

brightbyte: no strong feelings about that

JeroenDeDauw:
No strong objection against the concept, though no endorsement either, as I am not convinced an approach such as this makes sense.

Of course before any such thing is merged it needs to be reviewed beyond the conceptual level, which I have not yet done so. Please indicate when the time to proceed with that has arrived.

src/Facet/FacetContainer.php
```diff
@@ -0,0 +1,55 @@
...
+	 *
+	 * @return string[]
+	 */
+	public function listFacets();
+
+	/**
+	 * Returns a facet object.
+	 *
+	 * @param string $name
+	 * @param string|null $type The desired type
```
thiemowmde: 1. What "type"? We are using this term for a lot of different things. I honestly have no idea which one is meant here.
2. "Desired" how?
3. What is supposed to happen when this is null?

src/Facet/FacetContainer.php
```diff
@@ -0,0 +1,55 @@
...
+	 *
+	 * @throws NoSuchFacetException if no facet object is found for $name
+	 * @throws MismatchingFacetException if the facet object is not compatible with $type
+	 * @return object
+	 */
+	public function getFacet( $name, $type = null );
+
+	/**
+	 * Adds a facet object. Any facet object previously registered under the same name
+	 * is discarded.
```
thiemowmde: Should be a setter then.

src/Facet/FacetContainer.php
thiemowmde: Type hint against `object`?

src/Facet/FacetContainer.php
```diff
@@ -0,0 +1,55 @@
...
+	 * @param string|null $type The desired type
+	 *
+	 * @throws NoSuchFacetException if no facet object is found for $name
+	 * @throws MismatchingFacetException if the facet object is not compatible with $type
+	 * @return object
+	 */
+	public function getFacet( $name, $type = null );
+
+	/**
+	 * Adds a facet object. Any facet object previously registered under the same name
```
thiemowmde: I wonder why this interface does have a setter/adder at all? I expected the knowledge which facets are supported, how they are build and so on to be an implementation detail of the class that implements the interface. But with this public (!) setter everybody is free to add whatever object they want and call it a "facet". Why is this needed? Can't this be protected or omitted?

src/Facet/MismatchingFacetException.php
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * @licence GNU GPL v2+
+ * @author Daniel Kinzler
+ */
+class MismatchingFacetException extends RuntimeException {
+
+	/**
+	 * @param string $facetName
+	 */
+	public function __construct( $facetName, $expectedType ) {
```
thiemowmde: Type doc missing.

Can the type be anything but a string?

src/Internal/FacetManager.php
```diff
@@ -0,0 +1,80 @@
...
+	 * @return object
+	 */
+	public function getFacet( $name, $type = null ) {
+		if ( !isset( $this->facets[$name] ) ) {
+			throw new NoSuchFacetException( $name );
+		}
+
+		$facet = $this->facets[$name];
+
+		// TODO: if $facet is callable, call it, and replace $this->facets[$name] with the result.
```
thiemowmde: This is the first time I'm reading something about facets being callable. I assume this targets deferred deserialization. My issue with this comment is that it describes a very specific implementation. I suggest to leave this out as long as this implementation is not discussed.

src/Facet/FacetContainer.php
Benestar: @thiemowmde as far as I understand the facets should be added by services outside of this class/component and thus this setter needs to be public. However, I'm also concerned where we define which factes are available for which model. Currently this is flexible but also quite arbitrary.

src/Snak/DerivedValues.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
+namespace Wikibase\DataModel\Snak;
+
+use DataValues\DataValue;
+use InvalidArgumentException;
+
+/**
+ * Value objects for managing a set of derived DataValues.
```
thiemowmde: Imutable.

src/Snak/SnakObject.php
```diff
@@ -116,4 +125,46 @@ public function unserialize( $serialized ) {
...
+	/**
+	 * @param string $name
+	 * @param object $facetObject
+	 */
+	public function addFacet( $name, $facetObject ) {
+		if ( !$this->facetManager ) {
+			$this->facetManager = new FacetManager();
+		}
+
+		$this->facetManager->addFacet( $name, $facetObject );
```
thiemowmde: The whole implementation here is now copy-pasted several times, in several different classes. To me this screams "abstract base class!", but I do not know if we can agree on this.

src/Statement/Statement.php
```diff
@@ -298,4 +307,46 @@ public function equals( $target ) {
...
+	/**
+	 * @param string $name
+	 * @param object $facetObject
+	 */
+	public function addFacet( $name, $facetObject ) {
+		if ( !$this->facetManager ) {
+			$this->facetManager = new FacetManager();
+		}
+
+		$this->facetManager->addFacet( $name, $facetObject );
```
thiemowmde: Here it is the third time.

src/Term/Term.php
```diff
@@ -77,4 +86,46 @@ public function equals( $target ) {
...
+	/**
+	 * @param string $name
+	 * @param object $facetObject
+	 */
+	public function addFacet( $name, $facetObject ) {
+		if ( !$this->facetManager ) {
+			$this->facetManager = new FacetManager();
+		}
+
+		$this->facetManager->addFacet( $name, $facetObject );
```
thiemowmde: And again.

src/Snak/SnakObject.php
Benestar: Would this be a valid use case for [traits](http://php.net/manual/de/language.oop5.traits.php)?

src/Facet/FacetContainer.php
```diff
@@ -0,0 +1,55 @@
...
+	 * @return boolean
+	 */
+	public function hasFacet( $name );
+
+	/**
+	 * Returns a list of supported facet names.
+	 *
+	 * @return string[]
+	 */
+	public function listFacets();
```
adrianheine: What would this be used for?

src/Snak/SnakObject.php
thiemowmde: We need to switch to PHP 5.4 first before being able to use traits. And I remember hearing opinions that we should probably never use traits, because they do not solve the issues abstract base classes introduce.

JeroenDeDauw:
Since this is an experiment from 2015 with no updates since I'm assuming this can be closed.

PR #579, opened 2015-12-10
Filter out terms with empty string value in termlist

JeroenDeDauw:
This is for discussion~~; do not merge.~~

https://groups.google.com/a/wikimedia.de/d/topic/wikimedia-de-tech/NTb6EyOVMCE/discussion

[Bug: T121433](https://phabricator.wikimedia.org/T121433)

### DIFF
--- a/src/Term/TermList.php
+++ b/src/Term/TermList.php
@@ -36,7 +36,7 @@ class TermList implements Countable, IteratorAggregate, Comparable {
 				throw new InvalidArgumentException( 'Every element in $terms must be an instance of Term' );
 			}
 
-			$this->terms[$term->getLanguageCode()] = $term;
+			$this->setTerm( $term );
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


thiemowmde:
I like this very much. I believe this makes the life of all DataModel users easier.

src/Term/TermList.php
```diff
@@ -116,7 +116,12 @@ private function assertIsLanguageCode( $languageCode ) {
 	}
 
 	public function setTerm( Term $term ) {
```
thiemowmde: Please describe the new behaviour in the documentation of the constructor and this method.

Benestar:
I really like that implementation of my proposal and think it makes sense to do it this way. :+1:

brightbyte:
+2 from me. Not merged, since we haven't decided yet which pull requests should go into the next release.

PR #582, opened 2015-12-10
Name ClassLoader object $classLoader

thiemowmde:
Also see:
- https://github.com/wmde/WikibaseDataModelServices/pull/98
- https://github.com/wmde/WikibaseInternalSerialization/pull/89


addshore:
No strong opinion either way, but we do seem to have it called $classLoader in other places.

PR #583, opened 2015-12-10
Add EntityRedirect::__toString for easy debugging

thiemowmde:


Benestar:
Do we have __toString methods anywhere else? I'd suggest to first define generally if and how we want __soString to be defined in our datamodel and then implement it for all models we think it makes sense?

thiemowmde:
We have it in many cases where it makes sense, e.g. Entity::__toString. It probably doesn't make much sense for more complex classes like Fingerprint.

JeroenDeDauw:
@thiemowmde did you mean EntityId or something in your last comment? Else I don't get the remark on complexity, as Entity is more complex than Fingerprint

src/Entity/EntityRedirect.php
```diff
@@ -71,4 +71,8 @@ public function equals( $that ) {
 			&& $this->targetId->equals( $that->targetId );
 	}
 
+	function __toString() {
```
JeroenDeDauw: public function

JeroenDeDauw:
Needs test to detect regressions and be documented as a new thing in the packages public interface

PR #584, opened 2015-12-10
Remove SnakRole

thiemowmde:
I have never seen this before. Not used anywhere in my code base, including the major extensions we use. I have not checked for other external usages yet.


Benestar:
+1 for the next major release

JeroenDeDauw:
Used in Wikibase QueryEngine

Benestar:
Wikibase QueryEngine has been abandoned and thus not relevant here anymore.

JeroenDeDauw:
I have a fork that is not abandoned. Besides, relevance can come in the form of demonstrating something has a certain use, even if that something is not used in that manner at present.

PR #585, opened 2015-12-11
SnakRole is not an interface

thiemowmde:


JeroenDeDauw:
What does this gain us?

thiemowmde:
Nobody can implement this. It's not supposed to be used as an interface. The "private constructor" pattern is something I do not see very often, but it's what I have seen used for such cases in the past years.

PR #589, opened 2015-12-15
Use current lower limits for PHPCS and MD

thiemowmde:
This uses the current limits of the actual code base for all numeric PHPCS and PHPMD rules. This does have two effects. First, now the rule sets reflect the actual state of the code.

Second, whenever a commit introduces code that's more complex than what's currently in this component, we will be aware of that. In most cases we will simply raise the limit in the same commit. So this is not about blocking such commits. It should be allowed to raise these limits if it makes sense. This is about awareness.

Also see https://github.com/wmde/WikibaseDataModelServices/pull/101.

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -6,7 +6,7 @@
 
 	<rule ref="Generic.Files.LineLength">
 		<properties>
-			<property name="lineLimit" value="120" />
+			<property name="lineLimit" value="116" />
 		</properties>
 	</rule>
 
--- a/phpmd.xml
+++ b/phpmd.xml
@@ -8,10 +8,15 @@
         <exclude name="ExcessiveClassComplexity" />
         <exclude name="TooManyPublicMethods" />
     </rule>
+    <rule ref="rulesets/codesize.xml/ExcessiveClassComplexity">
+        <properties>
+            <property name="maximum" value="57" />
+        </properties>
+    </rule>
     <rule ref="rulesets/codesize.xml/TooManyPublicMethods">
         <properties>
             <property name="ignorepattern" value="(^(set|get|test))i" />
-            <property name="maxmethods" value="16" />
+            <property name="maxmethods" value="14" />
         </properties>
     </rule>
 


JeroenDeDauw:
+0

Slightly surprised by this given earlier argumentation regarding these kinds of restrictions

thiemowmde:
What I do not like are hard, binary limits that feel like they have been "picked by a dice role". With this patch none of the limits is arbitrary any more. They all have a single, simple reason: they reflect the current state of this component.

The critical bit is: Everybody should be allowed to raise these numbers if somebody happens to write code that hits a limit. We should be aware (that's what the failing CI will tell us) and have a second look if it makes sense to split and refactor such code. But if it's fine, we raise the limit and be done.

phpcs.xml
```diff
@@ -14,8 +14,8 @@
     <rule ref="Generic.Files.LineEndings" />
     <rule ref="Generic.Files.LineLength">
         <properties>
-            <property name="lineLimit" value="125" />
-            <property name="absoluteLineLimit" value="125" />
+            <property name="lineLimit" value="119" />
+            <property name="absoluteLineLimit" value="119" />
```
brightbyte: This limit feels very low for the absolute lines in a file. I would probably have gone fro 1000 or so.
Yes, any change that needs a larger file can simply bump this number. But how does the contributor know or discover this? Is it obvious in any way?

Benestar:
I'm not sure I can follow this argumentation. In my opinion we shouldn't try to keep the configuration numbers as low as possible but rather enter reasonable and sophisticated limits. By saying that these limits just reflect the status quo and can be increased if needed we change this from a rule set to a piece of statistics about the code base. Some limits may not be perfect but generally spoken this document supports our understanding of good code and should not be altered on demand.

phpcs.xml
thiemowmde: Every other number but the current maximum of 119 would be arbitrary, and that's what I do not like and want to avoid, as explained. Discovering this is not totally obvious (the output will not tell you that you "should increase a number in the phpcs.xml file" or something), but it should not be that hard to understand what's going on. PHPCS output is pretty good. Much better than PHPMD, which doesn't even tell you the name of the rule you hit. :-(

thiemowmde:
The ultimate goal for the line length limit is 100. As long as we are not there, the least we should do is to limit is as much as possible.

For all other, so called "complexity" metrics, I strongly oppose every arbitrary limit. Either work with the maximum of the actual code, or disable the rule. The later tells contributors "you happened to hit an arbitrary limit, better not ask where it comes from, there is no actual reason we picked this specific number". There is no way to discuss this. The former tells contributors "your contribution currently is the most complex code in this repository, are you sure?" This is easy to discuss.

phpcs.xml
Benestar: That naming is really confusing. I also thought of numbers of lines in a file immediately :S Btw do we also have a limit for that number?

Benestar:
Thanks Thiemo for the explanation, I now understand your reasoning and mostly follow it. I agree that it doesn't make sense to set arbitrary limits and that we cannot discuss if 9 or 10 should be the limit of complexity. The only thing we can discuss or agree on is that lower is better. So +1 from my side.

I still think that at some point we should have global limits on all repositories though that would need a lot of work to be done in Wikibase.git. :S

JeroenDeDauw:
@Benestar global limits or "one size fits all" is not a good idea IMO. There are differences between codebases and sometimes for good reason. Plus going though all repos and reducing line lengths or similar to satisfy some such limit seems like a poor use of time for anyone that has actual stuff to get done.

Benestar:
Hmm, I see your point Jeroen. My intention was rather that we should have a uniform code style (not only formatting but also best practice) on our repositories but as mentioned that is a effort hardly worth the result.

thiemowmde:
I agree with Jeroen that an enforced unified ruleset is not a good idea. This is especially true for metrics like the ones touched in this patch. This doesn't mean we should not have a shared base, as proposed in #565.

Checks for number of lines [exist in PHPMD](http://phpmd.org/rules/codesize.html#excessiveclasslength).

PR #593, opened 2016-01-05
Do not allow self-redirects

thiemowmde:
I consider this a bug. But note that this is a pretty critical breaking change too. We must make sure no self-redirect exists before we deploy this on wikidata.org. Otherwise bad things happen, the worst being that you can not even revert edits that turned an entity into a self-redirect.

### DIFF
--- a/src/Entity/EntityRedirect.php
+++ b/src/Entity/EntityRedirect.php
@@ -37,6 +37,10 @@ class EntityRedirect {
 			);
 		}
 
+		if ( $entityId->equals( $targetId ) ) {
+			throw new InvalidArgumentException( 'An entity can not redirect to itself' );
+		}
+
 		$this->entityId = $entityId;
 		$this->targetId = $targetId;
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
I would not class this as a bug

Benestar:
> We must make sure no self-redirect exists before we deploy this on wikidata.org.

Are you going to take care of this? +1 for the patch itself.

thiemowmde:
Discussed with @brightbyte and agreed that this is not a bug. Self-redirects should be forbidden, but not here on the data model level. This would cause major pain in way to many places. For example: Think of a redirect that was done in the wrong direction on accident. One way to fix such an issue is to move the page, then revert the edit that made the entity a redirect and then make the other page a redirect. This would fail hard right away in the first step.

Self-redirect should be detected and blocked, but only when you create a redirect, not when you do something else that involves a redirect.

PR #597, opened 2016-01-19
More tests for ItemId/PropertyId::unserialize

thiemowmde:
I find it odd that this method can be misused to inject unsupported stuff into the object. But I do not think this is relevant. As far as I know we are not using this code in production.

I'm adding tests to make this behaviour more obvious. If you think this should be fixed somehow, please tell me, and I will create an other patch. However, please merge this patch as it is because this here does not count as a change, while an actual fix will be a breaking change.

### DIFF
--- a/tests/unit/Entity/ItemIdTest.php
+++ b/tests/unit/Entity/ItemIdTest.php
@@ -84,10 +84,27 @@ class ItemIdTest extends PHPUnit_Framework_TestCase {
 		$this->assertSame( '["item","Q1"]', $id->serialize() );
 	}
 
-	public function testUnserialize() {
+	/**
+	 * @dataProvider serializationProvider
+	 */
+	public function testUnserialize( $json, $expected ) {
 		$id = new ItemId( 'Q1' );
-		$id->unserialize( '["item","Q2"]' );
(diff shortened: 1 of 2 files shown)


tests/unit/Entity/ItemIdTest.php
```diff
@@ -84,10 +84,25 @@ public function testSerialize() {
...
+	 */
+	public function testUnserialize( $json, $expected ) {
 		$id = new ItemId( 'Q1' );
-		$id->unserialize( '["item","Q2"]' );
-		$this->assertSame( 'Q2', $id->getSerialization() );
+		$id->unserialize( $json );
+		$this->assertSame( $expected, $id->getSerialization() );
+	}
+
+	public function jsonProvider() {
```
JeroenDeDauw: serializationProvider? The JSON is an implementation detail.

thiemowmde: There is already an idSerializationProvider above. I wanted to avoid confusion. A possible alternative is phpSerializationProvider, because this is about PHP's Serialization interface. Is this ok for you?

JeroenDeDauw: yes

PR #598, opened 2016-01-19
Add since tags to EntityId::PATTERN constants

thiemowmde:


JeroenDeDauw:
I don't like the increase in size of the public interface and don't think the "duplication removal" here justifies it. "duplication removal" since this essentially just trades string constants with PHP constants. Also note that Item and Property already have such constants.

thiemowmde:
I wanted to avoid ItemId pointing to Item. Or do you think it is ok to have some `Property::…` in PropertyId?

JeroenDeDauw:
I think it is indeed good to avoid this, and that in general it is not worthwhile to bind to such a constant for these particular strings, as they will not change anyway. They are part of our public interface, and most users can't even use the PHP constants to begin with.

thiemowmde:
I agree that code outside of this component can and should use the string constants, e.g. `'item'`. But inside of this component? The fact that the string `'item'` is part of our public interface and can not change is not touched by this patch. This is purely to make the code easier to read and maintain.

JeroenDeDauw:
Sure. To me it does neither of those things

thiemowmde:
I reworked this patch completely.

PR #600, opened 2016-01-20
Simplify ReferenceList::unserialize

thiemowmde:
Follow up to  #597.

Originally, this patch changed `ItemId::unserialize` and `PropertyId::unserialize` to throw exceptions and call the constructor. But I was not sure if …

1. … this is relevant in production. As far as I can tell this is unused.
2. … this should not be done because these methods should be as fast as possible and not do any checks.

I reworked this patch completely because I realized these `unserialize` methods are not intended to be called manually. The should only be called internally when using PHP's `unserialize( $string )`. This should be as fast as possible.

### DIFF
--- a/src/Entity/EntityIdValue.php
+++ b/src/Entity/EntityIdValue.php
@@ -52,12 +52,12 @@ class EntityIdValue extends DataValueObject {
 	 *
 	 * @since 0.5
 	 *
-	 * @param string $value
+	 * @param string $serialized
 	 *
 	 * @throws IllegalValueException
 	 */
-	public function unserialize( $value ) {
-		list( $entityType, $numericId ) = json_decode( $value );
+	public function unserialize( $serialized ) {
+		list( $entityType, $numericId ) = json_decode( $serialized );
 
 		try {
 			$entityId = LegacyIdInterpreter::newIdFromTypeAndNumber( $entityType, $numericId );
--- a/src/ReferenceList.php
+++ b/src/ReferenceList.php
@@ -217,10 +217,10 @@ class ReferenceList implements Comparable, Countable, IteratorAggregate, Seriali
 	 *
 	 * @since 2.1
 	 *
-	 * @param string $data
+	 * @param string $serialized
 	 */
-	public function unserialize( $data ) {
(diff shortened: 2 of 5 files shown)


JeroenDeDauw:
I'm not sure what threat model this works against. The only way to you can something there that does not match the ItemId contract is by writing to the database somehow bypassing the PHP. If someone can do that, we have bigger problems on our hands.

If you do want to do an extra check here, then just call the constructor.

thiemowmde:
As I said in the commit message, this is unused. This patch not about fixing an actual injection vector, but about clean code.

thiemowmde:
I reworked this patch completely, see the commit message.

src/Entity/EntityIdValue.php
```diff
@@ -29,10 +29,7 @@ public function __construct( EntityId $entityId ) {
 	 * @return string
 	 */
 	public function serialize() {
-		return json_encode( array(
-			$this->entityId->getEntityType(),
-			$this->getNumericId()
-		) );
+		return json_encode( array( $this->entityId->getEntityType(), $this->getNumericId() ) );
```
JeroenDeDauw: I find this form less easy to read

JeroenDeDauw:
+1

src/Entity/EntityIdValue.php
thiemowmde: Reverted.

src/ReferenceList.php
```diff
@@ -217,10 +217,10 @@ public function serialize() {
 	 *
 	 * @since 2.1
 	 *
-	 * @param string $data
+	 * @param string $serialized
 	 */
-	public function unserialize( $data ) {
-		$this->__construct( unserialize( $data ) );
+	public function unserialize( $serialized ) {
+		$this->references = unserialize( $serialized );
```
mariushoch: Constructor does validation, this doesn't … not sure that's desired

thiemowmde: Yes, that's the intention of this patch. `unserialize` must not do validation that was already done when the original object was created. Note that PHP only calls this method with serialization strings that are clearly marked as `ReferenceList` serializations.

PR #605, opened 2016-02-06
Remove set/remove labels/descriptions from Entity

Benestar:
This patch removes the deprecated methods setLabels, setDescriptions,
getLabels and getDescriptions from Entity.

The reason I want to have this in 5.0 is because we can then start to implement `LabelsProvider` and `DescriptionsProvider` in `Item` and `Property` directly and get more and more independent from `Fingerprint`. This is needed for new entity types that only support parts of a `Fingerprint`.

I checked in Wikibase.git and dependencies and found only two usages of these methods in the serialization component that are trivial to fix.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -5,6 +5,7 @@
 * Removed `Claims` class (deprecated since 1.0)
 * Removed `Entity::getClaims` (deprecated since 1.0)
 * Removed `HashableObjectStorage`
+* Removed `getLabels`, `getDescriptions`, `setLabels` and `setDescriptions` from `Entity`
 
 ## Version 4.4.0 (2016-01-20)
 
--- a/src/Entity/Entity.php
+++ b/src/Entity/Entity.php
@@ -162,34 +162,6 @@ abstract class Entity implements Comparable, FingerprintHolder, EntityDocument {
 	}
 
 	/**
-	 * Returns the descriptions of the entity in the provided languages.
-	 *
-	 * @deprecated since 0.7.3 - use getFingerprint and setFingerprint
-	 *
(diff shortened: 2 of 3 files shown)


JeroenDeDauw:
Rebased

JeroenDeDauw:
+1.5, assuming this will indeed not make it very difficult to upgrade to the new version. Not merging as I think it's good if someone currently working on the Wikibase.it code also sees this.

JeroenDeDauw:
Needs rebase

thiemowmde:
I find it odd that this removes the methods for labels and descriptions only, but not the same for aliases.

Some of the removed methods are still used in https://github.com/wmde/WikibaseInternalSerialization/. Please prepare a patch for this.

In general this patch does not give us much, mostly because the whole class is deprecated anyway.

-1 for now because of all these reasons combined.

Benestar:
> The reason I want to have this in 5.0 is because we can then start to implement `LabelsProvider` and `DescriptionsProvider` in `Item` and `Property` directly and get more and more independent from `Fingerprint`. This is needed for new entity types that only support parts of a `Fingerprint`.

@thiemowmde do you have any answer to how we can achieve this then? We need to get rid of typechecks againts `FingerprintProvider` and move to `LabelsProvider` etc. That is not possible if there are still conflicting methods in `Item` and `Property`.

thiemowmde:
Why should there be a release with a LabelProvider and a DescriptionProvider, but no AliasesProvider? How is this helpful?

What about #506 ([T87237](https://phabricator.wikimedia.org/T87237))?

Benestar:
`AliasesProvider` doesn't conflict with existing methods... I tried to make the breaking changes as minimal as possible.

PR #607, opened 2016-02-06
Remove HashableObjectStorage

Benestar:
No breaking changes to ReferenceList as all inherited methods got inlined
into that class.

This class is not used outside of the data model component and the removal won't break any existing components we currently have.


JeroenDeDauw:
Rebased

PR #608, opened 2016-02-06
Get rid of unused removeDuplicates method

Benestar:
There is no single usage of this method in our whole code base.
Furthermore, if no duplicates should be allowed, a different data
structure should be used.

PR #603, opened 2016-02-04
Release 5.0.0

thiemowmde:
[Bug: T125636](https://phabricator.wikimedia.org/T125636)

PR #609, opened 2016-02-06
Refactor ReferenceList

Benestar:
This removes the inheritance between ReferenceList and SplObjectStorage
and replaces it with a native array. To make the change as minimal as
possible, the Countable interface and the IteratorAggregate interface are
now implemented by ReferenceList so that it can be used almost the same as
before (like iterating over it and calling count() on it).

### DIFF
--- a/src/ReferenceList.php
+++ b/src/ReferenceList.php
@@ -2,10 +2,11 @@
 
 namespace Wikibase\DataModel;
 
+use ArrayIterator;
 use Comparable;
-use Hashable;
+use Countable;
 use InvalidArgumentException;
-use SplObjectStorage;
+use IteratorAggregate;
 use Traversable;
 use Wikibase\DataModel\Internal\MapValueHasher;
 use Wikibase\DataModel\Snak\Snak;
@@ -13,13 +14,9 @@ use Wikibase\DataModel\Snak\Snak;
 /**
  * List of Reference objects.
  *
- * Note that this implementation is based on SplObjectStorage and
- * is not enforcing the type of objects set via it's native methods.
- * Therefore one can add non-Reference-implementing objects when
- * not sticking to the methods of the References interface.
- *
  * @since 0.1
  * Does not implement References anymore since 2.0
+ * Does not extend SplObjectStorage since 5.0
  *
  * @licence GNU GPL v2+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
@@ -27,7 +24,12 @@ use Wikibase\DataModel\Snak\Snak;
  * @author Thiemo Mättig
  * @author Bene* < benestar.wikimedia@gmail.com >
  */
-class ReferenceList extends SplObjectStorage implements Comparable {
+class ReferenceList implements Comparable, Countable, IteratorAggregate {
+
+	/**
+	 * @var Reference[]
+	 */
+	private $references = array();
 
 	/**
 	 * @param Reference[]|Traversable $references
@@ -50,6 +52,7 @@ class ReferenceList extends SplObjectStorage implements Comparable {
 
 	/**
 	 * Adds the provided reference to the list.
+	 * Empty references are ignored.
 	 *
 	 * @since 0.1
 	 *
@@ -59,27 +62,19 @@ class ReferenceList extends SplObjectStorage implements Comparable {
 	 * @throws InvalidArgumentException
 	 */
 	public function addReference( Reference $reference, $index = null ) {
-		if ( !is_int( $index ) && $index !== null ) {
-			throw new InvalidArgumentException( '$index must be an integer or null' );
+		if ( $index !== null && ( !is_int( $index ) || $index < 0 ) ) {
+			throw new InvalidArgumentException( '$index must be a non-negative integer or null' );
 		}
 
-		if ( $index === null || $index >= count( $this ) ) {
+		if ( $reference->isEmpty() ) {
+			return;
+		}
+
+		if ( $index === null || $index >= count( $this->references ) ) {
 			// Append object to the end of the reference list.
-			$this->attach( $reference );
+			$this->references[] = $reference;
 		} else {
 			$this->insertReferenceAtIndex( $reference, $index );
-		}
-	}
-
-	/**
-	 * @see SplObjectStorage::attach
-	 *
-	 * @param Reference $reference
-	 * @param mixed $data Unused in the ReferenceList class.
-	 */
-	public function attach( $reference, $data = null ) {
-		if ( !$reference->isEmpty() ) {
-			parent::attach( $reference, $data );
 		}
 	}
 
@@ -104,26 +99,7 @@ class ReferenceList extends SplObjectStorage implements Comparable {
 	 * @param int $index
 	 */
 	private function insertReferenceAtIndex( Reference $reference, $index ) {
-		$referencesToShift = array();
-		$i = 0;
-
-		// Determine the references that need to be shifted and detach them:
-		foreach ( $this as $object ) {
-			if ( $i++ >= $index ) {
-				$referencesToShift[] = $object;
-			}
-		}
-
-		foreach ( $referencesToShift as $object ) {
-			$this->detach( $object );
-		}
-
-		// Attach the new reference and reattach the previously detached references:
-		$this->attach( $reference );
-
-		foreach ( $referencesToShift as $object ) {
-			$this->attach( $object );
-		}
+		array_splice( $this->references, $index, 0, array( $reference ) );
 	}
 
 	/**
@@ -136,8 +112,7 @@ class ReferenceList extends SplObjectStorage implements Comparable {
 	 * @return boolean
 	 */
 	public function hasReference( Reference $reference ) {
-		return $this->contains( $reference )
-			|| $this->hasReferenceHash( $reference->getHash() );
+		return $this->hasReferenceHash( $reference->getHash() );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
Needs rebase

JeroenDeDauw:
Includes #608

JeroenDeDauw:
Rebased

src/ReferenceList.php
```diff
@@ -104,26 +99,7 @@ public function addNewReference( $snaks = array() /*...*/ ) {
...
-			$this->detach( $object );
-		}
-
-		// Attach the new reference and reattach the previously detached references:
-		$this->attach( $reference );
-
-		foreach ( $referencesToShift as $object ) {
-			$this->attach( $object );
-		}
+		array_splice( $this->references, $index, 0, array( $reference ) );
```
JeroenDeDauw: Hah, this is a lot nicer!

src/ReferenceList.php
```diff
@@ -286,4 +263,35 @@ public function equals( $target ) {
...
 	}
 
+	/**
+	 * @see Countable::count
+	 *
+	 * @since 5.0
+	 *
+	 * @return int
+	 */
+	public function count() {
```
JeroenDeDauw: This one is not new

JeroenDeDauw:
But but but! `SplObjectStorage` is more efficient!

+2, when follow up is merged

Benestar:
> But but but! SplObjectStorage is more efficient!

Will make a huge difference for our immense lists of references ;p

src/ReferenceList.php
```diff
@@ -150,13 +125,10 @@ public function hasReference( Reference $reference ) {
 	 * @return int|boolean
 	 */
 	public function indexOf( Reference $reference ) {
-		$index = 0;
-
-		foreach ( $this as $object ) {
-			if ( $object === $reference ) {
+		foreach ( $this->references as $index => $ref ) {
+			if ( $ref->equals( $reference ) ) {
```
thiemowmde: Is this breaking change (from identity to equality) documented?

src/ReferenceList.php
```diff
@@ -59,31 +62,23 @@ public function __construct( $references = array() ) {
 	 * @throws InvalidArgumentException
 	 */
 	public function addReference( Reference $reference, $index = null ) {
-		if ( !is_int( $index ) && $index !== null ) {
-			throw new InvalidArgumentException( '$index must be an integer or null' );
+		if ( $index !== null && ( !is_int( $index ) || $index < 0 ) ) {
```
thiemowmde: An other breaking change that must be documented. There should also be a test for this.

src/ReferenceList.php
```diff
@@ -59,31 +62,23 @@ public function __construct( $references = array() ) {
...
 		}
 	}
 
 	/**
-	 * @see SplObjectStorage::attach
-	 *
-	 * @param Reference $reference
-	 * @param mixed $data Unused in the ReferenceList class.
-	 */
-	public function attach( $reference, $data = null ) {
```
thiemowmde: In my opinion all removed methods (this also includes `contains`, `getHash` and more) should be listed in the release notes.

src/ReferenceList.php
```diff
@@ -50,6 +52,7 @@ public function __construct( $references = array() ) {
 
 	/**
 	 * Adds the provided reference to the list.
+	 * Empty references are ignored.
```
thiemowmde: Missing in the release notes.

src/ReferenceList.php
```diff
@@ -286,4 +263,33 @@ public function equals( $target ) {
...
+	}
+
+	/**
+	 * @see IteratorAggregate::getIterator
+	 *
+	 * @since 5.0
+	 *
+	 * @return Traversable
+	 */
+	public function getIterator() {
```
thiemowmde: Release notes? Test? Why was this added, in a "refactoring" patch?

src/ReferenceList.php
```diff
@@ -286,4 +263,33 @@ public function equals( $target ) {
...
+		return new ArrayIterator( $this->references );
+	}
+
+	/**
+	 * @since 5.0
+	 *
+	 * @return Reference[]
+	 */
+	public function toArray() {
```
thiemowmde: Release notes? Test? Why was this added, in a "refactoring" patch?

src/ReferenceList.php
```diff
@@ -286,4 +263,33 @@ public function equals( $target ) {
...
+	 * @return Traversable
+	 */
+	public function getIterator() {
+		return new ArrayIterator( $this->references );
+	}
+
+	/**
+	 * @since 5.0
+	 *
+	 * @return Reference[]
```
thiemowmde: "Numerically indexed (non-sparse) array."

src/ReferenceList.php
```diff
@@ -194,15 +166,23 @@ public function hasReferenceHash( $referenceHash ) {
...
+				$this->removeReferenceAtIndex( $index );
+			}
 		}
 	}
 
 	/**
-	 * Returns the reference with the provided hash, or null if there is no such reference in the list.
+	 * @param int $index
+	 */
+	private function removeReferenceAtIndex( $index ) {
```
thiemowmde: Typically I find single-line methods like this one confusing and not helpful. Did you planned to make this public?

src/ReferenceList.php
```diff
@@ -211,12 +191,9 @@ public function removeReferenceHash( $referenceHash ) {
...
 	 */
 	public function getReference( $referenceHash ) {
-		/**
-		 * @var Hashable $hashable
-		 */
-		foreach ( $this as $hashable ) {
-			if ( $hashable->getHash() === $referenceHash ) {
-				return $hashable;
+		foreach ( $this->references as $reference ) {
+			if ( $reference->getHash() === $referenceHash ) {
```
thiemowmde: There are two loops in this class that search for a hash, but no method to access elements by index. Therefore it would probably be more efficient (less code, maybe faster) to index the array by hash, and not numerically.

tests/unit/ReferenceListTest.php
```diff
@@ -372,14 +372,4 @@ public function testGivenNonEmptyList_isNotEmpty() {
 		$this->assertFalse( $references->isEmpty() );
 	}
 
-	public function testGivenNonEmptyListWithForwardedIterator_isNotEmpty() {
-		$references = new ReferenceList();
-		$references->addNewReference( new PropertyNoValueSnak( 1 ) );
-		$references->next();
```
thiemowmde: `next` is not removed, as far as I can see. Why is this test removed then?

Benestar: Afaik `next` got removed? /me is confused

src/ReferenceList.php
Benestar: I don't think it makes sense to list all the methods `SplObjectStorage` contains (http://php.net/manual/en/class.splobjectstorage.php). Isn't it enough to note that `ReferenceList` no longer extends that class?

src/ReferenceList.php
Benestar: Good point, we can do this in another patch.

src/ReferenceList.php
Benestar: Intention is that we don't break loops that iterate over a `ReferenceList`. `foreach ( $referenceList as $reference ) {` should still work with this.

src/ReferenceList.php
Benestar: Done

src/ReferenceList.php
Benestar: Fixed

tests/unit/ReferenceListTest.php
thiemowmde: `next` is in the `Iterator` interface, which is not removed, because `ReferenceList` still `implements IteratorAggregate`.

Benestar: `IteratorAggregate` does not extend the `Iterator` interface but only `Traversable`. (`IteratorAggregate extends Traversable`) I tested it manually and I cannot call a `next` method on `ReferenceList. (`Call to undefined method Wikibase\DataModel\ReferenceList::next()`)

src/ReferenceList.php
thiemowmde: That's exactly what I said, why I feel so lost and what I consider the top one blocking issue in releasing this: Without detailed release notes it's extremely hard, if not impossible to update code that uses this class.

Also see my comments in #603.

Benestar: So you propose to add the complete list of functions in `SplObjectStorage` to the release notes? That might indeed help to see which methods exactly got removed from the public interface of `ReferenceList`.

Your latest comment? I didn't tell Jeroen to merge that patch. Should I add a "DNM until three persons have seen this" or "DNM for 3 hours" in the description of my pull requests?

tests/unit/ReferenceListTest.php
thiemowmde: Oh, wow. Is this an other line missing in the release notes? "ReferenceList does not implement Iterator any more". Including "ReferenceList does not implement current, next, key, valid and rewind any more" to make it absolutely clear what this means.

Benestar: See https://github.com/wmde/WikibaseDataModel/pull/609/files#r52314875

In release notes we say "ReferenceList no longer derives from SplObjectStorage, though still implements Countable and Traversable" which implies it does no longer implement the other interfaces of SplObjectStorage. Also please avoid comments like "Oh, wow" which are kind of disrespectful. We are not entirely stupid...

src/ReferenceList.php
thiemowmde: > add the complete list of functions in SplObjectStorage to the release notes?

No. Why should one do that? List methods that can not be called any more.

tests/unit/ReferenceListTest.php
thiemowmde: Code typically does not type hint against SplObjectStorage or Countable or Traversable. Sure, all this should be mentioned in the release notes, but is barely helpful.

"Oh, wow" was meant as "oh, wow, PHP is broken". I reviewed the changes that touch this code a long, long time ago and I'm looking at them again for almost a full day now, and I still don't fully understand what's going on. There is a reason I never merged these patches. Verbose, crystal clear documentation is absolutely crucial here.

PR #614, opened 2016-02-07
Remove clear from Entity, Item and Property

Benestar:
New instances should be created instead of clearing an existing Entity.
The only real usecase for this seems to be the clear=true parameter in the
wbeditentity api module.

Property::clear even had a bug in that it didn't clear the statements but
only the fingerprint.

### DIFF
--- a/src/Entity/Item.php
+++ b/src/Entity/Item.php
@@ -311,18 +311,6 @@ class Item implements EntityDocument, FingerprintHolder, StatementListHolder,
 	}
 
 	/**
-	 * Removes all content from the Item.
-	 * The id is not part of the content.
-	 *
-	 * @since 0.1
-	 */
-	public function clear() {
-		$this->fingerprint = new Fingerprint();
-		$this->siteLinks = new SiteLinkList();
-		$this->statements = new StatementList();
-	}
-
-	/**
 	 * @since 1.0
 	 *
 	 * @return StatementList
--- a/src/Entity/Property.php
+++ b/src/Entity/Property.php
@@ -264,17 +264,6 @@ class Property implements EntityDocument, FingerprintHolder, StatementListHolder
 	}
 
 	/**
-	 * Removes all content from the Property.
-	 * The id and the type are not part of the content.
-	 *
-	 * @since 0.1
-	 */
-	public function clear() {
-		$this->fingerprint = new Fingerprint();
-		$this->statements = new StatementList();
-	}
-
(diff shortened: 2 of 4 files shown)


JeroenDeDauw:
IIRC that use case will need a replacement that can create an "empty" entity of the same type. We have no code that can do this currently. Could have something like the `clear` method in `EntityDocument` (though then create and return a new instance), or some service in DM Services. Neither is nice.

Perhaps we should double check if we actually care about this use case? If the property one was broken, then perhaps we could just have it support items, and not have to deal with this in a more general way?

Benestar:
We just got rid fo that single use case: https://gerrit.wikimedia.org/r/#/c/268970/2/repo/includes/api/EditEntity.php

It is not really solved yet but I don't like the idea of adding something like this to EntityDocument. Rather create some abstract method in EntityContent or EntityHandler to do that.

JeroenDeDauw:
In that case, why not remove the method already?

Benestar:
Agree, we can also just remove it. Will change this PR accordingly.

JeroenDeDauw:
+2 if someone else is at least +1.

thiemowmde:
* Version 0.7.4 deprecated `Entity::clear`, "to be removed in 1.0".
* It was even removed between 08cd23c (2014-04-25) and e24aaae (2014-06-07), which was released as 0.8.0 and 0.8.1.
* Version 0.8.2 undeprecated `Entity::clear`.
* Version 1.0.0 deprecated it again, but this was not mentioned in the release notes.
Did somebody noticed `Property::clear` is bogus and does not do what it claims to do? See https://github.com/wmde/WikibaseDataModel/blob/master/src/Entity/Property.php#L232.

We discussed changes like this many, many times. Again, what's the benefit of removing methods instead of simply **not using** them? There is no simple replacement, as Jeroen correctly said.

-2.

thiemowmde:
Why is this tagged as a bug?

JeroenDeDauw:
> Property::clear even had a bug in that it didn't clear the statements but
only the fingerprint.

thiemowmde:
Rebased.

JeroenDeDauw:
@thiemowmde still -2?

thiemowmde:
Let's please do this in a later release (probably 6.0) together with the removal of Entity.

thiemowmde:
Rebased. +2 from my side. Waiting for the tests.

PR #615, opened 2016-02-08
Add equals to EntityDocument

Benestar:
This method is needed in EntityDocument to be able to compare two entities
where their type is not known. Entity now also doesn't implement
Comparable anymore.

The signature of the equals method has changed from mixed param type to
typehinting against EntityDocument.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -9,6 +9,8 @@
 * Removed `ReferenceList::removeDuplicates`
 * `ReferenceList` no longer derives from `SplObjectStorage`, though still implements `Countable`
 * Removed `HashableObjectStorage`
+* Added `EntityDocument::equals`
+* `Entity` no longer implements `Comparable`
 
 #### Other changes
 
--- a/src/Entity/Entity.php
+++ b/src/Entity/Entity.php
@@ -2,8 +2,6 @@
 
 namespace Wikibase\DataModel\Entity;
 
-use Comparable;
-use Wikibase\DataModel\Statement\Statement;
 use Wikibase\DataModel\Term\AliasGroup;
 use Wikibase\DataModel\Term\AliasGroupList;
 use Wikibase\DataModel\Term\FingerprintHolder;
@@ -19,7 +17,7 @@ use Wikibase\DataModel\Term\TermList;
  * @licence GNU GPL v2+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
  */
-abstract class Entity implements Comparable, FingerprintHolder, EntityDocument {
+abstract class Entity implements FingerprintHolder, EntityDocument {
 
 	/**
 	 * Sets the value for the label in a certain value.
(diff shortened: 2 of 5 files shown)


brightbyte:
+1 from me. I think this makes more sense than using the overly generic and oddly named Comparable interface. I'd merge this, but I'd like others to have a chance to look at this. Poke me if nothing happens.

JeroenDeDauw:
> oddly named Comparable interface

Do you even interfaces? ;)

thiemowmde:
-2. I don't get it. An `equals` method that fails hard when called with something that is not equal instead of returning false is horrifying. What was the original intent of this patch? Why does it change so many other things? For what reason? What's the benefit of this painful and unnecessary breaking change? How can it be that three (3!) people don't see this?

We have 29 (!) `equals` functions in our combined code base. None (!) of them does have a type hint. There is a single (!) one with a type hint in `PageEntityUsages`.

Where is the Phabricator ticket for this? Who put this in our current sprint? Who decided this is in the milestone I wanted to release today?

JeroenDeDauw:
Creating sprint items for refactorings is generally considered to be an anti-pattern.

I was somewhat sceptical about this change as well, but could not find any concrete reason of why not to do this. @thiemowmde do you have any such concrete reasons? If we made a mistake we can still easily correct it.

thiemowmde:
The Comparable interfaces is designed as it is for a reason. An equals method should never fail, not with an exception and not with a type error.
```php
private function isEqual( EntityDocument $a, $b ) {
    if ( $b instanceof EntityDocument ) {
        return $a->equals( $b );
    } else {
        return false;
    }
}
```
Really, what's the benefit of the added type hint? How does it make it easier or more safe to use the method? Will somebody now go around and do the same to all other equals methods we have in our code base?

Benestar:
Sorry for not sharing this publicly. I have discussed this with @brightbyte in irc along with some other things:

> &lt;benestar> DanielK_WMDE: should it then only accept EntityDocument params or still "mixed" as type?
> &lt;DanielK_WMDE> benestar:  only EntityDocument. To compare with anything else makes no sense.
> &lt;DanielK_WMDE> another reason not to use the generic interface
> (...)
> &lt;DanielK_WMDE> benestar: addign the type hint would only break code that is already broken.

thiemowmde:
> To compare with anything else makes no sense.

"No sense" to whom? Under which circumstances? With which preconditions and additional knowledge? Really, I don't get it. What's wrong with an `equals` method that simply returns false when something is not equal? Why should something so straightforward ever fail hard with a type error? What's wrong with, for example, a `StatementList::equals` that also accepts a `Statement[]` array and returns true if that contains the same statements? That would be impossible with the proposed design pattern.

Benestar:
To answer the last bit, equals should be an equivalence relation which for one part means it must be symmetric. That wouldn't be the case for a StatementList and an array of statements.

thiemowmde:
I don't see what's wrong with my example. A StatementList and an array containing the same Statement objects *is* equivalent.

PR #616, opened 2016-02-09
Fix major bug in Property::clear

thiemowmde:


JeroenDeDauw:
Conflicts with #614

thiemowmde:
It's not like this is doing the opposite of #614. #614 can easily be rebased and still be merged later.

PR #617, opened 2016-02-09
Cleanup ReferenceList changes

Benestar:
Follow up of #609

### DIFF
--- a/src/ReferenceList.php
+++ b/src/ReferenceList.php
@@ -126,7 +126,7 @@ class ReferenceList implements Comparable, Countable, IteratorAggregate {
 	 */
 	public function indexOf( Reference $reference ) {
 		foreach ( $this->references as $index => $ref ) {
-			if ( $ref->equals( $reference ) ) {
+			if ( $ref === $reference ) {
 				return $index;
 			}
 		}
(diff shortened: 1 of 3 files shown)


src/ReferenceList.php
```diff
@@ -126,7 +126,7 @@ public function hasReference( Reference $reference ) {
 	 */
 	public function indexOf( Reference $reference ) {
 		foreach ( $this->references as $index => $ref ) {
-			if ( $ref->equals( $reference ) ) {
+			if ( $ref === $reference ) {
```
thiemowmde: Test?

tests/unit/ReferenceListTest.php
```diff
@@ -191,6 +191,17 @@ public function testAddReferenceAtIndexZero() {
 		$this->assertSameReferenceOrder( $expectedList, $references );
 	}
 
+	public function testAddReferenceAtNegativeIndex() {
+		$reference1 = new Reference( array( new PropertyNoValueSnak( 1 ) ) );
+		$reference2 = new Reference( array( new PropertyNoValueSnak( 2 ) ) );
```
thiemowmde: The first two are not needed for this test.

src/ReferenceList.php
Benestar: This is now the same behaviour as before. I can create a test for the old behaviour though.

thiemowmde: As I already tried to explain: If it was possible to overlook a breaking change like this, with obviously no test, what else is untested and broken now? 1000 lines got deleted, 200 added (see https://github.com/wmde/WikibaseDataModel/compare/4.4.0...master).

Before doing any breaking change to the ReleaseList class I would have rewritten the whole test, adding dozens of small test cases for all public features of the class (including everything SPL related). Then it would be way, way more obvious what feature breaks and what not.

Now we have no good idea and no easy way to find out and be sure.

PR #619, opened 2016-02-09
Make ReferenceList implement Serializable

Benestar:

PR #620, opened 2016-02-09
[DNM] Make ReferenceList no longer implement Serializable

Benestar:
This got added in #260 because SplObjectStorage didn't
work nicely for serializing objects. However, this is
now solved as we now longer extend that class.

### DIFF
--- a/src/ReferenceList.php
+++ b/src/ReferenceList.php
@@ -7,6 +7,7 @@ use Comparable;
 use Countable;
 use InvalidArgumentException;
 use IteratorAggregate;
+use Serializable;
 use Traversable;
 use Wikibase\DataModel\Internal\MapValueHasher;
 use Wikibase\DataModel\Snak\Snak;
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
Alternative to #619

JeroenDeDauw:
I think this is only a good idea if we do not rely on the serialization stuff. If we do, then keeping the Serializable implementation is nice, as it prevents a break in being able to deserialize old serializations (unless there was some change I don't have in mind now).

Benestar:
That's actually a very good point in favour of the alternative of this patch. If no other opinions arise, I'd go with the other one and close this.

thiemowmde:
What's the benefit of removing these methods?

This is an empty ReferenceList with the custom serializer:
`C:32:"Wikibase\DataModel\ReferenceList":6:{a:0:{}}`

Without:
`O:32:"Wikibase\DataModel\ReferenceList":1:{s:44:"Wikibase\DataModel\ReferenceListreferences";a:0:{}}`

I suggest to keep the methods, mostly to not break existing serializations.

PR #622, opened 2016-02-11
Rework test for ReferenceList::indexOf using identity

thiemowmde:
The braking change to the indexOf method was introduced in #609 and reverted in #617. The test was added in #617 but did not mentioned what it is testing: identity vs. equality.

### DIFF
--- a/tests/unit/ReferenceListTest.php
+++ b/tests/unit/ReferenceListTest.php
@@ -236,14 +236,15 @@ class ReferenceListTest extends \PHPUnit_Framework_TestCase {
 		}
 	}
 
-	public function testIndexOf_falseForMissingReferences() {
-		$referenceList = new ReferenceList();
-		$reference = new Reference( array( new PropertyNoValueSnak( 1 ) ) );
+	public function testIndexOf_checksForIdentity() {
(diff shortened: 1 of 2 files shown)


tests/unit/ReferenceListTest.php
```diff
@@ -236,14 +236,15 @@ public function testIndexOf( ReferenceList $array ) {
...
-		$referenceList->addNewReference( new PropertyNoValueSnak( 1 ) );
+	public function testIndexOf_checksForIdentity() {
+		$reference1 = new Reference( array( new PropertyNoValueSnak( 1 ) ) );
+		$reference2 = new Reference( array( new PropertyNoValueSnak( 1 ) ) );
+		$referenceList = new ReferenceList( array( $reference1 ) );
 
-		$this->assertFalse( $referenceList->indexOf( new Reference() ) );
-		$this->assertFalse( $referenceList->indexOf( $reference ) );
+		$this->assertNotSame( $reference1, $reference2, 'post condition' );
+		$this->assertTrue( $reference1->equals( $reference2 ), 'post condition' );
```
JeroenDeDauw: These two assertions do not test the method under test here

thiemowmde: So? That's why they are marked as post conditions.

PR #623, opened 2016-02-11
Add more regression tests for ReferenceList

thiemowmde:
This is split from #621:
* This adds a few more regression tests.
* This rearranges the release notes. The only substantial change is that I removed the "still implements" line.


Benestar:
Thanks!

PR #624, opened 2016-02-12
Add missing since tag to EntityDocument

thiemowmde:
The only substantial change in this patch is the added `@since 5.0` and the use of `self`.

I'm also removing newlines to make use of the common 100 character line length limit, and rearranging the methods:

1. Type.
2. Id getter and setter.
3. isEmpty and equals.


Benestar:
manually rebased

PR #625, opened 2016-02-12
Add EntityDocument::copy

thiemowmde:
We may not need this at all when we make use of the `clone` keyword and PHP's magic `__clone` methods instead, see http://php.net/manual/en/language.oop5.cloning.php. I will upload a separate patch for this alternative proposal.

### DIFF
--- a/src/Entity/EntityDocument.php
+++ b/src/Entity/EntityDocument.php
@@ -65,4 +65,15 @@ interface EntityDocument {
 	 */
 	public function equals( EntityDocument $entity );
 
+	/**
+	 * Returns a deep clone of the entity. The clone must be equal in all details, including the id.
+	 * Since EntityDocuments are not immutable (at least the id can be set) the method is not
+	 * allowed to return $this.
+	 *
+	 * @since 5.0
+	 *
+	 * @return self
+	 */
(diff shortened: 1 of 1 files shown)


thiemowmde:
@brightbyte, I just realized that this patch we talked about is already there. We could (this is just a suggestion) include this in the 5.0 release.

For the record: We started discussing if entities should have a `copy` method or implement `__clone` (see #626). There is a third way: We can have `EntityDocument::copy`, and the implementation can still use PHP's `clone` feature. In other words: `$entity->copy()` would just `return clone $this;`, and calling `$clone = clone $entity;` would be identical to `$clone = $entity->copy();`. One would see the method clearly documented in the interface, and can still use `clone`, if preferred.

TL;DR: This patch can be the base for this third approach. I would rebase #626 on top of this one then. But this needs more tests and should go into a 6.0, not in the planned 5.0.

Benestar:
After discussion with Daniel I agree with you that it makes most sense to have `copy` in EntityDocument as the public interface method. We can use clone or something different internally and change that without breaking the public interface. However, outside of this component one should never call `clone` because it is not guaranteed how exactly it is implemented.

Issue #611, opened 2016-02-07
Drop PHP 5.3 and 5.4 support


JeroenDeDauw:

JeroenDeDauw:
Any reason to not do this in 5.0 yet?

Benestar:
I don't know when https://gerrit.wikimedia.org/r/#/c/266931/ will be merged. That is the blocker to remove php 5.3 support in Wikibase.git afaik.

JeroenDeDauw:
Oh... looks like fail happened at https://github.com/Wikidata-lib/PropertySuggester/pull/155 then

brightbyte:
MediaWiki 1.27 requires compat with PHP 5.5 [1], and that's what CI at WMF will be running for a while. We are dropping 5.3 and 5.4. I think it will be a while until we can drop support for 5.5.

[1] RFC T118932 sais: "In the IRC meeting it was approved that our minimum version will be the common subset of features supported by PHP 5.5 and HHVM 3.1."

JeroenDeDauw:
Right. Typo in the description fixed.

Benestar:
Patch is merged and PHP 5.5 required. https://lists.wikimedia.org/pipermail/wikitech-l/2016-February/084751.html

JeroenDeDauw:
I had a look at the [publicly listed dependants on DataModel](https://packagist.org/packages/wikibase/data-model/dependents), seems like the stuffs from @addshore, @Tpt and me all requires 5.5 or above already (or rather "already"... :D). So I'm not aware of any users who will be negatively affected by doing this bump.

mariushoch:
+1 Wikibase is already PHP 5.5+ by now, so this wouldn't block us using DM 5.0 immediately.

thiemowmde:
What does "drop support" actually mean? In my opinion, the require tag in the composer.json file should only be touched when this code actually uses a PHP 5.5 feature. Why should the requirement be changed while the code is still compatible?

Or to ask the question an other way: What are you proposing to do to fix this issue?

brightbyte:
Can we do this now?

thiemowmde:
Done in #680.

PR #626, opened 2016-02-12
Implement __clone to fully support deep cloning

thiemowmde:
Alternative to #625. Personally I prefer this one. It's the native PHP way of copying/(deep)cloning objects.

I'm not totally happy with the test because it's so long.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -2,6 +2,7 @@
 
 ## Version 5.1 (dev)
 
+* `Item::copy` and `Property::copy` do not clone immutable objects any more, saving time and memory
 * Deprecated `FingerprintHolder` and `StatementListHolder`
 
 ## Version 5.0.2 (2016-02-23)
--- a/src/Entity/EntityDocument.php
+++ b/src/Entity/EntityDocument.php
@@ -70,6 +70,9 @@ interface EntityDocument extends Comparable {
 
 	/**
 	 * Returns a deep clone of the entity. The clone must be equal in all details, including the id.
+	 * No change done to the clone is allowed to interfere with the original object. Only properties
+	 * containing immutable objects are allowed to (and should) reference the original object.
+	 *
 	 * Since EntityDocuments are not immutable (at least the id can be set) the method is not
 	 * allowed to return $this.
 	 *
--- a/src/Entity/Item.php
+++ b/src/Entity/Item.php
@@ -327,7 +327,19 @@ class Item extends Entity implements StatementListHolder {
 	 * @return self
 	 */
 	public function copy() {
-		return unserialize( serialize( $this ) );
+		return clone $this;
+	}
+
+	/**
+	 * @see http://php.net/manual/en/language.oop5.cloning.php
+	 *
+	 * @since 5.1
+	 */
+	public function __clone() {
+		$this->fingerprint = clone $this->fingerprint;
+		// SiteLinkList is mutable, but SiteLink is not. No deeper cloning necesarry.
+		$this->siteLinks = clone $this->siteLinks;
+		$this->statements = clone $this->statements;
 	}
 
 }
--- a/src/Statement/StatementList.php
+++ b/src/Statement/StatementList.php
@@ -325,4 +325,15 @@ class StatementList implements IteratorAggregate, Comparable, Countable {
 		return $statementList;
 	}
 
+	/**
+	 * @see http://php.net/manual/en/language.oop5.cloning.php
+	 *
+	 * @since 5.1
+	 */
+	public function __clone() {
+		foreach ( $this->statements as &$statement ) {
+			$statement = clone $statement;
+		}
+	}
+
 }
--- a/src/Term/Fingerprint.php
+++ b/src/Term/Fingerprint.php
@@ -278,4 +278,17 @@ class Fingerprint implements Comparable, LabelsProvider, DescriptionsProvider, A
 		$this->aliasGroups = $groups;
 	}
 
+	/**
+	 * @see http://php.net/manual/en/language.oop5.cloning.php
+	 *
+	 * @since 5.1
+	 */
+	public function __clone() {
+		// TermList is mutable, but Term is not. No deeper cloning necesarry.
+		$this->labels = clone $this->labels;
+		$this->descriptions = clone $this->descriptions;
+		// AliasGroupList is mutable, but AliasGroup is not. No deeper cloning necesarry.
+		$this->aliasGroups = clone $this->aliasGroups;
+	}
+
 }
--- a/tests/unit/Entity/ItemTest.php
+++ b/tests/unit/Entity/ItemTest.php
@@ -4,9 +4,9 @@ namespace Wikibase\DataModel\Tests\Entity;
 
 use Wikibase\DataModel\Entity\Item;
 use Wikibase\DataModel\Entity\ItemId;
-use Wikibase\DataModel\Entity\PropertyId;
 use Wikibase\DataModel\SiteLink;
 use Wikibase\DataModel\Snak\PropertyNoValueSnak;
+use Wikibase\DataModel\Snak\PropertySomeValueSnak;
 use Wikibase\DataModel\Statement\Statement;
 use Wikibase\DataModel\Statement\StatementList;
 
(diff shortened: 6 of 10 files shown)


thiemowmde:
Pinging @brightbyte, because we talked about this and I'm quite happy with this solution.

Benestar:
From a first glance I like this approach. I'm not sure which implications the __clone has inside of PHP's dark magic but I think this is a sane solution using a language feature designed explicitely for this purpose.

For the test: You could move out a block of assertions into a private custom `assertFoo` method. Furthermore, you may consider splitting the modification part into its own test method.

JeroenDeDauw:
This is not equivalent to #625, as for this approach, there is no guarantee that when one gets an `EntityDocument`, using `clone` will get you a deep clone, while with the other PR, calling `copy` does provide this guarantee. (This is not an objection, just an observation.)

What motivated these changes? I'm guessing that for some reason using `unserialize( serialize( $foo ) )` is an approach some people want to avoid. Is that the case? I'm also wondering why this is just done for EntityDocument derivatives.

thiemowmde:
> there is no guarantee that when one gets an EntityDocument, using clone will get you a deep clone, while with the other PR, calling copy does provide this guarantee.

This is not true. Calling `copy` does not provide any better or weaker guarantee than using `clone`. The same guarantees can be given for both. What's missing in this pull request is the necessary documentation for the required cloning behavior of entity classes. You can even have the line `public function __clone();` in the interface, enforcing subclasses to implement it. But even that still does not guarantee all subclasses implement it correctly. There is no abstract way to test this.

> using unserialize( serialize( $foo ) ) is an approach some people want to avoid.

I find this very obvious, even if you do not consider all possible issues with that approach.
* It's scary and wastes resources. It does not distinguish between mutable and immutable value objects. It rebuilds everything because it can not know this is pointless and a waste of time and memory for immutable objects.
* It does not distinguish between pure value objects and other properties. The serialize approach will completely break the moment any of the DataModel classes contains a service. The whole service object will be serialized and deserialized then. And since we want to allow others to come up with their own entity types, we can not guarantee they are all pure value objects.

> why this is just done for EntityDocument derivatives

What do you mean by that? This pull request is complete: it removes all `copy` methods and implements all `__clone` methods necessary for the expected deep copy behavior.

Benestar:
> You can even have the line public function __clone(); in the interface, enforcing subclasses to implement it. But even that still does not guarantee all subclasses implement it correctly. There is no abstract way to test this.

We should definitely add a sentence or two as documentation to the `EntityDocument` interface to define the expected behaviour for all implementations of this. While this doesn't give us a guarantee on the technical level, it is still part of the contract and we can be sure that every implementation matching that contract by implementing the interface also has to follow that guideline.

tests/unit/Entity/ItemTest.php
```diff
@@ -400,4 +401,44 @@ public function testNotEquals( Item $firstItem, Item $secondItem ) {
...
+		$this->assertSame( $id, $clone->getId(), 'id is immutable and must not be cloned' );
+		$this->assertNotSame( $fingerprint, $clone->getFingerprint() );
+		$this->assertNotSame( $siteLinks, $clone->getSiteLinkList() );
+		$this->assertSame(
+			$siteLink,
+			$clone->getSiteLinkList()->toArray()['enwiki'],
+			'SiteLink is immutable and must not be cloned'
+		);
+		$this->assertNotSame( $statements, $clone->getStatements() );
+		$this->assertNotSame( $statement, $clone->getStatements()->toArray()[0] );
```
Benestar: Does not work in PHP 5.3 (https://github.com/wmde/WikibaseDataModel/issues/611#issuecomment-182474997)

tests/unit/Entity/ItemTest.php
```diff
@@ -400,4 +401,44 @@ public function testNotEquals( Item $firstItem, Item $secondItem ) {
...
+
+		// Modify the clone
+		$clone->getFingerprint()->setLabel( 'en', 'clone' );
+		$clonedStatement = $clone->getStatements()->toArray()[0];
```
Benestar: Does not work in PHP 5.3 (https://github.com/wmde/WikibaseDataModel/issues/611#issuecomment-182474997)

tests/unit/Entity/PropertyTest.php
```diff
@@ -229,4 +230,37 @@ public function testPropertyWithStatementsIsNotEmpty() {
...
+
+		$clone = clone $original;
+
+		// The clone must not reference the same mutable objects
+		$this->assertSame( 'property', $clone->getType() );
+		$this->assertSame( 'string', $clone->getDataTypeId() );
+		$this->assertSame( $id, $clone->getId(), 'id is immutable and must not be cloned' );
+		$this->assertNotSame( $fingerprint, $clone->getFingerprint() );
+		$this->assertNotSame( $statements, $clone->getStatements() );
+		$this->assertNotSame( $statement, $clone->getStatements()->toArray()[0] );
```
Benestar: Does not work in PHP 5.3

src/Entity/EntityDocument.php
```diff
@@ -65,4 +65,16 @@ public function isEmpty();
...
+	 * Classes implementing this interface must make sure they support deep cloning via PHP's clone
+	 * keyword. The clone must be equal and have the same id as the original object. However, no
+	 * change done to the clone is allowed to interfere with the original object. Only properties
+	 * containing immutable objects are allowed to (and should) reference the original object.
+	 *
+	 * @see http://php.net/manual/en/language.oop5.cloning.php
+	 *
+	 * @since 5.0
+	 */
+	public function __clone();
```
Benestar: Not needed anymore, since `copy` is now the public interface.

src/Entity/Item.php
```diff
@@ -319,4 +319,16 @@ public function equals( EntityDocument $entity ) {
...
 	}
 
+	/**
+	 * @see EntityDocument::__clone
+	 *
+	 * @since 5.0
+	 */
+	public function __clone() {
+		$this->fingerprint = clone $this->fingerprint;
+		// SiteLinkList is mutable, but SiteLink is not. No deeper cloning necesarry.
```
Benestar: I don't see how this comment helps in this place since you wouldn't do deeper cloning here anyways. Better move it somewhere into `SiteLinkList`.

src/Entity/Item.php
```diff
@@ -319,4 +319,16 @@ public function equals( EntityDocument $entity ) {
 		       && $this->statements->equals( $entity->statements );
 	}
 
+	/**
+	 * @see EntityDocument::__clone
+	 *
+	 * @since 5.0
```
Benestar: This shouldn't be called outside of this component so I'd remove the `@since` tag and perhaps mark this somehow as package-private.

src/Term/Fingerprint.php
```diff
@@ -278,4 +278,17 @@ public function setAliasGroups( AliasGroupList $groups ) {
...
+	/**
+	 * @see EntityDocument::__clone
+	 *
+	 * @since 5.0
+	 */
+	public function __clone() {
+		// TermList is mutable, but Term is not. No deeper cloning necesarry.
+		$this->labels = clone $this->labels;
+		$this->descriptions = clone $this->descriptions;
+		// AliasGroupList is mutable, but AliasGroup is not. No deeper cloning necesarry.
```
Benestar: As above, you wouldn't do deeper cloning at this place but in `TermList` resp. `AliasGroupList`.

tests/unit/Entity/ItemTest.php
```diff
@@ -329,10 +328,10 @@ public function equalsProvider() {
 		$secondItem = new Item();
 		$secondItem->getStatements()->addNewStatement( new PropertyNoValueSnak( 42 ) );
 
-		$secondItemWithId = $secondItem->copy();
+		$secondItemWithId = clone $secondItem;
 		$secondItemWithId->setId( 42 );
 
-		$differentId = $secondItemWithId->copy();
+		$differentId = clone $secondItemWithId;
```
Benestar: These changes need to be undone now we decided to keep `copy`.

Benestar:
I think we still want something like this to improve performance of `copy`. However, this should not be part of the public interface of the classes (ie. cloning should only happen within this component).

Tests are missing for the other classes that now support cloning.

Benestar:
Also, I have created [T126795](https://phabricator.wikimedia.org/T126795) which you may want to add to the sprint.

JeroenDeDauw:
Needs rebase

src/Entity/Item.php
thiemowmde: There is nothing in `SiteLinkList` where this comment could be placed. I can't comment on non-existing code.

src/Term/Fingerprint.php
thiemowmde: Yes. So? These comments explain why there is no `__clone` in these classes.

thiemowmde:
Rebased. Sure, we could add tests for the inner classes that now implement `__clone`. But I think this is only necessary when we want to make the possibility to use `clone` part of the public interface of these classes. I had the impression we do not want this.

The tests in the top level classes `Item` and `Property` are a bit more complex to cover everything we care about. Splitting these tests would not help much, in my opinion.

Benestar:
An alternative to this would be to add `copy` methods to the classes and create new instances by calling the constructor while calling `copy`. I don't know if `clone` is better performance-wise, but with adding `copy` to more classes, we would allow users of this component to also use that mechanism (which we don't want to if we implement `clone`).

thiemowmde:
Adding `copy` to all other mutable classes can be done in a later patch. This is independent of what I do in this patch. This patch is exclusively about not wasting time and memory. Currently even immutable objects are cloned (snaks, terms and so on), which is enormous on bigger entities. Calling constructors would, in most cases, not gain us much.

Example: Even if a TermList constructor does not clone all Term objects, it does a lot of checks that are just not necessary, because we already know everything in the TermList we want to clone is valid.

Benestar:
That's actually a good point. Needs rebase though.

Benestar:
Is this a breaking change? Otherwise, I'd suggest to do this in a 5.1 release. (Or maybe 5.0.1 since we only change an implementation detail here?) That allows us to use the benefits of this right away since 6.0.0 perhaps still needs some work on it.

thiemowmde:
At this point, I would argue that this is purely an implementation detail. I changed the patch to talk about a 5.0.1 now.

Later, we may decide that we want to add the possibility to use `clone` to the public documentation. But even this will not need a further version bump. This would actually be a bugfix, because nothing prevents you from using `clone` now, but you will not get a proper deep clone.

Benestar:
+1 from me but someone else must approve

adrianheine:
Needs rebase again :/

thiemowmde:
Rebased.

RELEASE-NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Wikibase DataModel release notes
 
+## Version 5.0.2 (alpha)
```
Benestar: This introduces new functionality and should go into a 5.1

thiemowmde: This is not new, as already argued in https://github.com/wmde/WikibaseDataModel/pull/626#issuecomment-185171983. The interfaces are untouched.
* `copy` behaves as before. The fact that immutable objects are not cloned any more can only have an impact on a user if he was using it wrong before.
* The possibility to use `clone` with Item and Property objects was there before, but not documented as part of the public interface of these classes. This did not changed.

Benestar:
Needs rebase and I wonder if @brightbyte wants to review this again? Would be nice to have this in the next 5.1 release.

thiemowmde:
Rebased.

RELEASE-NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Wikibase DataModel release notes
 
+## Version 5.0.3 (alpha)
```
Benestar: I still think this must be a 5.1 release.

> Patch version Z (x.y.Z | x > 0) MUST be incremented if only backwards compatible bug fixes are introduced. A bug fix is defined as an internal change that fixes incorrect behavior.

This is no bug but a performance improvement.

> Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API. It MUST be incremented if any public API functionality is marked as deprecated. It MAY be incremented if substantial new functionality or improvements are introduced within the private code. It MAY include patch level changes. Patch version MUST be reset to 0 when minor version is incremented.

I think "substantial improvements in private code" fits pretty well here.

JeroenDeDauw: Indeed, welcome to the SemVer police @Benestar! :)

thiemowmde: If this is substantial or not depends a lot. I don't even know if we will see this having any impact. I hope, but I do not know. Do we even have a graf for memory usage?

0.Y steps are for added functionality. This is no added functionality. This is a bugfix, fixing a performance issue.

Benestar: Performance issues aren't bugs, bugs are wrong functionality.

> A bug fix is defined as an internal change that fixes incorrect behavior.

Behaviour of `copy` is totally correct, it's not buggy at all.

brightbyte:
@benestar I'll review this if there is time. I'm trying not to get side-tracked by things that are not blocking the implementation of new entity types.

src/Statement/StatementList.php
```diff
@@ -325,4 +325,15 @@ public function filter( StatementFilter $filter ) {
 		return $statementList;
 	}
 
+	/**
+	 * @see http://php.net/manual/en/language.oop5.cloning.php
+	 *
+	 * @since 5.1
+	 */
+	public function __clone() {
+		foreach ( $this->statements as &$statement ) {
```
brightbyte: It's really a shame we have to do that even in a context where no modification will ever happen.
Can we have an Immutable marker interface, so cloning can be skipped for anything that implements it?
That way, we could support mutable and immutable implementations of StatementList, TermList, etc.

brightbyte:
The approach is good, and the code looks fine too. I'm inclined to merge this. However, if something was missed here, the result would be subtle data corruption. Hard to detect, and hard to fix. So let's please get more eyes on this.

src/Statement/StatementList.php
thiemowmde: The issue here is not that StatementList is mutable but that Statement is mutable. What you are suggesting requires an immutable Statement. Personally I do not think it's worth our time to go this route. As shown on Phabricator this patch saves 90% memory. Having an immutable Statement will probably save some additional 100 KB. Is that worth it?

thiemowmde:
The way to be sure about this is pretty straightforward: Think of anything that can be changed in an Item or Property. You can add and remove labels, descriptions and aliases. But you can not edit them. The only thing you can do is replacing a label, description or set of aliases. You can add and remove sitelinks and statements. You can not edit a sitelink but must replace it. You can edit a lot of details in a statement. I think the later is covered sufficiently by editing some details of a statement, e.g. the rank.

thiemowmde:
I added additional assertions for descriptions, aliases, statement guids, main snaks and sitelinks.

brightbyte:
Seems solid enough. Would have been nice for someone else to double-check, but let's keep moving.

PR #630, opened 2016-02-16
Move EntityTest into subclasses

Benestar:
By moving all the test methods from EntityTest into subclasses, we make
sure that still all tests are executed when removing the inheritance
between Item/Property and Entity. We can then start to remove tests that
test methods no longer existing while still having the old tests.

**Note:** I changed this patch fundamentally.


JeroenDeDauw:
Had very quick look. +0.5

tests/unit/Entity/EntityTest.php
```diff
@@ -14,21 +14,20 @@
...
+class EntityTest extends \PHPUnit_Framework_TestCase {
 
 	/**
-	 * @since 0.1
-	 *
 	 * @return Entity
 	 */
-	protected abstract function getNewEmpty();
+	private function getNewEmpty() {
+		return new Item();
```
thiemowmde: Huh? This appears to be wrong.

tests/unit/Entity/EntityTest.php
```diff
@@ -14,21 +14,20 @@
...
  *
- * @group Wikibase
- * @group WikibaseDataModel
+ * @covers Wikibase\DataModel\Entity\Entity
  *
  * @licence GNU GPL v2+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
  * @author Daniel Kinzler
  */
-abstract class EntityTest extends \PHPUnit_Framework_TestCase {
```
thiemowmde: Why is this not abstract any more?

tests/unit/Entity/EntityTest.php
```diff
@@ -14,21 +14,20 @@
  * @deprecated
  * This test class is to be phased out, and should not be used from outside of the component!
  *
- * @group Wikibase
- * @group WikibaseDataModel
```
thiemowmde: Why did you removed these?

tests/unit/Entity/ItemTest.php
```diff
@@ -368,4 +391,18 @@ public function testNotEquals( Item $firstItem, Item $secondItem ) {
 		$this->assertFalse( $secondItem->equals( $firstItem ) );
 	}
 
+	public function testCopy() {
```
thiemowmde: Conflicts with #626.

tests/unit/Entity/EntityTest.php
Benestar: We can also just remove the test in this patch already. It doesn'r really matter anyways. However, I didn't want to leave `Entity` untested and thus made the test non-abstract so that it tests a `Entity` implementation instead.

tests/unit/Entity/EntityTest.php
Benestar: see above

tests/unit/Entity/EntityTest.php
thiemowmde: Oh, I see. Makes sense. I confused this with the abstract `Entity` class.

tests/unit/Entity/EntityTest.php
thiemowmde: As said in an other pull request, this is unrelated and should be done in a separate pull request. Personally I prefer to add both `@group Wikibase` and `@group WikibaseDataModel` to all tests.

Benestar:
I re-added the `@group` annotations so maybe you can have another look into this.

tests/unit/Entity/PropertyTest.php
```diff
@@ -229,4 +262,18 @@ public function testPropertyWithStatementsIsNotEmpty() {
 		$this->assertFalse( $property->isEmpty() );
 	}
 
+	public function testCopy() {
```
thiemowmde: This conflicts heavily with what I did in #626.

tests/unit/Entity/PropertyTest.php
```diff
@@ -229,4 +262,18 @@ public function testPropertyWithStatementsIsNotEmpty() {
...
+		$property = Property::newFromType( 'string' );
+		$property->setLabel( 'en', 'Foo' );
+
+		$copy = $property->copy();
+
+		$this->assertNotSame( $property, $copy );
+		$this->assertTrue( $property->equals( $copy ) );
```
thiemowmde: The original implementation compared the id, which is still necessary.

thiemowmde:
This copies only a few test cases, but other are now missing and not run on Property any more.

PR #631, opened 2016-02-16
Make Item and Property no longer extend Entity

Benestar:
This patch removes the inheritance from Item and Property to Entity. It
also removes Entity and EntityTest as Entity no longer has an
implementation in this component and cannot be used or tested now.

Depends on #630 

**Note that this shouldn't be merged before we are sure the next release will be 6.0.0**

Bug: [T78287](https://phabricator.wikimedia.org/T78287)


JeroenDeDauw:
> Note that this shouldn't be merged before we are sure the next release will be 6.0.0

Can always bump master to 6.x alpha and make new releases of a 5.x branch. Really depends on what we expect to be easiest.

JeroenDeDauw:
Had a very quick look. +0.5 (though agree this should not be merged before thought is given on how and when to do the next breaking rel)

RELEASE-NOTES.md
```diff
@@ -1,5 +1,13 @@
 # Wikibase DataModel release notes
 
+## Version 6.0.0 (dev)
+
+* Removed `Entity` class (deprecated since 1.0)
+* `Item` and `Property` no longer extend `Entity`
+    * Removed `getLabel`, `getLabels`, `getDescription`, `getDescriptions`, `getAliases`,
+      `getAllAliases`, `setLabels`, `setDescriptions`, `addAliases`, `setAllAliases`,
+      `removeLabel`, `removeDescription` and `removeAliases` methods
+
```
brightbyte: Verified that these cover all the methods that used to be in Entity, and no longer are in Item and Property.

PR #632, opened 2016-02-16
Add interfaces to Item and Property

Benestar:
With this patch applied, Item and Property implement the term related
interfaces LabelsProvider, DescriptionsProvider and AliasesProvider.

Note that the signature of getLabels and getDescriptions changed and those
methods now return TermList objects instead of plain arrays.

Depends on #631 

**Note that this shouldn't be merged before we are sure the next release will be 6.0.0**

Bug: [T126152](https://phabricator.wikimedia.org/T126152)

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -4,9 +4,13 @@
 
 * Removed `Entity` class (deprecated since 1.0)
 * `Item` and `Property` no longer extend `Entity`
-    * Removed `getLabel`, `getLabels`, `getDescription`, `getDescriptions`, `getAliases`,
-      `getAllAliases`, `setLabels`, `setDescriptions`, `addAliases`, `setAllAliases`,
+    * Removed `getLabel`, `getDescription`, `getAliases`, `getAllAliases`,
+      `setLabels`, `setDescriptions`, `addAliases`, `setAllAliases`,
       `removeLabel`, `removeDescription` and `removeAliases` methods
+* `Item` and `Property` now implement `LabelsProvider`, `DescriptionsProvider` and `AliasesProvider`
+* `Item::getLabels` and `Property::getLabels` now return a `TermList`
+* `Item::getDescriptions` and `Property::getDescriptions` now return a `TermList`
+* Added `Item::getAliasGroups` and `Property::getAliasGroups`
 * `TermList` and `AliasGroupList` no longer throw an `InvalidArgumentException` for invalid language codes.
     * `getByLangauge` now throws an `OutOfBoundsException`.
     * `removeByLanguage` does nothing for invalid values.
--- a/src/Entity/Item.php
+++ b/src/Entity/Item.php
@@ -8,8 +8,13 @@ use Wikibase\DataModel\SiteLink;
 use Wikibase\DataModel\SiteLinkList;
 use Wikibase\DataModel\Statement\StatementList;
 use Wikibase\DataModel\Statement\StatementListHolder;
+use Wikibase\DataModel\Term\AliasesProvider;
+use Wikibase\DataModel\Term\AliasGroupList;
+use Wikibase\DataModel\Term\DescriptionsProvider;
 use Wikibase\DataModel\Term\Fingerprint;
 use Wikibase\DataModel\Term\FingerprintHolder;
+use Wikibase\DataModel\Term\LabelsProvider;
+use Wikibase\DataModel\Term\TermList;
 
 /**
  * Represents a single Wikibase item.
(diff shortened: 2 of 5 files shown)


JeroenDeDauw:
Had a quick look. +1

tests/unit/Entity/PropertyTest.php
```diff
@@ -229,4 +320,18 @@ public function testPropertyWithStatementsIsNotEmpty() {
 		$this->assertFalse( $property->isEmpty() );
 	}
 
+	public function testCopy() {
```
thiemowmde: Unrelated. Please move to a separate patch.

tests/unit/Entity/PropertyTest.php
```diff
@@ -7,32 +7,20 @@
...
+use Wikibase\DataModel\Term\Term;
+use Wikibase\DataModel\Term\TermList;
 
 /**
  * @covers Wikibase\DataModel\Entity\Property
- * @covers Wikibase\DataModel\Entity\Entity
- *
- * @group Wikibase
- * @group WikibaseProperty
- * @group WikibaseDataModel
```
thiemowmde: What's wrong with these tags? At least the `@group WikibaseDataModel` should stay.

tests/unit/Entity/ItemTest.php
```diff
@@ -368,4 +460,18 @@ public function testNotEquals( Item $firstItem, Item $secondItem ) {
 		$this->assertFalse( $secondItem->equals( $firstItem ) );
 	}
 
+	public function testCopy() {
```
thiemowmde: Unrelated. Please move to a separate patch.

tests/unit/Entity/ItemTest.php
```diff
@@ -4,43 +4,26 @@
...
- * @covers Wikibase\DataModel\Entity\Entity
- *
- * Some tests for this class are located in ItemMultilangTextsTest,
- * ItemNewEmptyTest and ItemNewFromArrayTest.
- *
- * @since 0.1
- *
- * @group Wikibase
- * @group WikibaseItem
- * @group WikibaseDataModel
```
thiemowmde: As below.

src/Entity/Property.php
```diff
@@ -251,4 +292,15 @@ public function setStatements( StatementList $statements ) {
...
 	}
 
+	/**
+	 * @see EntityDocument::copy
+	 *
+	 * @since 0.1
+	 *
+	 * @return self
+	 */
+	public function copy() {
```
thiemowmde: Unrelated. Also conflicts with #626. Please move to a separate patch.

src/Entity/Item.php
```diff
@@ -319,4 +360,15 @@ public function equals( $target ) {
...
 	}
 
+	/**
+	 * @see EntityDocument::copy
+	 *
+	 * @since 0.1
+	 *
+	 * @return self
+	 */
+	public function copy() {
```
thiemowmde: Unrelated. Also conflicts with #626. Please move to a separate patch.

RELEASE-NOTES.md
```diff
@@ -1,5 +1,17 @@
 # Wikibase DataModel release notes
 
+## Version 6.0.0 (dev)
+
+* Removed `Entity` class (deprecated since 1.0)
```
thiemowmde: I don't understand why we need to kill the class and refactor everything in a single step, in a patch that exclusively talks about new interfaces. We can easily introduce the interfaces and only remove the conflicting methods from Entity.

Benestar: You havev to read the patches in the correct order... This pull request includes three commits and it states in the description on which pull request it depends. Unfortunately, github doesn't handle patch chains as nice as gerrit.

tests/unit/Entity/PropertyTest.php
Benestar: We don't use `@group` annotations in most of our tests. Either use them everywhere or remove them, everything else is just confusing.

thiemowmde: This is unrelated to all tree commits in this chain. Please revert this and upload a separate patch. Personally, I prefer to have the said `@group WikibaseDataModel` in all tests.

thiemowmde:
Makes soooooooooo much sense now. @Benestar, thanks a lot for making these commits so much fun to review!

Benestar:
> Does this mean Fingerprint can also implement LabelsProvider, DescriptionsProvider and AliasesProvider now?

It already does :)

PR #633, opened 2016-02-17
Copy copy methods to Item and Property

thiemowmde:
This acts as a baseline for a whole bunch of patches that all touch the same code, including #626, #630, #631 and #632. All these patches should become a bit easier to review with this baseline.

The `@since` tags are intentional, because both classes are guaranteed to have a `copy` method since 0.1

### DIFF
--- a/src/Entity/Property.php
+++ b/src/Entity/Property.php
@@ -251,4 +251,15 @@ class Property extends Entity implements StatementListHolder {
 		$this->statements = $statements;
 	}
 
+	/**
+	 * @see EntityDocument::copy
+	 *
+	 * @since 0.1
+	 *
+	 * @return self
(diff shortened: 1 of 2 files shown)


src/Entity/Property.php
```diff
@@ -251,4 +251,15 @@ public function setStatements( StatementList $statements ) {
 		$this->statements = $statements;
 	}
 
+	/**
+	 * @see EntityDocument::copy
+	 *
+	 * @since 0.1
+	 *
+	 * @return Property
```
Benestar: self

thiemowmde: Uh, right. I had the wrong idea that `self` refers to `Entity`. o_O

PR #634, opened 2016-02-18
Fix ReferenceList not removing all reference with same hash

Benestar:
I'm actually very confused by this because as far as I can see the old implementation of this only removed the first reference with that hash in the list. [[1]](https://github.com/wmde/WikibaseDataModel/blob/4.4.0/src/ReferenceList.php#L192) However, without this patch, tests in Wikibase are broken because of this missing feature. [[2]](https://integration.wikimedia.org/ci/job/mwext-Wikibase-repo-tests-sqlite-hhvm/8473/consoleFull)

Anyways, it is very bad to change the array keys while iterating over it in a for loop so we need to fix that to use it in Wikibase and release a 5.0.1.

### DIFF
--- a/src/ReferenceList.php
+++ b/src/ReferenceList.php
@@ -117,7 +117,7 @@ class ReferenceList implements Comparable, Countable, IteratorAggregate, Seriali
 	}
 
 	/**
-	 * Returns the index of a reference or false if the reference could not be found.
+	 * Returns the index of the Reference object or false if the Reference could not be found.
 	 *
 	 * @since 0.5
 	 *
@@ -160,29 +160,31 @@ class ReferenceList implements Comparable, Countable, IteratorAggregate, Seriali
 	}
 
 	/**
-	 * Removes the reference with the provided hash if it exists in the list.
+	 * Looks for the first Reference object in this list with the provided hash.
+	 * Removes all occurences of that object.
 	 *
 	 * @since 0.3
 	 *
 	 * @param string $referenceHash	`
 	 */
 	public function removeReferenceHash( $referenceHash ) {
-		foreach ( $this->references as $index => $reference ) {
-			if ( $reference->getHash() === $referenceHash ) {
-				$this->removeReferenceAtIndex( $index );
+		$reference = $this->getReference( $referenceHash );
+
+		if ( $reference === null ) {
+			return;
+		}
+
+		foreach ( $this->references as $index => $ref ) {
+			if ( $ref === $reference ) {
+				unset( $this->references[$index] );
 			}
 		}
+
+		$this->references = array_values( $this->references );
 	}
 
 	/**
-	 * @param int $index
-	 */
-	private function removeReferenceAtIndex( $index ) {
-		array_splice( $this->references, $index, 1 );
-	}
-
-	/**
-	 * Returns the reference with the provided hash, or
+	 * Returns the first Reference object with the provided hash, or
 	 * null if there is no such reference in the list.
 	 *
 	 * @since 0.3
--- a/tests/unit/ReferenceListTest.php
+++ b/tests/unit/ReferenceListTest.php
@@ -157,6 +157,26 @@ class ReferenceListTest extends PHPUnit_Framework_TestCase {
 		$this->assertTrue( true );
 	}
 
+	public function testRemoveReferenceRemovesIdenticalObjects() {
+		$reference = new Reference( array( new PropertyNoValueSnak( 1 ) ) );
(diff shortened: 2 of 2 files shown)


Benestar:
I think I now understand what's going on here:

`ChangeOpReferenceRemoveTest` in Wikibase checks that two references with the same hash both get removed. Usually they would not get removed with the old implementation of `ReferenceList`. However, in that test case, they got removed because the both are the very same object and thus both got removed by `detach`.

With the new implementation (which is broken anyways) that isn't the case anymore, because hashes are compared and not the objects themselves. This results in a lot of weirdness (wrong tests in Wikibase still succeeding).

If we want to fix this without breaking changes, we have to restore the behaviour of the old version of `ReferenceList`. That means, fetching the (first) object before removing, then removing all occurences of that object. I strongly suggest however to fix the weird behaviour in the next major release of this component.

Benestar:
I'm going to change this commit to restore the original behaviour.

Benestar:
I added a regression test and improved the documentation to say what actually happens in those methods.

src/ReferenceList.php
```diff
@@ -160,29 +160,27 @@ public function hasReferenceHash( $referenceHash ) {
...
 	 *
 	 * @param string $referenceHash	`
 	 */
 	public function removeReferenceHash( $referenceHash ) {
-		foreach ( $this->references as $index => $reference ) {
-			if ( $reference->getHash() === $referenceHash ) {
-				$this->removeReferenceAtIndex( $index );
+		$reference = $this->getReference( $referenceHash );
+
+		foreach ( $this->references as $index => $ref ) {
```
thiemowmde: This should not loop when the reference was not found.

src/ReferenceList.php
```diff
@@ -160,29 +160,27 @@ public function hasReferenceHash( $referenceHash ) {
...
 			}
 		}
-	}
 
-	/**
-	 * @param int $index
-	 */
-	private function removeReferenceAtIndex( $index ) {
-		array_splice( $this->references, $index, 1 );
+		$this->references = array_values( $this->references );
```
thiemowmde: There should be a test case for the reindexing feature, see #635.

src/ReferenceList.php
Benestar: True

src/ReferenceList.php
thiemowmde: I'm suggesting these tests, all tested with version 4.4:
```php
public function testRemoveReferenceHashRemovesIdenticalObjects() {
	$reference = new Reference( array( new PropertyNoValueSnak( 1 ) ) );
	$references = new ReferenceList( array( $reference, $reference ) );

	$references->removeReferenceHash( $reference->getHash() );

	$this->assertTrue( $references->isEmpty() );
}

public function testRemoveReferenceHashDoesNotRemoveCopies() {
	$reference = new Reference( array( new PropertyNoValueSnak( 1 ) ) );
	$references = new ReferenceList( array( $reference, clone $reference ) );

	$references->removeReferenceHash( $reference->getHash() );

	$this->assertFalse( $references->isEmpty() );
	$this->assertTrue( $references->hasReference( $reference ) );
	$this->assertNotSame( $reference, $references->getReference( $reference->getHash() ) );
}

public function testRemoveReferenceHashUpdatesIndexes() {
	$reference1 = new Reference( array( new PropertyNoValueSnak( 1 ) ) );
	$reference2 = new Reference( array( new PropertyNoValueSnak( 2 ) ) );
	$references = new ReferenceList( array( $reference1, $reference2 ) );

	$references->removeReferenceHash( $reference1->getHash() );

	$this->assertSame( 0, $references->indexOf( $reference2 ) );
}
```

tests/unit/ReferenceListTest.php
```diff
@@ -330,6 +330,21 @@ public function testRemoveReferenceHash( ReferenceList $references ) {
 		$this->assertEquals( 0, count( $references ) );
 	}
 
+	public function testRemoveReferenceHash_occurrencesOfSameObjectGetRemoved() {
```
thiemowmde: Should be moved down, below the generic `removeReferenceHash` test.

Benestar: Done

thiemowmde:
I added a bunch of additional test cases, also for `removeReference`, which does have the same issue. +2 from my side. I suggest to merge this into master and release a 5.0.1. There are more commits in the current master, but they are all fine for a 5.0.1 in my opinion.

PR #635, opened 2016-02-18
[WIP] Fix unexpected behavior of ReferenceList::removeReferenceHash

thiemowmde:
@Benestar found these issues in #634. Actually fixing these is a breaking change, because this behavior was identical in 4.x and 5.x.


Benestar:
Please rebase this on top of my changes so that travis doesn't fail. Also, the title suggests that this changes some behaviour while only tests got added.

thiemowmde:
Similar to #634 this is a breaking change that must be in a 6.x.

thiemowmde:
I made this part of #638 and set that on hold.

PR #637, opened 2016-02-18
Remove dead code from ReferenceListTest

thiemowmde:
Can you imagine what the `$this->assertTrue( true )` and the lines above was meant to do? I tracked it down to commit c721869f41fa0637a7171a61526b34acfaee4101 (on Gerrit). To me it appears to be a mistake.

### DIFF
--- a/tests/unit/ReferenceListTest.php
+++ b/tests/unit/ReferenceListTest.php
@@ -25,27 +25,13 @@ use Wikibase\DataModel\Snak\SnakList;
 class ReferenceListTest extends PHPUnit_Framework_TestCase {
 
 	public function instanceProvider() {
-		$instances = array();
-
-		foreach ( $this->getConstructorArg() as $arg ) {
-			$instances[] = array( new ReferenceList( $arg ) );
-		}
-
-		return $instances;
-	}
-
-	public function getElementInstances() {
 		return array(
-			new Reference(),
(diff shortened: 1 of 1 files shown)


Benestar:
Maybe testing that it doesn't throw an exception? Strange indeed...

As you suggested once, it would be best to rewrite the whole test while re-working the methods on `ReferenceList` to be more intuitive and less magic.

thiemowmde:
Rewrite step by step, along with the code we touch.

tests/unit/ReferenceListTest.php
```diff
@@ -147,14 +133,6 @@ public function testRemoveReference( ReferenceList $array ) {
...
 			$this->assertEquals( --$elementCount, count( $array ) );
 		}
-
-		$elements = $this->getElementInstances();
-		$element = array_shift( $elements );
-
-		$array->removeReference( $element );
-		$array->removeReference( $element );
-
-		$this->assertTrue( true );
```
JeroenDeDauw: Now the test will not make any assertions for empty lists. This test design is not nice, though now you will get warnings in strict mode if I'm not mistaken

thiemowmde: I tried again to be sure. No bad things happen in strict mode. CI does not complain, too.

JeroenDeDauw: Maybe PHPUnit changed this... I remember running into this several times near the start of the project

PR #638, opened 2016-02-18
[WIP] ReferenceList should never contain any duplicates

thiemowmde:
This patch is not about ignoring identical objects (that was #639), but about ignoring references with the same hash. The tests describe the proposed behavior.

We discussed this with @lydiapintscher and @Benestar and came to the conclusion that the current behavior is most probably what we want. Users may run into this when they try to save an identical reference as an intermediate step. So **do not merge** this. However, I would like to have this open for a while to gather opinions.


thiemowmde:
#639 is about an actual regression. This patch here is not about ignoring identical objects, but about ignoring references with the same hash. This is a breaking change and must be in a 6.x release.

PR #639, opened 2016-02-18
Fix regression when adding the same reference object twice

thiemowmde:
~~This is really ugly performance-wise, and we should think about this twice before merging it.~~ Reworked to use `spl_object_hash()` to replicate the behavior of the previous, SPL object store based implementation.

Marked as a bug because it is a regression between 4.4.0 (the tests in this patch succeed with 4.4.0) and 5.0.0.

### DIFF
--- a/src/ReferenceList.php
+++ b/src/ReferenceList.php
@@ -28,7 +28,7 @@ use Wikibase\DataModel\Snak\Snak;
 class ReferenceList implements Comparable, Countable, IteratorAggregate, Serializable {
 
 	/**
-	 * @var Reference[]
+	 * @var Reference[] Ordered list or references, indexed by SPL object hash.
 	 */
 	private $references = array();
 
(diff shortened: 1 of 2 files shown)


thiemowmde:
Rebased after #640 was merged.

thiemowmde:
@Benestar, I strongly suggest to release this fix as 5.0.2 before using it in Wikibase.git. We have quite some usages of `addReference` and the constructor. Following all code paths doesn't seem to be possible. I don't want duplicate references to show up in the serialization.

Benestar:
I agree that we should include this in a PATCH release although I doubt that we are working a lot with identical objects in production. Why does this pull request have the _breaking_ label attached?

thiemowmde:
This describes breaking change made between 4.4.0 and 5.0.0. Reverting this is technically a breaking change too. ;-P

JeroenDeDauw:
Please only using the breaking label for things requiring a major release. Else it becomes pretty pointless

thiemowmde:
Yea, it was a mistake and I removed it. I just wanted to explain why I added it originally.

PR #640, opened 2016-02-19
Add more regression tests for 4.x vs. 5.x compatibility

thiemowmde:
This is split from #639, which got pretty big when I reworked it this morning. All tests in this patch here succeed with version 4.4.0 and 5.0.1 (would be good if you could try this again). I would like to rebase #639 when this is merged to make it much smaller and easier to review.


thiemowmde:
Pinging @Benestar.

Benestar:
Verified that the tests also pass on 4.4.0

PR #641, opened 2016-02-19
Test 4.x behavior of ReferenceList::addReference at index

thiemowmde:
I wrote this test in 4.4.0. It describes the behavior of version 4.4.0 and succeeds there. It failed in 5.0.1. It succeeds in 5.0.2.

The test does test multiple things:
* It checks if identical objects are ignored. This was broken in 5.0.1 and got fixed in 5.0.2 by #639.
* It checks if it's possible to move an existing reference to a lower index.
* It tries to move a reference to a higher index, but expects this to fail, because 4.x was not able to do this.

### DIFF
--- a/tests/unit/ReferenceListTest.php
+++ b/tests/unit/ReferenceListTest.php
@@ -224,6 +224,24 @@ class ReferenceListTest extends PHPUnit_Framework_TestCase {
 		$this->assertCount( 1, $list );
 	}
 
+	public function testAddReferenceAtIndexMovesIdenticalObjects() {
+		$list = new ReferenceList();
+		$list->addNewReference( new PropertyNoValueSnak( 1 ) );
+		$reference = new Reference( array( new PropertyNoValueSnak( 2 ) ) );
+		$list->addReference( $reference );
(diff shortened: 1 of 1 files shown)


thiemowmde:
There are only two usages of this feature in `ChangeOpReference`.
* One is called after calling `removeReferenceHash`. This means none of the issues described above can happen.
* One is called when adding a new reference and an index was provided. This may run into the issues described above when the newly added reference happens to be identical to an existing reference. I can not reproduce this via the UI, because the JavaScript DataModel ignores duplicates. But this may happen via the API (still needs to be tested).

This means, #639 needs more investigation and should probably be released as a 5.0.2. This patch here is about the index parameter. This feature was broken before and is not critical in the use cases we currently have in Wikibase.git. Therefore I suggest to not restore the inconsistent 4.x behaviour but consider this a bug and fix it properly.

thiemowmde:
Rebased. This succeeds since 5.0.2, because #639 restored the original 4.x behavior. I suggest to merge this patch as it is and discuss a possible fix in a later one.

PR #644, opened 2016-02-24
Remove type checks from AliasGroup/TermList getters

thiemowmde:
This is about performance.

Why do I think these checks are not necessary? Because nothing unexpected happens when an array key is not found.
* When you provide an integer key or something else that can't be found in the array, all three methods have a well defined, expected behavior. These methods are not supposed to check if an array key is allowed to be in the array, this must be checked in the constructor/setter.
* On the other hand, when you provide some object that's then casted to string, and this string is then found in the array, why shouldn't that work?

### DIFF
--- a/src/Term/AliasGroupList.php
+++ b/src/Term/AliasGroupList.php
@@ -73,12 +73,9 @@ class AliasGroupList implements Countable, IteratorAggregate {
 	 * @param string $languageCode
 	 *
 	 * @return AliasGroup
-	 * @throws InvalidArgumentException
 	 * @throws OutOfBoundsException
 	 */
 	public function getByLanguage( $languageCode ) {
-		$this->assertIsLanguageCode( $languageCode );
-
 		if ( !array_key_exists( $languageCode, $this->groups ) ) {
 			throw new OutOfBoundsException( 'AliasGroup with languageCode "' . $languageCode . '" not found' );
 		}
@@ -99,17 +96,9 @@ class AliasGroupList implements Countable, IteratorAggregate {
 
 	/**
 	 * @param string $languageCode
-	 * @throws InvalidArgumentException
 	 */
 	public function removeByLanguage( $languageCode ) {
-		$this->assertIsLanguageCode( $languageCode );
 		unset( $this->groups[$languageCode] );
-	}
-
-	private function assertIsLanguageCode( $languageCode ) {
-		if ( !is_string( $languageCode ) || $languageCode === '' ) {
-			throw new InvalidArgumentException( '$languageCode must be a non-empty string' );
-		}
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


JeroenDeDauw:
+-0 not convinced this change makes any real performance difference

mariushoch:
+1 Function calls in PHP are quite expensive (especially with Xdebug enabled), but I don't think this is being called often enough to even remotely make a difference. Still I'm ok with that change, the check wasn't needed.

src/Term/AliasGroupList.php
```diff
@@ -99,19 +96,11 @@ public function getWithLanguages( array $languageCodes ) {
 
 	/**
 	 * @param string $languageCode
-	 * @throws InvalidArgumentException
 	 */
 	public function removeByLanguage( $languageCode ) {
-		$this->assertIsLanguageCode( $languageCode );
 		unset( $this->groups[$languageCode] );
```
Benestar: This should maybe throw if the language code wasn't set beforehand?

Benestar:
+1 but this seems to be a breaking change

thiemowmde:
What breaks?

src/Term/AliasGroupList.php
thiemowmde: That would indeed be a breaking change and is not what this patch is about.

Benestar:
Exactly the stuff where you had to change the tests. I don't think this will break anything in our Wikibase code but the behaviour of some public interface methods changed in an incompatible way. Lets just put this into 6.0 and take safe route ;-)

thiemowmde:
This is impossible to break anything. Lets pick an example: This patch changes `hasAliasGroup` to return false instead of throwing an exception. So callers must expect it to return false, but this is not new. The only thing that could happen is that you end with dead code doing `catch ( InvalidArgumentException )`. This `catch` was most probably dead code before (because invalid language codes could not make it that far anyway), or misplaced (since this kind of validity check is somewhat unexpected in getters), or even missing.

tests/unit/Term/AliasGroupListTest.php
```diff
@@ -162,11 +162,11 @@ public function testGivenSetLanguage_removeByLanguageRemovesIt() {
...
 	/**
 	 * @dataProvider invalidLanguageCodeProvider
-	 * @expectedException InvalidArgumentException
 	 */
-	public function testGivenInvalidLanguageCode_removeByLanguageThrowsException( $languageCode ) {
-		$list = new AliasGroupList();
+	public function testGivenInvalidLanguageCode_removeByLanguageIsNoOp( $languageCode ) {
+		$list = new AliasGroupList( array( new AliasGroup( 'en', array( 'foo' ) ) ) );
 		$list->removeByLanguage( $languageCode );
+		$this->assertFalse( $list->isEmpty() );
```
Benestar: Maybe assert more than this?

thiemowmde: Can also be assertCount, an assertion checking for hasGroupForLanguage, an equals check with a clone or a second object, … I prefer to keep such code as simple as possible. 1. Setup non-empty list. 2. Trigger remove. 3. Should still be non-empty. Even assertTrue( true ) would be valid here, because this is a no-op test.

PR #645, opened 2016-02-25
Simplify and clean up AliasGroupListTest assertions

thiemowmde:
- Use more strict `assertSame` whenever possible.
- ~~Use isEmpty when possible, instead of comparing objects.~~
- In `testGivenEmptyGroups_constructorRemovesThem` two objects were compared, but it was never checked if they are empty.

### DIFF
--- a/tests/unit/Term/AliasGroupListTest.php
+++ b/tests/unit/Term/AliasGroupListTest.php
@@ -48,7 +48,7 @@ class AliasGroupListTest extends PHPUnit_Framework_TestCase {
 
 		$list = new AliasGroupList( $array );
 
-		$this->assertEquals( $array, iterator_to_array( $list ) );
+		$this->assertSame( $array, iterator_to_array( $list ) );
 	}
 
 	public function testGivenGroupsWithTheSameLanguage_onlyTheLastOnesAreRetained() {
@@ -85,7 +85,7 @@ class AliasGroupListTest extends PHPUnit_Framework_TestCase {
 		 */
 		foreach ( $list as $key => $aliasGroup ) {
 			$this->assertEquals( $group, $aliasGroup );
-			$this->assertEquals( $aliasGroup->getLanguageCode(), $key );
+			$this->assertSame( $aliasGroup->getLanguageCode(), $key );
 		}
 	}
 
@@ -159,7 +159,7 @@ class AliasGroupListTest extends PHPUnit_Framework_TestCase {
 
 		$list->removeByLanguage( 'en' );
 
-		$this->assertEquals( new AliasGroupList(), $list );
+		$this->assertTrue( $list->isEmpty() );
 	}
 
 	/**
@@ -172,20 +172,13 @@ class AliasGroupListTest extends PHPUnit_Framework_TestCase {
 	}
 
 	public function testGivenEmptyGroups_constructorRemovesThem() {
-		$enGroup = new AliasGroup( 'en', [ 'foo' ] );
-
 		$list = new AliasGroupList( [
 			new AliasGroup( 'de' ),
-			$enGroup,
-			new AliasGroup( 'en' ),
-			new AliasGroup( 'nl' ),
-		] );
-
-		$expectedList = new AliasGroupList( [
+			new AliasGroup( 'en', [ 'foo' ] ),
 			new AliasGroup( 'en' ),
 		] );
 
-		$this->assertEquals( $expectedList, $list );
+		$this->assertTrue( $list->isEmpty() );
 	}
 
 	public function testGivenEmptyGroup_setGroupRemovesGroup() {
@@ -193,12 +186,10 @@ class AliasGroupListTest extends PHPUnit_Framework_TestCase {
 			new AliasGroup( 'en', [ 'foo' ] ),
 		] );
 
-		$expectedList = new AliasGroupList();
-
 		$list->setGroup( new AliasGroup( 'en' ) );
 		$list->setGroup( new AliasGroup( 'de' ) );
 
-		$this->assertEquals( $expectedList, $list );
+		$this->assertEquals( new AliasGroupList(), $list );
 	}
 
 	public function testEmptyListEqualsEmptyList() {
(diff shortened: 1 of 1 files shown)


tests/unit/Term/AliasGroupListTest.php
```diff
@@ -191,12 +184,10 @@ public function testGivenEmptyGroup_setGroupRemovesGroup() {
 			new AliasGroup( 'en', array( 'foo' ) ),
 		) );
 
-		$expectedList = new AliasGroupList();
-
 		$list->setGroup( new AliasGroup( 'en' ) );
 		$list->setGroup( new AliasGroup( 'de' ) );
 
-		$this->assertEquals( $expectedList, $list );
+		$this->assertTrue( $list->isEmpty() );
```
JeroenDeDauw: Now this test binds to more production code, and will also break when isEmpty is broken

thiemowmde: You know how `assertEquals` works? It compares **all** implementation details of the two objects. How is this better? All this test should care about is if the group is removed. It should not depend on everything else that can possibly happen in the object.

JeroenDeDauw: We do this on many places. Do you think we should add `equals` to all objects for which we do and use that instead?

Another difference is that when the new assertion here fails, you'll get "failed asserting that false is true", rather than a diff.

thiemowmde: There is only 1 assertion in this test. All you need to know is the name of the test.

We shared this argument already, at least one time. I think relying on the dark magic of `assertEquals` is bad and should be avoided whenever possible. Very similar to us avoiding the dark magic of PHP's `==`.

Adding an `equals` method does not change the argument that all this test needs to know is if the element was removed, which is perfectly expressed by a check for `isEmpty`.

JeroenDeDauw: Thanks for explaining your opinion. I do not agree that `assertEquals` should be avoided.

thiemowmde: I'm sorry? So you withdraw your original argument to avoid "binding to more production code"? Because `assertEquals` binds to **all** production code.

JeroenDeDauw: I don't agree with that. You can remove all methods apart from the constructor and it will still work. Conversely, if you use isEmpty, and it gets removed, these tests will break.

I cannot remember running into problems due to using `assertEquals`. I can remember running into problems due to tests using production code not under test.

thiemowmde: Shorter, simpler, more readable, easier to understand, more straightforward, not bound to implementation details that don't matter for that test, not checking things that are not relevant for that test, easier to maintain, … `assertEquals` on the other side gave me as much trouble as using `==` gave us. We agree on the later, don't we? `assertEquals` does not compare arrays the way you expect. It does not compare strings the way you expect. It compares way to many non-relevant implementation details when used on objects. Slow. Unclear. Clunky. Unpredictable in wort cases.

thiemowmde:
Partly reverted to get this moving.

tests/unit/Term/AliasGroupListTest.php
```diff
@@ -170,20 +171,13 @@ public function testGivenInvalidLanguageCode_removeByLanguageThrowsException( $l
...
-			new AliasGroup( 'nl' ),
-		) );
-
-		$expectedList = new AliasGroupList( array(
+			new AliasGroup( 'en', array( 'foo' ) ),
 			new AliasGroup( 'en' ),
 		) );
 
-		$this->assertEquals( $expectedList, $list );
+		$this->assertEquals( new AliasGroupList(), $list );
```
JeroenDeDauw: Looks like now this no longer tests if it only removes the empty ones

thiemowmde: This test is the whole reason why I touched this code. Multiple issues:

1. `$expectedList` was empty, because it was constructed with an empty group. This means the assertion compared an empty list with an empty list. This patch does not change that.
2. Sure, we could add a non-empty group to both constructors. Again, multiple issues with that:
 1. There are many other tests that already check if the constructor accepts non-empty groups. We do not need to repeat this assertion again, in a test that's about something else.
 2. The assertion would compare code with itself. If one constructors stops accepting non-empty groups, the other would do the same and the assertion would succeed. Which is, by the way, an other reason why I believe comparing objects with `assertEquals` is bogus. A test that compares the results of two constructors (or two identical method calls) with `assertEquals` is not worth much because of this issue. An assertion using `isEmpty` or `count` would be much more reliable in this specific example.

thiemowmde: Here is a good one: What's the most minimal implementation to make this test succeed?
```php
public function testGivenSetLanguage_removeByLanguageRemovesIt() {
	$list = new AliasGroupList( array( new AliasGroup( 'en', array( 'foo', 'bar' ) ) ) );

	$list->removeByLanguage( 'en' );

	$this->assertEquals( new AliasGroupList(), $list );
}
```
I will show you:
```php
class AliasGroupList {
	public function removeByLanguage( $languageCode ) {
	}
}
```

JeroenDeDauw: In reply to the previous comment: this test was not the first one written. So it's unclear to me how this was an issue when it was written or is remotely likely to be one in the future.

thiemowmde: What are you missing?

This test does not test what it says it tests.

1. The name suggests this is about the removal of empty groups in the constructor.
2. The assertion compares two objects that happen to be equal. Could be because they are both empty, could be because they both contain an empty group for `en`.

I found this an obvious mistake with an obvious fix when I uploaded this patch. So again, what are you missing?

JeroenDeDauw: FYI: I'm not looking into this further.

thiemowmde: This was meant as an obvious cleanup for an obvious mistake, and to make fragile tests more robust. Favor `assertSame` when possible, because it is more robust. Favor assertions that compare the results of two **different** calls, because this is more robust. Now this patch does have 15 comments, but they all fail to tell me what I should do to get this unstuck. Resubmit it? Split it? But how? Revert more? But what? I find this highly frustrating.

PR #648, opened 2016-02-29
Deprecate Holder interfaces

Benestar:
As decided in [T126152](https://phabricator.wikimedia.org/T126152)

Bug: [T128362](https://phabricator.wikimedia.org/T128362)

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,9 @@
 # Wikibase DataModel release notes
 
+## Version 5.1 (dev)
+
+* Deprecated `FingerprintHolder` and `StatementListHolder`
+
 ## Version 5.0.2 (2016-02-23)
 
 * Fixed regression in `ReferenceList::addReference` and the constructor possibly adding too many objects
--- a/src/Statement/StatementListHolder.php
+++ b/src/Statement/StatementListHolder.php
@@ -6,6 +6,7 @@ namespace Wikibase\DataModel\Statement;
  * Interface for classes that contain a StatementList.
  *
  * @since 3.0
+ * @deprecated since 5.1
  *
  * @license GNU GPL v2+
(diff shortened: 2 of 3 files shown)


thiemowmde:
+2, but want to give other team members a chance to notice.

RELEASE-NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Wikibase DataModel release notes
 
+## Version 5.1 (dev)
```
thiemowmde: 5.1.0

JeroenDeDauw:
+2

src/Statement/StatementListHolder.php
```diff
@@ -6,6 +6,7 @@
  * Interface for classes that contain a StatementList.
  *
  * @since 3.0
+ * @deprecated since 5.1
```
thiemowmde: I just realized this missed a migration path. What should I use instead? And that would be "use StatementListProvider instead", but the necessary guarantee that the provided list can be manipulated is missing. Adding this guarantee to the StatementListProvider is a breaking change. So in my opinion, adding this deprecation before 6.0 does not make much sense, because there is no alternative before 6.0.

brightbyte: I agree. That means that 6.0 will be breaking without a smooth migration path. You either migrate your code to 6.0, your you don't. Which is fine - that's what major releases are for.

brightbyte:
-2, don't merge. What Thiemo said: deprecation is pointless if there is no alternative. An advanced warning would be in order if there was a huge user base of this code. But I don't really see that. So let's just do the breaking change in a major release, and be done with it.

JeroenDeDauw:
This has already been merged

thiemowmde:
Personally, I'm fine with the deprecation, but only with the comment I added in #651, describing the migration path.

PR #649, opened 2016-02-29
Add clear to list classes

Benestar:
This is needed as a replacement for users of the holder interfaces to remove all elements from one of the given list classes.

[Bug: T128363](https://phabricator.wikimedia.org/T128363)

### DIFF
--- a/src/Statement/StatementList.php
+++ b/src/Statement/StatementList.php
@@ -342,6 +342,15 @@ class StatementList implements IteratorAggregate, Comparable, Countable {
 	}
 
 	/**
+	 * Removes all statements from this list.
+	 *
+	 * @since 6.0
+	 */
+	public function clear() {
+		$this->statements = [];
+	}
+
+	/**
 	 * @see http://php.net/manual/en/language.oop5.cloning.php
 	 *
 	 * @since 5.1
--- a/src/Term/AliasGroupList.php
+++ b/src/Term/AliasGroupList.php
@@ -204,4 +204,13 @@ class AliasGroupList implements Countable, IteratorAggregate {
 		return $array;
 	}
 
+	/**
+	 * Removes all alias groups from this list.
+	 *
+	 * @since 6.0
+	 */
+	public function clear() {
+		$this->groups = [];
+	}
+
 }
--- a/src/Term/TermList.php
+++ b/src/Term/TermList.php
@@ -187,4 +187,13 @@ class TermList implements Countable, IteratorAggregate, Comparable {
 			&& $this->terms[$term->getLanguageCode()]->equals( $term );
 	}
 
+	/**
+	 * Removes all terms from this list.
+	 *
+	 * @since 6.0
+	 */
+	public function clear() {
+		$this->terms = [];
+	}
+
 }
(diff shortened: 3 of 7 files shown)


src/Term/AliasGroupList.php
```diff
@@ -216,4 +216,13 @@ public function toTextArray() {
 		return $array;
 	}
 
+	/**
+	 * Removes all alias groups from this list.
+	 *
+	 * @since 5.1
+	 */
+	public function clear() {
```
thiemowmde: There is `Fingerprint::setAliasGroups`, so we should not need this.

src/Term/TermList.php
```diff
@@ -190,4 +190,13 @@ public function hasTerm( Term $term ) {
 			&& $this->terms[$term->getLanguageCode()]->equals( $term );
 	}
 
+	/**
+	 * Removes all terms from this list.
+	 *
+	 * @since 5.1
+	 */
+	public function clear() {
```
thiemowmde: Same here, there is `Fingerprint::setLabels` and `setDescriptions`, so this should not be needed. At least not at the moment. Did I missed a proposed use case?

Benestar: Yes, when we have entities that don't contain a `Fingerprint` but only a `TermList` or an `AliasGroupList`. We definitely need the `clear` method there.

JeroenDeDauw:
Looks good. Could perhaps also test that an empty list does not change state. But not that important

thiemowmde:
@Benestar and I had a look in our code base. Do we really need these methods, and where? What we found:
* EntityChangeFactory clears statements, descriptions and aliases for performance reasons. This is a hack, see T113468.
* Deserializers are currently written in a way they first create an entity and then set all labels, descriptions and aliases in one step. This can be rewritten: For example, first deserialize all labels, construct a Fingerprint from that, and then construct an Item from that. No clear nor "setTermList" or something like that needed, only constructors.
* There are "patchers" that manipulate terms in a very, very complicated way. This can be refactored. A label patcher, for example, should call removeLabel, setLabel, and nothing else. It should never create a new TermList and replace the old one. It#s not a "patcher" if it does this.

Benestar:
On hold, should not block 5.1.0

thiemowmde:
* Do we need this in 6.0?
* Or is this the opposite of #614 and should be closed when we merge the other one?

thiemowmde:
[What I wrote above](https://github.com/wmde/WikibaseDataModel/pull/649#issuecomment-190278907) is still true. There is still code that needs to clean these list classes. Refactoring all this code in a way that it does not need to do this is a lot of effort, and probably not really worth it.

I strongly supported #614 because it should not be possible to "repurpose" an existing entity and turn it into something completely different. There really should not be an `Item::clear`, on no entity type. But this here is very different. These are list classes. It's not really "repurposing" if code cleans a list of aliases.

Considering the odd fact that I just implemented a `clean` in https://github.com/wmde/WikibaseDataModelServices/pull/157 I suggest to merge this patch now. Pinging @brightbyte for the sake of completeness.

src/Statement/StatementList.php
```diff
@@ -342,6 +342,15 @@ public function filter( StatementFilter $filter ) {
 	}
 
 	/**
+	 * Removes all statements from this list.
+	 *
+	 * @since 6.0
```
manicki: this should have been 7.0 (or maybe 6.4) I guess

thiemowmde: Thanks! Fixed in #719.

PR #650, opened 2016-02-29
Use PropertyIdProvider in ByPropertyIdArray

thiemowmde:
Minimal patch that's guaranteed to not be a breaking change.

### DIFF
--- a/src/ByPropertyIdArray.php
+++ b/src/ByPropertyIdArray.php
@@ -5,6 +5,7 @@ namespace Wikibase\DataModel;
 use ArrayObject;
 use OutOfBoundsException;
 use RuntimeException;
+use Traversable;
 use Wikibase\DataModel\Entity\PropertyId;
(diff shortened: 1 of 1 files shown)


src/ByPropertyIdArray.php
```diff
@@ -52,7 +53,7 @@ class ByPropertyIdArray extends ArrayObject {
 	/**
```
mariushoch: `PropertyIdProvider[]|null`?

thiemowmde: It's an array of arrays of PropertyIdProvider. A possible type hint for that would be `PropertyIdProvider[][]`, but since not all tools understand that `array[]` is correct.

mariushoch: Oh right, it's an array of PropertyIdProvider array… I got that wrong.

PR #651, opened 2016-03-02
Prepare release 5.1

thiemowmde:


brightbyte:
+1: Looks good to me, but I don't have much experience with component releases, so I may have overlooked something.

JeroenDeDauw:
There is a presumable objection in #648. From my side this is fine though, so +1.

I probably would not make a release just for such a deprecation, esp not in a component that is actively developed anyway. However if you think it's worth the effort (which granted, is not all that much either), go with it.

PR #653, opened 2016-03-09
Add missing @param doc

JeroenDeDauw:


Benestar:
Nice catch

thiemowmde:
Huh? -2. We discussed this extensively in #348 and decided there can not be a type hint here, because this causes an LSP violation in all subclasses. Revert?

PR #654, opened 2016-03-10
Remove reindexing feature from ReferenceList::addReference

thiemowmde:
This behavior was inconsistent at least since 4.x, if not forever. I added these tests just recently in #641. They tested the current, but undocumented (and I believe unintentional) behavior: When adding the **same** reference a second time, it was possible to move it to a position with a smaller index, but not to a higher index. My first attempt was to fix the later. After reconsideration I'm now removing this reindexing feature entirely.

Technically breaking, but unused and therefore of no consequence in our relevant code bases.

### DIFF
--- a/src/ReferenceList.php
+++ b/src/ReferenceList.php
@@ -71,9 +71,15 @@ class ReferenceList implements Comparable, Countable, IteratorAggregate, Seriali
 			return;
 		}
 
+		$splHash = spl_object_hash( $reference );
+
+		if ( array_key_exists( $splHash, $this->references ) ) {
+			return;
+		}
+
 		if ( $index === null || $index >= count( $this->references ) ) {
 			// Append object to the end of the reference list.
-			$this->references[spl_object_hash( $reference )] = $reference;
+			$this->references[$splHash] = $reference;
 		} else {
 			$this->insertReferenceAtIndex( $reference, $index );
 		}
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
Tests failed. What's the deal with this?

JeroenDeDauw:
Mutation test is indicating that the code on master is not fully tested. Not going to touch it in favor of what you started here.

```
10) \Humbug\Mutator\Boolean\LogicalOr
Diff on \Wikibase\DataModel\ReferenceList::addReference() in /home/j/workspace/WikibaseDataModel/src/ReferenceList.php:
--- Original
+++ New
@@ @@
 
-		if ( $index === null || $index >= count( $this->references ) ) {
+		if ( $index === null && $index >= count( $this->references ) ) {
 			// Append object to the end of the reference list.
 			$this->references[spl_object_hash( $reference )] = $reference;
 		} else {
 			$this->insertReferenceAtIndex( $reference, $index );
 		}
 	}


11) \Humbug\Mutator\ConditionalBoundary\GreaterThanOrEqualTo
Diff on \Wikibase\DataModel\ReferenceList::addReference() in /home/j/workspace/WikibaseDataModel/src/ReferenceList.php:
--- Original
+++ New
@@ @@
 
-		if ( $index === null || $index >= count( $this->references ) ) {
+		if ( $index === null || $index > count( $this->references ) ) {
 			// Append object to the end of the reference list.
 			$this->references[spl_object_hash( $reference )] = $reference;
 		} else {
 			$this->insertReferenceAtIndex( $reference, $index );
 		}
 	}
```

thiemowmde:
I believe the results of the mutation tests are not relevant for this pull request. Note that I changed the fix entirely. Instead of "fixing" a behavior that is undocumented and unintentional anyway, I'm removing it now. I believe this fix is fine and should go into master now.

PR #659, opened 2016-04-07
Add missing PropertyIdProvider types

thiemowmde:


tests/unit/ByPropertyIdArrayTest.php
```diff
@@ -299,14 +299,10 @@ public function moveProvider() {
...
 	 * @dataProvider moveProvider
-	 * @param object[] $objectsSource
-	 * @param object $object
-	 * @param int $toIndex
-	 * @param object[] $objectsDestination
 	 */
 	public function testMoveObjectToIndex(
 		array $objectsSource,
-		$object,
+		PropertyIdProvider $object,
```
mariushoch: Could also directly hint against `Statement` here.

thiemowmde: With the current provider yes, but this may change. I would like to keep `PropertyIdProvider` for consistency with the other tests in this file.

PR #661, opened 2016-04-09
Switched arrays over to short array syntax

JeroenDeDauw:
Includes the necessary bump of the required PHP version


thiemowmde:
Doesn't this need a 7.0.0 version bump because it's a breaking change?

JeroenDeDauw:
Nope http://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api

mariushoch:
+1, diff looks good at a glance.

brightbyte:
needs rebase. any takers?

JeroenDeDauw:
Run script again, much easier than rebase. And don't do this while other work is going on in the component...

PR #662, opened 2016-04-21
Fix newline injection vector in EntityId validation

thiemowmde:
[Bug: T133299](https://phabricator.wikimedia.org/T133299)


mariushoch:
Should be documented in the recent changes.

PR #665, opened 2016-04-28
Block ByPropertyIdArray construction from object properties

thiemowmde:
I keep running into this problem. When you pass a StatementList to this class, which is what it is build for, it iterates the **properties** of the StatementList object, not the statements.

I'm aware this is deprecated and should die entirely. That's why I'm going for the most simple fix I can think of.

### DIFF
--- a/src/ByPropertyIdArray.php
+++ b/src/ByPropertyIdArray.php
@@ -3,6 +3,7 @@
 namespace Wikibase\DataModel;
 
 use ArrayObject;
+use InvalidArgumentException;
 use OutOfBoundsException;
 use RuntimeException;
 use Traversable;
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
+1

I'd class this as an enhancement of input validation, not a bug. Else we have a lot of very buggy code :)

src/ByPropertyIdArray.php
```diff
@@ -51,11 +52,18 @@ class ByPropertyIdArray extends ArrayObject {
...
 	/**
+	 * @deprecated since 5.0, use a DataModel Service instead
 	 * @see ArrayObject::__construct
 	 *
 	 * @param PropertyIdProvider[]|Traversable|null $input
+	 *
+	 * @throws InvalidArgumentException
 	 */
 	public function __construct( $input = null ) {
+		if ( is_object( $input ) && !( $input instanceof Traversable ) ) {
```
brightbyte: Assert::parameterType( 'array|Traversable', $input, '$input' )

PR #666, opened 2016-04-28
Deprecate Statement::getWithUniqueMainSnaks

thiemowmde:


JeroenDeDauw:
Why this deprecation? I have some code that uses this.

thiemowmde:
Based on the name of the method and it's implementation I really don't get what it's supposed to do. It's some kind of filter. It returns a filtered StatementList, but with statements removed that happen to have identical main snaks. Only one statement of such a group remains, but which one remains depends on the position in the list.

This is almost identical to getMainSnaks, except that you can access the qualifiers, references and rank of the first statement with that main snak. But you don't know anything about the other qualifiers, references and ranks of the other statements with the same main snak. How is this useful?

How is this so useful that is should be in this base implementation? One can easily build this simple loop in the application where it is needed.

I think the same could be said about other methods, but this is the one that makes me wonder the most.

JeroenDeDauw:
One of the main reasons for having this class is so we can have encapsulated filter methods. The one you are deprecating is real small, maintenance cost is basically not there. The method makes sense to me. Even if it does not make sense to you, do you think it's worthwhile removing?

thiemowmde:
It's just not self-explaining. Given all the possible confusion I tried to outline in my comment above, what is the actual use of this filter? As I wrote, what it really does is filtering by main snak, but then returning the first statement that happens to have the same main snak. The only difference to getMainSnaks is that you can access the qualifiers, references and rank of that "random" statement. But code that does that and ignores the ranks, qualifiers and references of the other statements with the same main snak is almost certainly doing something wrong.

We do have a `StatementFilter` framework. My impression is that the use case you are targeting with this filter should be a StatementFilter or, as said, a simple foreach loop in https://github.com/JeroenDeDauw/WikibaseQueryEngine/blob/master/src/SQLStore/EntityStore/BasicEntityInserter.php#L37, which seems to be the only user of this method.

thiemowmde:
I still want to understand the use cases for this method that can not be solved with `getMainSnaks`. As explained above the semantics of `getWithUniqueMainSnaks` are so unpredictable (it's like picking random statements that happen to have identical main snaks) that I have to assume all usages of this are unintentional, if not wrong.

I can find a single user [in WikibaseQueryEngine](https://github.com/wmde/WikibaseQueryEngine/blob/696117d92e4a99fa779d448435ea6d860f58610a/src/SQLStore/EntityStore/BasicEntityInserter.php#L40), and it does have this exact problem of being unpredictable. If an entity contains two different statements (different property, different rank, different qualifiers, different references), but with identical main snaks (which can easily happen, e.g. think of a women who gave birth to a child and died the exact same day), this date is only indexed a single time with whatever property, rank, qualifiers and references came last in the statements list. It should be pretty easy to write a test that makes this issue obvious.

I still vote for deprecating and removing this method.

PR #667, opened 2016-05-10
Add NumericEntityId interface

thiemowmde:
It's important that this interface is not added to EntityId. Not all entities can be expressed as numeric ids. This patch does not change this. What it does is basically a nice alternative for an `method_exists` check. Right now we call `getNumericId` 39 times in production and 26 times in tests. It will take us months to get rid of all these. This patch here may help us get moving.

For discussion, mainly with @brightbyte and @adrianheine.

[Bug: T135650](https://phabricator.wikimedia.org/T135650)

### DIFF
--- a/src/Entity/ItemId.php
+++ b/src/Entity/ItemId.php
@@ -10,7 +10,7 @@ use InvalidArgumentException;
  * @license GPL-2.0+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
  */
-class ItemId extends EntityId {
+class ItemId extends EntityId implements NumericEntityId {
 
 	/**
 	 * @since 0.5
@@ -38,6 +38,8 @@ class ItemId extends EntityId {
 	}
 
 	/**
+	 * @since 0.5
+	 *
 	 * @return int
 	 */
 	public function getNumericId() {
--- a/src/Entity/NumericEntityId.php
+++ b/src/Entity/NumericEntityId.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Wikibase\DataModel\Entity;
+
+/**
+ * Interface for EntityIds that can be converted to numbers, and back from the entity type and the
+ * number. Entity types that do not meet this criteria (e.g. when page titles or file names are used
+ * as IDs) should not implement this interface. Never return a fallback value like 0 that's the same
+ * for different IDs!
+ *
+ * Entity types are not required and not guaranteed to implement this interface. Use the full string
+ * serialization whenever you can and avoid using numeric IDs.
+ *
+ * @since 6.1
+ *
+ * @see EntityId::getSerialization
+ */
+interface NumericEntityId {
+
+	/**
+	 * @since 6.1
+	 *
+	 * @return int|float Numeric representation of the EntityId as a positive integer number in the
+	 *  [1..PHP_INT_MAX] range. A float when exceeding this range (e.g. when exceeding 2,147,483,647
+	 *  on a 32 bit system).
+	 */
+	public function getNumericId();
+
+}
--- a/src/Entity/PropertyId.php
+++ b/src/Entity/PropertyId.php
@@ -10,7 +10,7 @@ use InvalidArgumentException;
  * @license GPL-2.0+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
  */
-class PropertyId extends EntityId {
+class PropertyId extends EntityId implements NumericEntityId {
 
 	/**
 	 * @since 0.5
@@ -38,6 +38,8 @@ class PropertyId extends EntityId {
 	}
 
 	/**
+	 * @since 0.5
+	 *
 	 * @return int
 	 */
 	public function getNumericId() {


JeroenDeDauw:
Does this not just change `method_exists` for `instanceof` checks?

I killed most usages of `getNumericId` way back, and spend some effort on removing less easy cases afterwards. When I stopped with this, the remaining cases where often not worthwhile to change, or simply not possible without significant refactoring. The need to these changes might have become greater for some of these with the new entity types tho. I'm wondering if all usages will end up being removed though, and if this will be in a few months - that might not be as easy or as sensible as it appears.

JeroenDeDauw:
-2 to the deprecation of the `getNumericId` methods. I think it's entirely fine to call `getNumericId` on an ItemId, as it is guaranteed to have a numeric part.

The interface I would not deprecate either (and thus not have Deprecated in the name), and instead add a note that using it in certain situations is bad. There are good ones tho, for instance when you want to index just items and properties. I have code that does such things. -1 to that part.

thiemowmde:
> it's entirely fine to call getNumericId on an ItemId

I agree. I just discussed this with @adrianheine and modified my original proposal.

1. We make MediaInfoId implement getNumericId and wrap remaining calls to this method into `method_exists`, throwing an exception when an entity type does not support numeric ids. Advantages:
 * This will not make EntityDocument or EntityId worse.
 * For now we can move fast.
 * We can refactor code later, when one of the "this piece of code does not support non-numeric entity ids" exceptions hits us.
 * All 3 entity types we have are numeric anyway.
 * Some code that uses numeric IDs, e.g. in PropertySuggester, is not used on other entity types anyway.
2. The same, but with `instanceof NumericEntityIdProvider` instead of `method_exists`.

thiemowmde:
I reworked this patch as discussed.

src/Entity/NumericEntityIdProvider.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace Wikibase\DataModel\Entity;
+
+/**
+ * Common interface for EntityIds that can be converted to numbers, and back from the entity type
```
JeroenDeDauw: "Common"?

src/Entity/NumericEntityIdProvider.php
```diff
@@ -0,0 +1,29 @@
...
+ * and the number. Entity types that do not met this criteria (e.g. when page titles or file names
```
JeroenDeDauw: meet

src/Entity/NumericEntityIdProvider.php
```diff
@@ -0,0 +1,29 @@
...
+ * are used as IDs) should not implement this interface. Never return a fallback value like 0
+ * that's the same for multiple IDs!
```
JeroenDeDauw: The last sentence here is not grammatically correct, and is redundant considering the last line. I suggest removing.

src/Entity/NumericEntityIdProvider.php
```diff
@@ -0,0 +1,29 @@
...
+ * @see EntityId::getSerialization
+ */
+interface NumericEntityIdProvider {
+
+	/**
+	 * @since 6.1
+	 *
+	 * @return int|float Numeric representation of the EntityId as a positive integer number in the
+	 *  [1..PHP_INT_MAX] range. A float when exceeding this range (e.g. when exceeding 2,147,483,647
+	 *  on a 32 bit system).
```
JeroenDeDauw: >_> such PHP

Will using `int` s type hint here cause problems on 32 bit systems, even if PHP is using a float for it? ie, will `is_int` checks fail in this special case? (Or the real `int` type hint in modern PHP for that matter)

JeroenDeDauw:
I suggest to rename to `NumericEntityId`, by dropping the `Provider` suffix. +2 when that is done.

thiemowmde:
Renamed to NumericEntityId and addressed the comments.

> redundant considering the last line.

I find it helpful to repeat this as an explicit "don't". Otherwise an implementation may do a naive `(int)` cast, which results in 0 for strings that do not start with a number.

> will is_int checks fail in this special case?

is_int fails on floats, as it should.

Code calling getNumericId should expect to get a float and never cast to int. When it does, numbers larger than 2147483647 (9 digits) will all become 2147483647.

Floats also have a limit, which starts at `"9007199254740993"` (16 digits). That string becomes 9007199254740992.0 (note the 2 at the end) when casted to float.

> did someone check all other relevant places for the same issue?

Yes, I did. In the past two years I added test cases that exceed the 32 bit PHP_INT_MAX to quite some places, reworked code that did naive `(int)` casting and added comments to make people aware of this limitation.

brightbyte:
@thiemowmde 
> Code calling getNumericId should expect to get a float and never cast to int. When it does, numbers larger than 2147483647 (9 digits) will all become 2147483647.

I don't think that is a good idea - getNumericId should be defined to be int. The reason is practical: we *only* need the numeric ID only for use with database tables that have a numeric ID column. That column is defined to be an INT, which MySQL defined to be 32 bit. So anything beyond 32 integer range will fail anyway.

thiemowmde:
Accepting data loss and designing an interface after that does not make any sense to me. The components we are discussing here do not and should neither know nor care that they are used wrong.

That 32 bit limit is not purely hypothetical. When wikidata.org becomes massively popular, we will hit this limit in a few years.

JeroenDeDauw:
> integer / float stuff

:( Can't we just require 64 bit? This ain't the 90s any more...

Tpt:
> :( Can't we just require 64 bit? This ain't the 90s any more...

I believe that we could assume that most of PHP installs are 64bits now
and, so, that their integers are on 64 bitd. And it's going to be even more
true in the future.
Le 13 mai 2016 2:47 AM, "Jeroen De Dauw" <notifications@github.com> a
écrit :

thiemowmde:
I disagree. Look at IPv4. Look at Y2K. Look at [Y2K38](https://en.wikipedia.org/wiki/Year_2038_problem). Look at [YouTubes counter hitting the 32 bit limit](http://arstechnica.com/business/2014/12/gangnam-style-overflows-int_max-forces-youtube-to-go-64-bit/).

Tpt:
@thiemowmde Yes, I agree that a 32bits limit is too low, and so that we will probably have to change the database format in the future. But I don't see the issue with using 64 bits integers.

JeroenDeDauw:
I suggest to create a new task on phab for the 32 bit / 64 bit thing as this is way more general than the type hint in this interface. Let's move this along with using `int` as we did before, and then holistically tackle the 32 bit / 64 bit thing?

thiemowmde:
I don't see how discussing this could help, and how this is relevant for this patch? You can not discuss if PHP should have an `int64` type. It does not have one. Falling back to floats is what PHP itself does. Using floats or strings and passing them to something like `sprintf( '%.0f', $float )` is what almost all our code already does. Not only when working with numeric IDs. This is actually critical when working with years.

I can reword the contract of this interface a bit, but I don't see how simply removing the only working solution and replacing it with nothing would do anything good.

brightbyte:
@Tpt 
>Yes, I agree that a 32bits limit is too low, and so that we will probably have to change the database format in the future. But I don't see the issue with using 64 bits integers.

No - we should not use integers in IDs at all. Any database change we would need to go from 32 to 64 bit ints we could use to move away from integers completely, and use full string IDs everywhere. 

This interface is being introduced to cater to the legacy fields in the database. It is not intended for any other use, and shouldn't be used for anything else. So, I don't see why it should support anythign but 32 bit integers. In fact, I think it shouldn't.

thiemowmde:
That makes an interface completely pointless. If that's all you want then just do a method_exists check and call it a day. Crippling what I propose here to nothing but the fact that a method with that name exists, and allow it to return whatever it wants, even the same return value for different EntityId objects, is actually harmful.

brightbyte:
@thiemowmde Perhaps I misunderstood your intent. Our original comment was:

> What it does is basically a nice alternative for an method_exists check. Right now we call getNumericId 39 times in production and 26 times in tests. It will take us months to get rid of all these. This patch here may help us get moving.

For that purpose, all we need, and all we want, is a 32 bit integer.

Why should we *add* support for bigger numeric IDs, if we want to get rid of numeric IDs anyway? What you are saying now  makes me think you want to establish and extend support for numeric ids. Why?

thiemowmde:
If the current `getNumericId` implementations are underspecified and bogus, let's either fix this (which is the only advantage this patch has over method_exists checks), or deprecate and remove them.

brightbyte:
@thiemowmde The current implementations is declared to return an int (which may be 32 or 64 bit in PHP). But they are generated *only* by SqlIdGenerator, which uses the wb_id_counters table, which has an INT field called id_value. 

So, the current interface is declared to be an int (underspecified in terms of bits - but this is true for all usages of "int" in any interface declaration anywhere in the code). The current code guarantees that entities that are actually saved to the database will always have a 32 bit numeric id. It's possible to create an ItemId with a number beyond the 32 bit range, yea. But if you try to save that to the database, it fails.

The existing usages to getNumericId that I am aware of all relate to database storage. That's why we still have them: because it's hard to change the DB schema. We want to fix the schema and get rid of numeric IDs eventually. I like the idea of having a proper interface to declare the status quo, but I see no reason for that interface to make guarantees that we can't even meet with the current code. If we want to allow getNumericId to return integers beyond the 32 bit range, we'd have to change the DB schema first.

thiemowmde:
I don't see how any of this is an argument for an underspecified interface and what else I should say what was not already said multiple times. SqlIdGenerator is in WikibaseRepo. DataModel neither depends on WikibaseRepo nor does it know anything about a database schema. All other methods (e.g. ItemId::newFromNumber) are already specified (via tests) like this: they do not use is_int and accept floats when the int32 range is exceeded. Nobody ever objected changes like #258.

brightbyte:
I'm not arguing for bing less specific. I'm arguing for being *more* specific: restrict numeric IDs (a legacy concept!) to 32 bits (that is, the values that actually work with WikibaseRepo). You can allow float IDs in the model, but how is that useful? What would that ever be used for? You'd  suggest that using floats (or very big integers) is actually in option, while in fact, it isn't. Things will break if you do that.

thiemowmde:
When we add this interface, it will never be marked as deprecated - following the arguments given in this thread. Entity ids that can be represented as "entity type and number" will implement this, and continue to do so. An "official" interface for this feature must be as specific as possible, which means that all known edge cases like the int32 issue we are talking about must be described somehow. Personally I can see the following solutions:
* When exceeding int32, getNumericId must throw an exception. Open questions: Which exception? How to detect that edge case?
* When an entity id can possibly exceed int32, it's now allowed to use this interface. The issue with this is that all our current entity id implementations can exceed int32, so this interface can not be used and it's pointless to introduce it.
* The most simple implementation can always return floats to avoid all int32 issues. But this does not mean the interface must require a float. It can still allow `int|float` as all other places dealing with numeric IDs in our code base do.
* We could somehow describe that this interface must be used on a 64 bit system with a 64 bit PHP version. I'm not sure how such a description would look like.

thiemowmde:
> I'm arguing for being more specific: restrict numeric IDs […] to 32 bits

This would only work if all entity id constructors throw exceptions when you try to construct, for example, a `new ItemId( 'Q2147483648' )`. Such an exception does not make much sense because it's supposed to be a string and not an int32.

I suggest to read http://php.net/manual/en/language.types.integer.php#language.types.integer.overflow again.

brightbyte:
I am arguing based on the following assumption: We are introducing this interface for one reason only: for interoperability with out existing DB schema. 

Is this assumption correct? I see no other reason to have such an interface, and I would indeed recommend to deprecate it right away. It's a compatibility clutch, and it does indeed bleed implementation knowledge of the wikibase db schema into the data model component. That's not nice, but that's how it is.

You can define the interface to do all kinds of things, but in the end, if you actually put anything in here that exceeds the 32 bit unsigned range, you will not be able to store it in the database. If you try, things will break. 

So, to be useful for its original purpose, the value returned by getNumbericId needs to be in the range that is actually supported by the database. That's the need we have, no matter whether it's nice to have this knowledge in this component (it's not). 

This, the interface should guarantee that the value returned by getNumericId is in the 32 bit unsigned int range. If and how this is actually assured is up to the implementation. Checking in the constructor would be nice, but it's not a requirement.

thiemowmde:
I'm repeating myself for the like fifth time: This is not Wikibase.git. This component does not know anything about databases. This is an interface for entity IDs that can be represented as "entity type and number". This is not an implementation detail from Wikibase.git. This is a property of an entity ID. This is the inverse of newFromNumber. We can discuss what "number" means, as I tried in my last post, but arguing with implementation details from Wikibase.git is not helpful.

brightbyte:
@thiemowmde PlLease provide a reason we want this interface, if not because of implementation details in Wikibase.

adrianheine:
@brightbyte Because of implementation details in WikibasePropertySuggester.

thiemowmde:
Why are you repeating something I already replied to multiple times? Again: This component does **not** depend on Wikibase.git. If all you want is a method_exists check then please do this and be done. This can not be used as an argument against the interface proposed here.

Again: Some entity IDs can be and are represented as "entity type and number". "Q1" is literally an entity type identifier and a number. newFromNumber allows constructing this, getNumericId allows deconstructing this. This and only this is all one needs to know to review this pull request.

brightbyte:
What is the benefit of having this interface? Where will it be used, and for what purpose?

thiemowmde:
We are going in circles. All this was already said multiple times.

brightbyte:
@thiemowmde I don't see any concrete use cases explained anywhere. The one use case I have in mind will not work if this is not restricted to 32 bits.

thiemowmde:
```php
$number = $id->getNumericId();
if ( !is_int( $number ) ) {
    throw new OutOfBoundsException( 'Integer limit exceeded' );
}
// proceed
```

thiemowmde:
How does this work better with an interface that silently starts returning 2147483647 for all IDs that exceed PHPs integer limit? Such an interface is broken and will get a straight -2 any time.

No other alternative was given throughout this whole thread.

brightbyte:
If we have to do a range check everywhere this is used in a DB context (which, I supspect, are all the usages we have left), then the interface is pointless. Also note that is_int isn't sufficient. It has to be a positive integer in the 32 bit range. 

The alternative is to let the interface require the ID to be in the range 0..2^32-1. How and when the implementation makes sure this is the case is another question. newFromNumber can easily check this. Checking it in the constructor is a bit annoying, but possible. Checking in getNumericId is a bit late, and throwing an exception there isn't too pretty, but it's still better than having manual checks all over the place. 

Again: the ONLY use case we have is to cater to a legacy db schema, and that schema ONLY supports positive 32 bit  ints.

thiemowmde:
You are more or less repeating what I already wrote in https://github.com/wmde/WikibaseDataModel/pull/667#issuecomment-219747387. What's your proposal?

Do you realize what we are discussing here? The wording of a `@return` tag comment that may become relevant in a hundred years. Can we please get this moving? Throwing an exception is not horrible (implementations would be more or less the same, but throw an exception instead of switching to float), but really odd and bogus given that newFromNumber supports floats and absolutely nothing else in this component ever specified an int32 limit. I really don't see why we must come up with such an out-of-place concept (that actually encodes Wikibase.git knowledge into an independent component!) just to end this.

I strongly suggest reading http://php.net/manual/en/language.types.integer.php#language.types.integer.overflow again.

thiemowmde:
Obsolete via #670.

PR #669, opened 2016-05-11
Add id field to EntityIdValue serialization

thiemowmde:
The most ugly bit of this patch is the static `EntityIdValue::setEntityIdParser` method. I explored a few possibilities and decided this one is clean enough, without ripping everything apart.
* The relevant caller of this `newFromArray` method (it's not even in an interface!) is hidden in DataValueDeserializer::getDeserialization, see https://github.com/DataValues/Serialization/blob/master/src/Deserializers/DataValueDeserializer.php#L113. Changing this seems painful to me.
* Note that there is no default value for the ID parser. This is on purpose. Code should fail hard when you forget to set this, and not accept some but reject other entity types.
* Note that the legacy method calls the static `LegacyIdInterpreter::newIdFromTypeAndNumber`, which also does not accept new entity types. This means that user provided JSON like `{ "entity-type": "media-info", "numeric-id" => 1 }` can never succeed, even if `{ "id": "M1" }` passes. In other words: The legacy format is restricted to Items and Propertys. In my opinion this is a good thing.
* Thats also why I did not added checks for mismatches as in `{ "entity-type": "media-info", "numeric-id" => 1, "id": "Q2" }`. This can't be done without teaching the legacy code path all new entity types. Instead the legacy stuff is completely (and always) ignored when the new field is present.

[Fixes: T132592](https://phabricator.wikimedia.org/T132592)

### DIFF
--- a/src/Entity/EntityIdValue.php
+++ b/src/Entity/EntityIdValue.php
@@ -12,8 +12,16 @@ use Wikibase\DataModel\LegacyIdInterpreter;
  *
  * @license GPL-2.0+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ * @author Thiemo Mättig
  */
 class EntityIdValue extends DataValueObject {
+
+	/**
+	 * @since 6.1
+	 *
+	 * @var EntityIdParser|null
+	 */
+	private static $idParser;
 
 	private $entityId;
 
@@ -43,7 +51,7 @@ class EntityIdValue extends DataValueObject {
 	 *
 	 * @return float Numeric id as a whole number. Can not be int because of 32-bit PHP.
 	 */
-	protected function getNumericId() {
+	private function getNumericId() {
 		return floatval( substr( $this->entityId->getSerialization(), 1 ) );
 	}
 
@@ -121,7 +129,17 @@ class EntityIdValue extends DataValueObject {
 		return array(
 			'entity-type' => $this->entityId->getEntityType(),
 			'numeric-id' => $this->getNumericId(),
+			'id' => $this->entityId->getSerialization(),
 		);
+	}
+
+	/**
+	 * @since 6.1
+	 *
+	 * @param EntityIdParser $idParser
+	 */
+	public static function setEntityIdParser( EntityIdParser $idParser ) {
+		self::$idParser = $idParser;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
This `newFromArray` stuff is kinda dodgy. As you pointed out, its not in an interface, which is because it's static, and can't be. To not have to deal with this stuff, DataValueDeserializer and DataValueSerializer where added. Apparently I forgot to deprecate the newFromArray here. I suggest to continue down that migration path and mark the static method in EntityId as deprecated. The approach in this PR makes the crappy stuff work with new entity types, though in the process adds global state, adding to the tech debt.

thiemowmde:
I don't understand what you are suggesting. How does DataValueDeserializer help "to not have to deal with this stuff"? It calls newFromArray. In fact it's the only caller. What do you suggest to deprecate and how would that help us?

JeroenDeDauw:
Oh great if `DataValueDeserializer` is the only caller. That means just changes to that class and the code constructing it are needed.

You can add a second constructor parameter to `DataValueDeserializer`, that allows deserialization via something more sane than the `newFromArray` thing. For instance a map of data type id to callback, or to `Deserializer`. That was the general idea when I added this class, and still seems applicable.

thiemowmde:
It feels really weird to have this knowledge scattered across four (!) different components.

1. Serialization is declared in DataValue::getArrayValue (component "data-values/data-values") …
2. … and implemented in EntityIdValue::getArrayValue (component "wikibase/data-model"). I don't think there is anything wrong with this.
3. Sure, we could teach DataValueDeserializer (component "data-values/serialization") to accept callbacks.
4. This callback must live in WikibaseRepo::getDataValueDeserializer (component "wikibase/wikibase").

I don't know. With what I currently propose in this pull request 1, 2 and 4 are still involved, but only 2 and 4 must be changed. What I really want to avoid is to make the dependency hell of all the components that must be changed more painful. Hm. I will try to draft that change in DataValueDeserializer.

thiemowmde:
Done, see https://github.com/DataValues/Serialization/pull/14.

thiemowmde:
I will remove the static parser here when https://github.com/DataValues/Serialization/pull/14 is ok for you.

JeroenDeDauw:
I've already indicated that change is OK for me

src/Entity/EntityIdValue.php
```diff
@@ -138,27 +156,45 @@ public function getArrayValue() {
...
-		if ( !array_key_exists( 'numeric-id', $data ) ) {
-			throw new IllegalValueException( "'numeric-id' field required" );
+	/**
+	 * @param string $idSerialization
+	 *
+	 * @throws IllegalValueException
+	 * @return self
+	 */
+	private static function newFromIdSerialization( $idSerialization ) {
+		try {
```
brightbyte: What if self::$idParser is null? Would be nice to get a meaningful error message in that case, instead of a fatal error.

src/Entity/EntityIdValue.php
```diff
@@ -121,10 +129,20 @@ public function getArrayValue() {
...
+			'id' => $this->entityId->getSerialization(),
 		);
 	}
 
 	/**
+	 * @since 6.1
+	 *
+	 * @param EntityIdParser $idParser
+	 */
+	public static function setEntityIdParser( EntityIdParser $idParser ) {
```
brightbyte: Yea... probably the best we can do without overhauling the entire infrastructure. But we *should* overhaul the entire infrastructure. This sucks.

brightbyte:
@thiemowmde Now that DataValues/Serialization#14 is merged, can the static member no go away here?

thiemowmde:
This patch is most probably obsolete, see #671. But I did not closed it yet because both proposals do have issues and even if there is a tendency to favor #671 over this older proposal, I'm not yet sure if the issue in #671 is worth it.

src/Entity/EntityIdValue.php
thiemowmde: I did not wanted to waste time just for a nicer message. I wanted this do die hard.

brightbyte:
CR-2: The alternative pull request, #671, is now merged.

PR #670, opened 2016-05-20
Add Int32EntityId

thiemowmde:
Alternative proposal for [T135650](https://phabricator.wikimedia.org/T135650). The main difference to #667 is that this here is a breaking change while #667 is not. #667 will require consumers to do the int32 check, while this patch requires the same int32 check on construction time.

[Bug: T135650](https://phabricator.wikimedia.org/T135650)

### DIFF
--- a/src/Entity/Int32EntityId.php
+++ b/src/Entity/Int32EntityId.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace Wikibase\DataModel\Entity;
+
+/**
+ * Interface for EntityIds that can be converted to a unique, positive, signed 32 bit integer in the
+ * range [1..2147483647], and back from the entity type and the number. When an ID can not be
+ * represented as an unique integer, an InvalidArgumentException must be thrown on construction
+ * time. For example, when "Q1" and "Q01" have the same integer representation, only one should be
+ * allowed.
+ *
+ * Entity types that do not meet this criteria should not implement this interface.
+ *
+ * Entity types are not required and not guaranteed to implement this interface. Use the full string
+ * serialization whenever you can and avoid using numeric IDs.
+ *
+ * @since 6.1
+ *
+ * @license GPL-2.0+
+ * @author Thiemo Mättig
+ */
+interface Int32EntityId {
+
+	const MAX = 2147483647;
+
+	/**
+	 * @since 6.1
+	 *
+	 * @return int Guaranteed to be a unique integer in the range [1..2147483647].
+	 */
+	public function getNumericId();
+
+}
--- a/src/Entity/ItemId.php
+++ b/src/Entity/ItemId.php
@@ -10,12 +10,12 @@ use InvalidArgumentException;
  * @license GPL-2.0+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
  */
-class ItemId extends EntityId {
+class ItemId extends EntityId implements Int32EntityId {
 
 	/**
 	 * @since 0.5
 	 */
-	const PATTERN = '/^Q[1-9]\d*\z/i';
+	const PATTERN = '/^Q[1-9]\d{0,9}\z/i';
 
 	/**
 	 * @param string $idSerialization
@@ -29,11 +29,18 @@ class ItemId extends EntityId {
 
 	private function assertValidIdFormat( $idSerialization ) {
 		if ( !is_string( $idSerialization ) ) {
-			throw new InvalidArgumentException( '$idSerialization must be a string; got ' . gettype( $idSerialization ) );
+			throw new InvalidArgumentException( '$idSerialization must be a string' );
 		}
 
 		if ( !preg_match( self::PATTERN, $idSerialization ) ) {
 			throw new InvalidArgumentException( '$idSerialization must match ' . self::PATTERN );
+		}
+
+		if ( strlen( $idSerialization ) > 10
+			&& substr( $idSerialization, 1 ) > Int32EntityId::MAX
+		) {
+			throw new InvalidArgumentException( '$idSerialization can not exceed '
+				. Int32EntityId::MAX );
 		}
 	}
 
--- a/src/Entity/PropertyId.php
+++ b/src/Entity/PropertyId.php
@@ -10,12 +10,12 @@ use InvalidArgumentException;
  * @license GPL-2.0+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
  */
-class PropertyId extends EntityId {
+class PropertyId extends EntityId implements Int32EntityId {
 
 	/**
 	 * @since 0.5
 	 */
-	const PATTERN = '/^P[1-9]\d*\z/i';
+	const PATTERN = '/^P[1-9]\d{0,9}\z/i';
 
 	/**
 	 * @param string $idSerialization
@@ -29,11 +29,18 @@ class PropertyId extends EntityId {
 
 	private function assertValidIdFormat( $idSerialization ) {
 		if ( !is_string( $idSerialization ) ) {
-			throw new InvalidArgumentException( '$idSerialization must be a string; got ' . gettype( $idSerialization ) );
+			throw new InvalidArgumentException( '$idSerialization must be a string' );
 		}
 
 		if ( !preg_match( self::PATTERN, $idSerialization ) ) {
 			throw new InvalidArgumentException( '$idSerialization must match ' . self::PATTERN );
+		}
+
+		if ( strlen( $idSerialization ) > 10
+			&& substr( $idSerialization, 1 ) > Int32EntityId::MAX
+		) {
+			throw new InvalidArgumentException( '$idSerialization can not exceed '
+				. Int32EntityId::MAX );
 		}
 	}
 
--- a/tests/unit/Entity/EntityIdTest.php
+++ b/tests/unit/Entity/EntityIdTest.php
@@ -27,7 +27,7 @@ class EntityIdTest extends \PHPUnit_Framework_TestCase {
 		$ids[] = array( new ItemId( 'Q1' ) );
 		$ids[] = array( new ItemId( 'Q42' ) );
 		$ids[] = array( new ItemId( 'Q31337' ) );
-		$ids[] = array( new ItemId( 'Q2147483648' ) );
+		$ids[] = array( new ItemId( 'Q2147483647' ) );
 		$ids[] = array( new PropertyId( 'P101010' ) );
 
 		return $ids;
--- a/tests/unit/Entity/EntityIdValueTest.php
+++ b/tests/unit/Entity/EntityIdValueTest.php
@@ -40,8 +40,7 @@ class EntityIdValueTest extends PHPUnit_Framework_TestCase {
 			new ItemId( 'Q1' ),
 			new ItemId( 'Q42' ),
 			new ItemId( 'Q31337' ),
-			// Check for 32-bit integer overflow on 32-bit PHP systems.
-			new ItemId( 'Q2147483648' ),
+			new ItemId( 'Q2147483647' ),
 			new PropertyId( 'P1' ),
 			new PropertyId( 'P42' ),
 			new PropertyId( 'P31337' ),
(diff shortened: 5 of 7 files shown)


src/Entity/Int32EntityId.php
```diff
@@ -0,0 +1,50 @@
...
+
+	/**
+	 * @since 6.1
+	 *
+	 * @param string $idSerialization
+	 *
+	 * @throws InvalidArgumentException When the provided serialization does not allow getNumericId
+	 *  to return a unique positive 32 bit integer.
+	 */
+	public function __construct( $idSerialization );
```
brightbyte: Seeing a constructor in an interface is odd to me - it's essentially a static function, right? And when we call a constructor, we should always know the concrete class... 
Specifying this here means that no class that implements this can have any additional constructor arguments, either. What do we need that guarantee/restricting for?

src/Entity/Int32EntityId.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+namespace Wikibase\DataModel\Entity;
+
+use InvalidArgumentException;
+
+/**
+ * Interface for EntityIds that can be converted to a unique positive 32 bit integer in the range
+ * [1..2147483647], and back from the entity type and the number. Entity types that do not meet this
```
brightbyte: If the range is [1..2147483647], the interface should be called Int31EntityId, right?
MySQL supports the full 32 bit unsigned integer range. If we want to support 32bit PHP, we'll need to limit ourselves to 31 bits, though. Postgres also doesn't seem to have an unsigned 4 byte int type, so 31 bit is safer.

src/Entity/Int32EntityId.php
```diff
@@ -0,0 +1,50 @@
...
+	 *
+	 * @throws InvalidArgumentException When the provided serialization does not allow getNumericId
+	 *  to return a unique positive 32 bit integer.
+	 */
+	public function __construct( $idSerialization );
+
+	/**
+	 * @since 6.1
+	 *
+	 * @return int Guaranteed to be a unique positive 32 bit integer.
```
brightbyte: 31 bit , according to the class level docs

src/Entity/Int32EntityId.php
```diff
@@ -0,0 +1,50 @@
...
+	 */
+	public function getNumericId();
+
+	/**
+	 * @since 6.1
+	 *
+	 * @param int|float|string $numericId
```
brightbyte: if a strign is passed in, how is it parsed? Most importantly, do leading zeros indicate octal notation?

src/Entity/ItemId.php
```diff
@@ -29,12 +29,16 @@ public function __construct( $idSerialization ) {
...
 		if ( !is_string( $idSerialization ) ) {
-			throw new InvalidArgumentException( '$idSerialization must be a string; got ' . gettype( $idSerialization ) );
+			throw new InvalidArgumentException( '$idSerialization must be a string' );
 		}
 
 		if ( !preg_match( self::PATTERN, $idSerialization ) ) {
 			throw new InvalidArgumentException( '$idSerialization must match ' . self::PATTERN );
 		}
+
+		if ( strlen( $idSerialization ) > 10 && substr( $idSerialization, 1 ) > 2147483647 ) {
```
brightbyte: comparing a substring to an integer is a bit scary, the implicit conversion isn't obvious

src/Entity/ItemId.php
```diff
@@ -29,12 +29,16 @@ public function __construct( $idSerialization ) {
...
+			throw new InvalidArgumentException( '$idSerialization can not exceed 2147483647' );
```
brightbyte: Can we have a constant for the maximum id, instead of repeating the literal everywhere?

brightbyte:
There's a bit of confusion about 32 bit  signed and unsigned here. For compatibility with all reasonable platform, we need to restrict the range to 31 bit unsigned (which fits into a 32 bit signed int, but isn't the same thing). The class level docs of the proposed interface are compatible with this, but it's a bit ambiguous in other places.

src/Entity/Int32EntityId.php
thiemowmde: We need the `@throws`, otherwise this whole interface is bogus and unusable, as we discussed.

src/Entity/Int32EntityId.php
thiemowmde: There is no unsigned integer in PHP, not in the 32 bit version and not in the 64 bit version. Calling this "31 bit" when the whole world calls this concept 32 bit is doing nothing but adding confusion.

src/Entity/Int32EntityId.php
thiemowmde: No.

src/Entity/Int32EntityId.php
thiemowmde: What?

src/Entity/Int32EntityId.php
brightbyte: ok, then call it "a 32 bit integer in the range [1..2147483647]".

src/Entity/ItemId.php
thiemowmde: I played around with multiple ways of writing this line, and everything was ugly and not easier to read than this one. The only thing that may help is adding `.0` at the end of the constant. But this doesn't really do anything. What's critical here is the test, not the implementation.

src/Entity/ItemId.php
thiemowmde: "Everywhere"? I see it a single time in the code.

src/Entity/Int32EntityId.php
brightbyte: The integer type has 32 bits. This interface only allows us to use 31. I don't care too much about what the interface is called, but I do care that saying "this is a 32 bit int" is *not* sufficient.

src/Entity/Int32EntityId.php
brightbyte: We need the getter to satisfy the contract declared below. That would usually be done with throwing an exception from the constructor. But there are other ways to do this (e.g. a static factory method, and a private constructor). 

Yes, we need the throws, but declaring it in the interface forces us to lock down the signature. That's problematic, and I don't see an advantage, since we should always know the concrete class when calling the constructor.

thiemowmde:
> compatibility with all reasonable platform

What other platform would that be? This is PHP, not a cross-compiling interface. PHP comes in two versions, 32 bit and 64 bit, and they are both literally called like this, because that's what they are.

I'm afraid the sheer amount, length and general off-topic-ness of the comments is making me upset and helpless again. I don't know what else I should do other than closing all my proposals and making room for others to try to come up with a clean solution to all this legacy mess that doesn't make the whole situation worse. Who cares if this takes another year?

Oh, wait, I do care. :-(

src/Entity/Int32EntityId.php
thiemowmde: That's the same as saying "positive 32 bit integer".

src/Entity/Int32EntityId.php
thiemowmde: No, that would not be sufficient, and that's why it does not say that.

src/Entity/Int32EntityId.php
brightbyte: What should newFromNumber( '020' ) return? "X020" or "X20" or "X16"?
That's 'X' . $numericId  vs.  'X' . (int)$numericId  vs.  'X' . intval($numericId, 0).  
I checked again now, and it seems the case of X16 isn't as relevant as I thought, since intval doesn't do "automatic octal" per default. But the spec should protect against "X020".

src/Entity/Int32EntityId.php
brightbyte: No. 32 bit unsigned int (in C or MySQL) has the range [0..4294967295]. That's *not* what we want here.

src/Entity/ItemId.php
brightbyte: I see it twice as a literal, twice in a string, once in a comment (should probably be more), and various times in tests.

src/Entity/Int32EntityId.php
thiemowmde: This must be in the interface. What you are proposing creates again the main issue we already discussed: implementations are not allowed to construct EntityId objects that don't fit into the contract of this interface.

Throw an exception. Early. Always. Not just when that getter is called.

src/Entity/Int32EntityId.php
brightbyte: Up here it gives the range, but it's unclear in the declaration if getNumericId.

src/Entity/Int32EntityId.php
thiemowmde: What do you want to prove here? This is all irrelevant and/or covered since forever. There is nothing to discuss here.

src/Entity/Int32EntityId.php
thiemowmde: Why are you talking about "unsigned" again? This is PHP, not C or MySQL.

src/Entity/Int32EntityId.php
brightbyte: I agree that an exception should be thrown early. I just think locking down constructor signature is a very high price to pay just for a place to put the @throws tag. We don't even get any kind of enforcement for that price.

src/Entity/Int32EntityId.php
brightbyte: In order to use the return value from this method, I need to know that it's => 0, and <= 2147483647. Only then can I use it with our current DB schema with both MySQL and PostGres. That guarantee is the reason we have this interface. So that guarantee should be mentioned here explicitly.

src/Entity/Int32EntityId.php
brightbyte: I want a contract that actually *says* how leading zeros are treated. How is it covered for forever? As far as I can see, it's currently unspecified.

src/Entity/Int32EntityId.php
thiemowmde: That enforced constructor behavior should be in EntityId, since IDs are guaranteed to be strings anyway. They must all always support setting this string somehow. When we turn this around and move the constructor taking a string serialization to a static method, and allow `__constructor` to be anything, we don't win anything.

EntityIds that can be constructed from other stuff should use static methods for that. As this interface actually does.

src/Entity/Int32EntityId.php
thiemowmde: I'm repeating myself: This guarantee is mentioned.

src/Entity/Int32EntityId.php
thiemowmde: There are no leading zeros. Add a test if you think it's missing, but I'm quite sure it's already there.

The interface demands "unique" numbers. If numbers with leading zeros are allowed, it's not unique any more. So they are not.

src/Entity/Int32EntityId.php
brightbyte: Ok, let me rephrase: please mention the range *here* explicitly. This is the place where it matters most. Not everyone reads the class level docs when checking for the guarantees of a return value.

src/Entity/Int32EntityId.php
brightbyte: The existing implementations don't allow leading zeros. I would like this interface to guarantee that they will never be allowed (or will always be stripped).

thiemowmde: It does guarantee that, as I explained.

brightbyte: From reading "allow getNumericId to return a unique integer in the range...", it's not at all clear to me what the input "010" would to. Would that raise an error? Or would it ignore the leading "0"? The integer that is returned is unique in both cases (actually, all integers are unique). 
We could try to say that all different inputs to this method must result in a different integer being returned from getNumericId. But that's not strictly true: 0x10 and '16' and 16.0 all yield the same value for getNumericId. Why shouldn't '016' work, too? It seems natural that it would. 
The documentation seems unclear about this point: should '016' work, or trigger an exception? I think it would be good to clarify it.

thiemowmde: "can be converted to a unique integer […] **and back** from the entity type and the number."

thiemowmde:
* Removed `newFromNumber` from the interface, because MediaInfoId does not need and should not be forced to implement this.
* Removed constructor. Not as critical as newFromNumber, since all constructors are the same anyway, but having a constructor in this interface also does not help much. The only thing it should enforce (throwing an exception) can not really be enforced by this interface. It's a comment now.

PR #671, opened 2016-05-23
Serialize EntityIdValue with "id" without newFromArray support

thiemowmde:
Alternative for #669, with a very important difference: With this patch EntityIdvalue::newFromArray becomes a legacy thing that could be deprecated or even removed. It ~~can~~ should not be used for deserialization any more. It can not be used any more the moment we drop the two legacy keys "entity-type" and "numeric-id". Most importantly: It will never support new entity types. This will be done by https://github.com/DataValues/Serialization/pull/14.

The assumption that getArrayValue and newFromArray roundtrip is not true any more with this patch. I'm not sure if this makes this a breaking change.

I wanted separate pull requests so you can see the difference.

[Bug: T132592](https://phabricator.wikimedia.org/T132592)

### DIFF
--- a/src/Entity/EntityIdValue.php
+++ b/src/Entity/EntityIdValue.php
@@ -12,6 +12,7 @@ use Wikibase\DataModel\LegacyIdInterpreter;
  *
  * @license GPL-2.0+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
+ * @author Thiemo Mättig
  */
 class EntityIdValue extends DataValueObject {
 
@@ -43,7 +44,7 @@ class EntityIdValue extends DataValueObject {
 	 *
 	 * @return float Numeric id as a whole number. Can not be int because of 32-bit PHP.
 	 */
-	protected function getNumericId() {
+	private function getNumericId() {
 		return floatval( substr( $this->entityId->getSerialization(), 1 ) );
 	}
 
@@ -121,6 +122,7 @@ class EntityIdValue extends DataValueObject {
 		return array(
 			'entity-type' => $this->entityId->getEntityType(),
 			'numeric-id' => $this->getNumericId(),
+			'id' => $this->entityId->getSerialization(),
 		);
 	}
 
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
I'd be more happy if we retain the roundtrip behavior. This is easy to do no? Having an exception mentioning one should use a class for a component unknown to this one is somewhat odd (though not wrong).

thiemowmde:
"Easy" how? Please be more specific. The only way I could think of can be seen in #669. How else can a static method in EntityIdValue know how to deserialize arbitrary entity ID serializations?

One thing we could to is to simply remove EntityIdValue::newFromArray. There is neither an interface nor an base test that requires this method and the roundtrip.

JeroenDeDauw:
I was talking about retaining the roundtrip behaviour for items and properties

thiemowmde:
That did not changed. Legacy serializations of items and properties are still supported via LegacyIdInterpreter. See `validArrayProvider`.

JeroenDeDauw:
What about the non-legacy serializations? Those do break the roundtrip of item and property ids right?

thiemowmde:
Note that entity types other than items and properties always broke the roundtrip. `getArrayValue` always worked for all entity types, `newFromArray` did not.

As long as serializations contain the legacy keys (which they do, we will not remove them for at least half a year) nothing changes. Plus: we must keep `newFromArray` for forever because the wikidata.org database will always contain some of these legacy serializations.

Hm. I just realized that we wanted this to be configurable: What's stored in the database should not contain the legacy keys, only `"id"`. Exports (dumps, API, Special:EntityData) should contain all three keys. This is missing in this patch. This will blow up the database a bit, which we wanted to avoid. Suggestions?

src/Entity/EntityIdValue.php
```diff
@@ -121,6 +122,7 @@ public function getArrayValue() {
 		return array(
 			'entity-type' => $this->entityId->getEntityType(),
 			'numeric-id' => $this->getNumericId(),
+			'id' => $this->entityId->getSerialization(),
```
brightbyte: This is the only "functional" change in this patch, right?

brightbyte:
I would leave the "sparse" mode with only the id field set for later. It could be implemented by an optional flag parameter in getArrayValue().

src/Entity/EntityIdValue.php
thiemowmde: Correct.

PR #672, opened 2016-09-05
Remove "@group …Test" tags for individual tests

thiemowmde:
Multiple arguments to remove these:
* It's easy to run a single …Test.php file or a single directory of tests. These `@group` tags are not needed to do this.
* What these tags do is, in essence, repeating either the class or the directory name. We tend to forget this when renaming or moving files. A good amount is already outdated.
* Some of these `@group` tags are the result of copy pasting and do not belong to the file they are in.


JeroenDeDauw:
+1

I'd just remove all `@group` :) Most new stuff does not have them if I'm not mistaken

PR #674, opened 2016-09-05
Document usages of the Int32EntityId interface

thiemowmde:


src/Entity/ItemId.php
```diff
@@ -45,7 +45,9 @@ private function assertValidIdFormat( $idSerialization ) {
 	}
 
 	/**
-	 * @return int
+	 * @see Int32EntityId::getNumericId
+	 *
+	 * @return int Guaranteed to be a unique integer in the range [1..2147483647].
```
JeroenDeDauw: Oh sh*t merged too fast...

Don't think the "unique" part is correct. Or at least needs some very specific interpretation that is not specified. You can clearly have multiple instances with the same value. The uniqueness is only there in some context. (Which is why this class does not say anything about this in the first place)

JeroenDeDauw: Sorry for the chaos https://github.com/wmde/WikibaseDataModel/pull/677

thiemowmde: The `@see Int32EntityId` is critical here. I did not wanted to repeat the entirety of the interface. "Unique" means that non-equal EntityId objects are guaranteed to return different integer numbers. EntityId objects that are equal obviously return the same numeric ID.

This is not a GUID. Nothing is said about "global" or something.

PR #675, opened 2016-09-05
Add type hints to EntityIdValue

thiemowmde:
`newFromArray` is not in an interfaces, but assumed to be there when called from `DataValueDeserializer`. No other assumption is made about the parameter. Having the strict type hint is not different from the manual check-and-throw.


JeroenDeDauw:
> Having the strict type hint is not different from the manual check-and-throw.

Yes it is; the former results in a PHP fatal error, the later results into something that can (and should be) caught by calling code.

thiemowmde:
Replaced with #750.

PR #677, opened 2016-09-05
Document usages of the Int32EntityId interface

JeroenDeDauw:
Resubmission of https://github.com/wmde/WikibaseDataModel/pull/674 by @thiemowmde

Please regard this as if it where the original PR. I accidentally merged it, and then reverted plus resubmitted to undo my mistake. This should have no bearing on further review.

### DIFF
--- a/src/Entity/ItemId.php
+++ b/src/Entity/ItemId.php
@@ -49,9 +49,10 @@ class ItemId extends EntityId implements Int32EntityId {
 	}
 
 	/**
-	 * @return int
+	 * @see Int32EntityId::getNumericId
 	 *
 	 * @throws RuntimeException if called on a foreign ID.
+	 * @return int Guaranteed to be a distinct integer in the range [1..2147483647].
 	 */
 	public function getNumericId() {
 		if ( $this->isForeign() ) {
--- a/src/Entity/Int32EntityId.php
+++ b/src/Entity/Int32EntityId.php
@@ -3,11 +3,11 @@
 namespace Wikibase\DataModel\Entity;
 
 /**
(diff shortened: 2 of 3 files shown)


src/Entity/ItemId.php
```diff
@@ -45,7 +45,9 @@ private function assertValidIdFormat( $idSerialization ) {
 	}
 
 	/**
-	 * @return int
+	 * @see Int32EntityId::getNumericId
+	 *
+	 * @return int Guaranteed to be a unique integer in the range [1..2147483647].
```
JeroenDeDauw: Don't think the "unique" part is correct. Or at least needs some very specific interpretation that is not specified. You can clearly have multiple instances with the same value. The uniqueness is only there in some context. (Which is why this class does not say anything about this in the first place)

thiemowmde: See https://github.com/wmde/WikibaseDataModel/pull/674#discussion_r77503685.

brightbyte:
what about this?

thiemowmde:
I don't know. Help me understand what I should do to get this merged? Or just merge it?

brightbyte:
Seems to need a rebase now.

thiemowmde:
Rebased and updated to use "distinct" instead of "unique".

JeroenDeDauw:
I think `distinct` has the same problem as `unique`.

thiemowmde:
In https://github.com/wmde/WikibaseDataModel/pull/677#r77503892 I already explained the purpose and relevance of this line of documentation extensively. The method guarantees to return a different number for every non-equal object. Without this guarantee being documented the method is of not much value. Please provide a way forward.

PR #678, opened 2016-09-19
Support foreign EntityIds.

jakobw:
Task on Phabricator: https://phabricator.wikimedia.org/T145516

### DIFF
--- a/src/Entity/EntityId.php
+++ b/src/Entity/EntityId.php
@@ -3,6 +3,7 @@
 namespace Wikibase\DataModel\Entity;
 
 use Comparable;
+use InvalidArgumentException;
 use Serializable;
 
 /**
@@ -17,6 +18,32 @@ abstract class EntityId implements Comparable, Serializable {
 
 	protected $serialization;
 
+	const PATTERN = '/^:?(\w+:)*[^:]+\z/';
+
+	/**
+	 * @since 6.2
+	 *
+	 * @param string $serialization
+	 */
+	public function __construct( $serialization ) {
+		self::assertValidSerialization( $serialization );
+		$this->serialization = self::normalizeIdSerialization( $serialization );
+	}
+
+	private static function assertValidSerialization( $serialization ) {
+		if ( !is_string( $serialization ) ) {
+			throw new InvalidArgumentException( '$serialization must be a string' );
+		}
+
+		if ( $serialization === '' ) {
+			throw new InvalidArgumentException( '$serialization must not be an empty string' );
+		}
+
+		if ( !preg_match( self::PATTERN, $serialization ) ) {
+			throw new InvalidArgumentException( '$serialization must match ' . self::PATTERN );
+		}
+	}
+
 	/**
 	 * @return string
 	 */
@@ -27,6 +54,102 @@ abstract class EntityId implements Comparable, Serializable {
 	 */
 	public function getSerialization() {
 		return $this->serialization;
+	}
+
+	/**
+	 * Returns an array with 3 elements: the foreign repository name as the first element, the local ID as the last
+	 * element and everything that is in between as the second element.
+	 *
+	 * EntityId::joinSerialization can be used to restore the original serialization from the parts returned.
+	 *
+	 * @since 6.2
+	 *
+	 * @param string $serialization
+	 * @return string[] Array containing the serialization split into 3 parts.
+	 */
+	public static function splitSerialization( $serialization ) {
+		self::assertValidSerialization( $serialization );
+
+		$parts = explode( ':', self::normalizeIdSerialization( $serialization ) );
+		$localPart = array_pop( $parts );
+		$repoName = array_shift( $parts );
+		$prefixRemainder = implode( ':', $parts );
+
+		return array(
+			is_string( $repoName ) ? $repoName : '',
+			$prefixRemainder,
+			$localPart
+		);
+	}
+
+	/**
+	 * Builds an ID serialization from the parts returned by EntityId::splitSerialization.
+	 *
+	 * @since 6.2
+	 *
+	 * @param string[] $parts
+	 * @return string
+	 *
+	 * @throws InvalidArgumentException
+	 */
+	public static function joinSerialization( array $parts ) {
+		if ( end( $parts ) === '' ) {
+			throw new InvalidArgumentException( 'The last element of $parts must not be empty.' );
+		}
+
+		return implode(
+			':',
+			array_filter( $parts, function( $part ) {
+				return $part !== '';
+			} )
+		);
+	}
+
+	/**
+	 * Returns '' for local IDs and the foreign repository name for foreign IDs. For chained IDs (e.g. foo:bar:Q42) it
+	 * will return only the first part.
+	 *
+	 * @since 6.2
+	 *
+	 * @return string
+	 */
+	public function getRepositoryName() {
+		$parts = self::splitSerialization( $this->serialization );
+
+		return $parts[0];
+	}
+
+	/**
+	 * Returns the serialization without the first repository prefix.
+	 *
+	 * @since 6.2
+	 *
+	 * @return string
+	 */
+	public function getLocalPart() {
+		$parts = self::splitSerialization( $this->serialization );
+
+		return self::joinSerialization( array( '', $parts[1], $parts[2] ) );
+	}
+
+	/**
+	 * Returns true iff EntityId::getRepoName returns a non-empty string.
+	 *
+	 * @since 6.2
+	 *
+	 * @return bool
+	 */
+	public function isForeign() {
+		// not actually using EntityId::getRepoName for performance reasons
+		return strpos( $this->serialization, ':' ) > 0;
+	}
+
+	/**
+	 * @param string $id
+	 * @return string
+	 */
+	private static function normalizeIdSerialization( $id ) {
+		return ltrim( $id, ':' );
 	}
 
 	/**
--- a/src/Entity/ItemId.php
+++ b/src/Entity/ItemId.php
@@ -3,6 +3,7 @@
 namespace Wikibase\DataModel\Entity;
 
 use InvalidArgumentException;
+use RuntimeException;
 
 /**
  * @since 0.5
@@ -23,8 +24,12 @@ class ItemId extends EntityId implements Int32EntityId {
 	 * @throws InvalidArgumentException
 	 */
 	public function __construct( $idSerialization ) {
-		$this->assertValidIdFormat( $idSerialization );
-		$this->serialization = strtoupper( $idSerialization );
+		$serializationParts = self::splitSerialization( $idSerialization );
+		$localId = strtoupper( $serializationParts[2] );
+		$this->assertValidIdFormat( $localId );
+		parent::__construct( self::joinSerialization(
+			array( $serializationParts[0], $serializationParts[1], $localId ) )
+		);
 	}
 
 	private function assertValidIdFormat( $idSerialization ) {
@@ -46,8 +51,14 @@ class ItemId extends EntityId implements Int32EntityId {
 
 	/**
 	 * @return int
+	 *
+	 * @throws RuntimeException if called on a foreign ID.
 	 */
 	public function getNumericId() {
+		if ( $this->isForeign() ) {
+			throw new RuntimeException( 'getNumericId must not be called on foreign ItemIds' );
+		}
+
 		return (int)substr( $this->serialization, 1 );
 	}
 
--- a/src/Entity/PropertyId.php
+++ b/src/Entity/PropertyId.php
@@ -3,6 +3,7 @@
 namespace Wikibase\DataModel\Entity;
 
 use InvalidArgumentException;
+use RuntimeException;
 
 /**
  * @since 0.5
@@ -23,8 +24,12 @@ class PropertyId extends EntityId implements Int32EntityId {
 	 * @throws InvalidArgumentException
 	 */
 	public function __construct( $idSerialization ) {
-		$this->assertValidIdFormat( $idSerialization );
-		$this->serialization = strtoupper( $idSerialization );
+		$serializationParts = self::splitSerialization( $idSerialization );
+		$localId = strtoupper( $serializationParts[2] );
+		$this->assertValidIdFormat( $localId );
+		parent::__construct( self::joinSerialization(
+			array( $serializationParts[0], $serializationParts[1], $localId ) )
+		);
 	}
 
 	private function assertValidIdFormat( $idSerialization ) {
@@ -46,8 +51,14 @@ class PropertyId extends EntityId implements Int32EntityId {
 
 	/**
 	 * @return int
+	 *
+	 * @throws RuntimeException if called on a foreign ID.
 	 */
 	public function getNumericId() {
+		if ( $this->isForeign() ) {
+			throw new RuntimeException( 'getNumericId must not be called on foreign PropertyIds' );
+		}
+
 		return (int)substr( $this->serialization, 1 );
 	}
 
--- a/tests/unit/Entity/EntityIdTest.php
+++ b/tests/unit/Entity/EntityIdTest.php
@@ -2,6 +2,9 @@
 
 namespace Wikibase\DataModel\Tests\Entity;
 
+use ReflectionClass;
+use InvalidArgumentException;
+use RuntimeException;
 use Wikibase\DataModel\Entity\EntityId;
 use Wikibase\DataModel\Entity\ItemId;
 use Wikibase\DataModel\Entity\PropertyId;
@@ -24,11 +27,14 @@ class EntityIdTest extends \PHPUnit_Framework_TestCase {
 	public function instanceProvider() {
 		$ids = array();
 
-		$ids[] = array( new ItemId( 'Q1' ) );
-		$ids[] = array( new ItemId( 'Q42' ) );
-		$ids[] = array( new ItemId( 'Q31337' ) );
-		$ids[] = array( new ItemId( 'Q2147483647' ) );
-		$ids[] = array( new PropertyId( 'P101010' ) );
+		$ids[] = array( new ItemId( 'Q1' ), '' );
+		$ids[] = array( new ItemId( 'Q42' ), '' );
+		$ids[] = array( new ItemId( 'Q31337' ), '' );
+		$ids[] = array( new ItemId( 'Q2147483647' ), '' );
+		$ids[] = array( new ItemId( ':Q2147483647' ), '' );
+		$ids[] = array( new ItemId( 'foo:Q2147483647' ), 'foo' );
+		$ids[] = array( new PropertyId( 'P101010' ), '' );
+		$ids[] = array( new PropertyId( 'foo:bar:P101010' ), 'foo' );
 
 		return $ids;
 	}
@@ -79,6 +85,115 @@ class EntityIdTest extends \PHPUnit_Framework_TestCase {
 	 */
 	public function testReturnTypeOfToString( EntityId $id ) {
 		$this->assertInternalType( 'string', $id->__toString() );
+	}
+
+	public function testIsForeign() {
+		$this->assertFalse( ( new ItemId( 'Q42' ) )->isForeign() );
+		$this->assertFalse( ( new ItemId( ':Q42' ) )->isForeign() );
+		$this->assertTrue( ( new ItemId( 'foo:Q42' ) )->isForeign() );
+		$this->assertFalse( ( new PropertyId( ':P42' ) )->isForeign() );
+		$this->assertTrue( ( new PropertyId( 'foo:P42' ) )->isForeign() );
+	}
+
+	/**
+	 * @dataProvider instanceProvider
+	 */
+	public function testGetRepositoryName( EntityId $id, $repoName ) {
+		$this->assertSame( $repoName, $id->getRepositoryName() );
+	}
+
+	public function serializationSplitProvider() {
+		return array(
+			array( 'Q42', array( '', '', 'Q42' ) ),
+			array( 'foo:Q42', array( 'foo', '', 'Q42' ) ),
+			array( '0:Q42', array( '0', '', 'Q42' ) ),
+			array( 'foo:bar:baz:Q42', array( 'foo', 'bar:baz', 'Q42' ) ),
+		);
+	}
+
+	/**
+	 * @dataProvider serializationSplitProvider
+	 */
+	public function testSplitSerialization( $serialization, $split ) {
+		$this->assertSame( $split, EntityId::splitSerialization( $serialization ) );
+	}
+
+	/**
+	 * @dataProvider invalidSerializationProvider
+	 */
+	public function testSplitSerializationFails_GivenInvalidSerialization( $serialization ) {
+		$this->setExpectedException( InvalidArgumentException::class );
+		EntityId::splitSerialization( $serialization );
+	}
+
+	/**
+	 * @dataProvider serializationSplitProvider
+	 */
+	public function testJoinSerialization( $serialization, $split ) {
+		$this->assertSame( $serialization, EntityId::joinSerialization( $split ) );
+	}
+
+	/**
+	 * @dataProvider invalidJoinSerializationDataProvider
+	 */
+	public function testJoinSerializationFails_GivenEmptyId( $parts ) {
+		$this->setExpectedException( InvalidArgumentException::class );
+		EntityId::joinSerialization( $parts );
+	}
+
+	public function invalidJoinSerializationDataProvider() {
+		return array(
+			array( array( 'Q42', '', '' ) ),
+			array( array( '', 'Q42', '' ) ),
+			array( array( 'foo', 'Q42', '' ) ),
+		);
+	}
+
+	public function testGivenNotNormalizedSerialization_splitSerializationReturnsNormalizedParts() {
+		$this->assertSame( array( '', '', 'Q42' ), EntityId::splitSerialization( ':Q42' ) );
+		$this->assertSame( array( 'foo', 'bar', 'Q42' ), EntityId::splitSerialization( ':foo:bar:Q42' ) );
+	}
+
+	public function localPartDataProvider() {
+		return array(
+			array( 'Q42', 'Q42' ),
+			array( ':Q42', 'Q42' ),
+			array( 'foo:Q42', 'Q42' ),
+			array( 'foo:bar:Q42', 'bar:Q42' ),
+		);
+	}
+
+	/**
+	 * @dataProvider localPartDataProvider
+	 */
+	public function testGetLocalPart( $serialization, $localPart ) {
+		$id = new ItemId( $serialization );
+		$this->assertSame( $localPart, $id->getLocalPart() );
+	}
+
+	public function invalidSerializationProvider() {
+		return array(
+			array( 's p a c e s:Q42' ),
+			array( '::Q42' ),
+			array( '' ),
+			array( ':' ),
+			array( 42 ),
+			array( null ),
+		);
+	}
+
+	/**
+	 * @dataProvider invalidSerializationProvider
+	 */
+	public function testConstructor( $serialization ) {
+		$this->setExpectedException( InvalidArgumentException::class );
+
+		$mock = $this->getMockBuilder( EntityId::class )
+			->disableOriginalConstructor()
+			->getMockForAbstractClass();
+
+		$constructor = ( new ReflectionClass( EntityId::class ) )->getConstructor();
+		$constructor->invoke( $mock, $serialization );
 	}
 
 }
--- a/tests/unit/Entity/ItemIdTest.php
+++ b/tests/unit/Entity/ItemIdTest.php
@@ -4,6 +4,8 @@ namespace Wikibase\DataModel\Tests\Entity;
 
 use PHPUnit_Framework_TestCase;
 use Wikibase\DataModel\Entity\ItemId;
+use InvalidArgumentException;
+use RuntimeException;
 
 /**
  * @covers Wikibase\DataModel\Entity\ItemId
@@ -21,24 +23,27 @@ class ItemIdTest extends PHPUnit_Framework_TestCase {
 	/**
 	 * @dataProvider idSerializationProvider
 	 */
-	public function testCanConstructId( $idSerialization ) {
+	public function testCanConstructId( $idSerialization, $normalizedIdSerialization ) {
 		$id = new ItemId( $idSerialization );
 
 		$this->assertEquals(
-			strtoupper( $idSerialization ),
+			$normalizedIdSerialization,
 			$id->getSerialization()
 		);
 	}
 
 	public function idSerializationProvider() {
 		return array(
-			array( 'q1' ),
-			array( 'q100' ),
-			array( 'q1337' ),
-			array( 'q31337' ),
-			array( 'Q31337' ),
-			array( 'Q42' ),
-			array( 'Q2147483647' ),
+			array( 'q1', 'Q1' ),
+			array( 'q100', 'Q100' ),
+			array( 'q1337', 'Q1337' ),
+			array( 'q31337', 'Q31337' ),
+			array( 'Q31337', 'Q31337' ),
+			array( 'Q42', 'Q42' ),
+			array( ':Q42', 'Q42' ),
+			array( 'foo:Q42', 'foo:Q42' ),
+			array( 'foo:bar:q42', 'foo:bar:Q42' ),
+			array( 'Q2147483647', 'Q2147483647' ),
 		);
 	}
 
@@ -46,7 +51,7 @@ class ItemIdTest extends PHPUnit_Framework_TestCase {
 	 * @dataProvider invalidIdSerializationProvider
 	 */
 	public function testCannotConstructWithInvalidSerialization( $invalidSerialization ) {
-		$this->setExpectedException( 'InvalidArgumentException' );
+		$this->setExpectedException( InvalidArgumentException::class );
 		new ItemId( $invalidSerialization );
 	}
 
@@ -133,7 +138,7 @@ class ItemIdTest extends PHPUnit_Framework_TestCase {
 	 * @dataProvider invalidNumericIdProvider
 	 */
 	public function testNewFromNumberWithInvalidNumericId( $number ) {
-		$this->setExpectedException( 'InvalidArgumentException' );
+		$this->setExpectedException( InvalidArgumentException::class );
 		ItemId::newFromNumber( $number );
 	}
 
@@ -145,6 +150,11 @@ class ItemIdTest extends PHPUnit_Framework_TestCase {
 			array( 2147483648 ),
 			array( '2147483648' ),
 		);
+	}
+
+	public function testGetNumericIdThrowsExceptionOnForeignIds() {
+		$this->setExpectedException( RuntimeException::class );
+		( new ItemId( 'foo:Q42' ) )->getNumericId();
 	}
 
 }
--- a/tests/unit/Entity/PropertyIdTest.php
+++ b/tests/unit/Entity/PropertyIdTest.php
@@ -4,6 +4,8 @@ namespace Wikibase\DataModel\Tests\Entity;
 
 use PHPUnit_Framework_TestCase;
 use Wikibase\DataModel\Entity\PropertyId;
+use InvalidArgumentException;
+use RuntimeException;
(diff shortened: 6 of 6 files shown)


src/Entity/PropertyId.php
```diff
@@ -15,7 +15,7 @@ class PropertyId extends EntityId implements Int32EntityId {
 	/**
 	 * @since 0.5
 	 */
-	const PATTERN = '/^P[1-9]\d{0,9}\z/i';
+	const PATTERN = '/^(:?|(\w+:)+)P[1-9]\d{0,9}\z/i';
```
JeroenDeDauw: The exception case in the constructor is not tested

src/Entity/EntityId.php
```diff
@@ -30,6 +30,37 @@ public function getSerialization() {
 	}
 
 	/**
+	 * Returns the foreign repository name or empty string for local entities.
+	 * @return string
+	 */
+	public function getRepoName() {
```
JeroenDeDauw: non-private should have `@since` tag

src/Entity/EntityId.php
```diff
@@ -30,6 +30,37 @@ public function getSerialization() {
...
+
+	/**
+	 * @return bool
+	 */
+	public function isForeign() {
+		return strpos( $this->serialization, ':' ) > 0;
+	}
+
+	protected function sanitizeIdSerialization( $id ) {
+		return $this->upcaseLocalId( ltrim( $id, ':' ) );
```
JeroenDeDauw: The ltrim here is not tested as far as I can tell

src/Entity/EntityId.php
```diff
@@ -30,6 +30,37 @@ public function getSerialization() {
...
+
+	}
+
+	private function upcaseLocalId( $id ) {
+		$parts = explode( ':', $id );
+		$parts[count( $parts ) - 1] = strtoupper( end( $parts ) );
```
JeroenDeDauw: Was it decided to not have (local) IDs with colons in them? I'm not sure if such a thing would be needed for commons or some such.

If there is no such decision, then how about not doing this little code reuse via inheritance thing and simply putting the few lines in both ItemId and PropertyId? Same goes for the test. As far as I remember the reason for why we have this code and the test like this is that we started out with just having EntityId and only later introduced ItemId and PropertyId, ie for legacy reasons and not by design.

src/Entity/EntityId.php
```diff
@@ -30,6 +37,68 @@ public function getSerialization() {
...
+	public function getRepoName() {
+		$parts = self::splitSerialization( $this->serialization );
+
+		return $parts[0];
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getLocalPart() {
```
manicki: Just so it is clear for me: Given serialized ID `foo:bar:Q1337` is "local part" of it `Q1337` or `bar:Q1337`.
I thought the former but I am not sure if we're done with the naming discussion here. If the latter is the local part  what is `Q1337` then?
Sorry if I am just creating confusion.

src/Entity/EntityId.php
```diff
@@ -18,6 +18,13 @@
 	protected $serialization;
 
 	/**
+	 * @param $serialization
```
manicki: this is always a string, right? then you could have `@param string $serialization` here

src/Entity/EntityId.php
```diff
@@ -30,6 +37,68 @@ public function getSerialization() {
 	}
 
 	/**
+	 * Returns an array with 3 elements: the foreign repository name as the first element, the local ID as the last
+	 * element and everything that is in between as the second element.
+	 *
+	 * @param $serialization
```
manicki: string?

src/Entity/EntityId.php
```diff
@@ -30,6 +37,68 @@ public function getSerialization() {
...
+	 * @return array
```
manicki: it seems PHP's explode always returns an array of strings, so this method also does so. This might also be `@return string[]` then

src/Entity/EntityId.php
```diff
@@ -30,6 +37,68 @@ public function getSerialization() {
...
+		return array( $repoName, $prefixRemainder, $localPart );
+	}
+
+	/**
+	 * Concatenates parts of an EntityId serialization with ':'.
+	 *
+	 * @param array $parts
+	 * @return string
+	 */
+	public static function joinSerialization( $parts ) {
```
manicki: There could be a type hint for `$parts`, not only phpdoc `@param` comment

src/Entity/PropertyId.php
brightbyte: Do we want to allow "-" in the prefix? Might be nice.

src/Entity/PropertyId.php
```diff
@@ -48,7 +48,7 @@ private function assertValidIdFormat( $idSerialization ) {
 	 * @return int
 	 */
 	public function getNumericId() {
-		return (int)substr( $this->serialization, 1 );
+		return (int)substr( $this->getSerialization(), 1 );
```
manicki: I believe this is wrong, as it should take the last part of the serialized ID and drop `P` from it. E.g. this will break for `foo:P123`.
Also, I believe you forgot to do changes regarding accessing `$this->serialization` you did for this class in `ItemId` class as well.

src/Entity/EntityId.php
brightbyte: Local IDs never have colons in them (we decided that), but foreign IDs can be "chained": foo:bar:Q123 is Q123 in the repo that is called bar by the repo foo. That's also how interwiki prefixes work. You can do en:fr:wiktionary:de:wikibooks:it:wikipedia:en:Foo. This will result in a series of HTTP redirects, eventually giving you the Foo page on English Wikipedia.

My personal preference for code re-use would be: protected static upcaseLocalId() called from the constructors of ItemId and ProeprtyId.

src/Entity/EntityId.php
```diff
@@ -30,6 +37,68 @@ public function getSerialization() {
...
+		$parts = explode( ':', $serialization );
+		$localPart = array_pop( $parts );
+		$repoName = array_shift( $parts ) ?: '';
+		$prefixRemainder = implode( ':', $parts );
+
+		return array( $repoName, $prefixRemainder, $localPart );
+	}
+
+	/**
+	 * Concatenates parts of an EntityId serialization with ':'.
```
brightbyte: That doesn't state the contract. The contract is "builds an ID serialization from the parts returned by splitSerialization()".

src/Entity/EntityId.php
```diff
@@ -30,6 +37,68 @@ public function getSerialization() {
 	}
 
 	/**
+	 * Returns an array with 3 elements: the foreign repository name as the first element, the local ID as the last
+	 * element and everything that is in between as the second element.
+	 *
```
brightbyte: Should mention that joinSerialization can be used to restore the original serializertion from the parts returned.

src/Entity/EntityId.php
brightbyte: should be string[] and should say that this wall *always* be 3 elements.

src/Entity/EntityId.php
```diff
@@ -30,6 +37,68 @@ public function getSerialization() {
...
+		$repoName = array_shift( $parts ) ?: '';
+		$prefixRemainder = implode( ':', $parts );
+
+		return array( $repoName, $prefixRemainder, $localPart );
+	}
+
+	/**
+	 * Concatenates parts of an EntityId serialization with ':'.
+	 *
+	 * @param array $parts
```
brightbyte: string[]. Could be required to be exactly three elements, but maybe it's useful to leave this open.

src/Entity/EntityId.php
```diff
@@ -30,6 +37,68 @@ public function getSerialization() {
...
+	 * @return string
+	 */
+	public static function joinSerialization( $parts ) {
+		return implode( ':', array_filter( $parts ) );
+	}
+
+	/**
+	 * Returns the foreign repository name or empty string for local entities.
```
brightbyte: Should say that it returns '' if it's a local ID. Should also clarify that for a "chained" foreighn ID, this returns the first part only.

src/Entity/EntityId.php
```diff
@@ -30,6 +37,68 @@ public function getSerialization() {
...
+		return $parts[0];
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getLocalPart() {
+		$parts = self::splitSerialization( $this->serialization );
+
+		return self::joinSerialization( array( '', $parts[1], $parts[2] ) );
```
brightbyte: The '' bit isn't needed.

src/Entity/EntityId.php
```diff
@@ -30,6 +37,68 @@ public function getSerialization() {
...
+	}
+
+	/**
+	 * @return bool
+	 */
+	public function isForeign() {
+		return strpos( $this->serialization, ':' ) > 0;
```
brightbyte: should perhaps explain that we don't use getRepoName() !== '' here because this is a potential performance hotspot.

src/Entity/EntityId.php
```diff
@@ -30,6 +37,68 @@ public function getSerialization() {
...
+	 * @return string
+	 */
+	public function getLocalPart() {
+		$parts = self::splitSerialization( $this->serialization );
+
+		return self::joinSerialization( array( '', $parts[1], $parts[2] ) );
+	}
+
+	/**
+	 * @return bool
```
brightbyte: what, when?

The contract of this is that it will return true if and only if getRepoName() returns something other than the empty string. The semantics is that it returns true if the ID refers to an entity on another repo.

I'm being picky about the documentation here, because I think well defined contracts are especially important for value objects and interfaces, especially in the data model.

src/Entity/EntityId.php
```diff
@@ -30,6 +37,68 @@ public function getSerialization() {
...
+	 */
+	public function isForeign() {
+		return strpos( $this->serialization, ':' ) > 0;
+	}
+
+	protected static function normalizeIdSerialization( $id ) {
```
brightbyte: Please add at least a minimal doc block that states the type of the parameter. Perhaps also explain when this is (or should be) called.

src/Entity/EntityId.php
```diff
@@ -30,6 +37,68 @@ public function getSerialization() {
...
+		$parts = self::splitSerialization( ltrim( $id, ':' ) );
+
+		return self::joinSerialization( array(
```
brightbyte: The trimming of the leading ":" should not be tied to the upper-casing of the local part. The leading ":" must always be stripped, that's part of the convention for foreign repo prefixes, which use the same syntax for all IDs.
Making the local part upper case is tied to specific EntityId types, and should only be called by those that actually need that.

src/Entity/EntityId.php
```diff
@@ -30,6 +37,68 @@ public function getSerialization() {
...
 
 	/**
+	 * Returns an array with 3 elements: the foreign repository name as the first element, the local ID as the last
+	 * element and everything that is in between as the second element.
+	 *
+	 * @param $serialization
+	 * @return array
+	 */
+	public static function splitSerialization( $serialization ) {
+		$parts = explode( ':', $serialization );
```
brightbyte: Should strip any leading ":" first.

src/Entity/EntityId.php
```diff
@@ -18,6 +18,13 @@
 	protected $serialization;
 
 	/**
+	 * @param $serialization
+	 */
+	protected function __construct( $serialization ) {
+		$this->serialization = $serialization;
```
brightbyte: Should fail if $serialization[0] === ':', and also if $serialization is empty or has the wrong type. Keep the checks simple though, this is a potential performance hotspot, and callers (subclass constructors) should already be doing normalization.

src/Entity/EntityId.php
```diff
@@ -30,6 +37,68 @@ public function getSerialization() {
...
+	 * @return string
+	 */
+	public function getRepoName() {
+		$parts = self::splitSerialization( $this->serialization );
+
+		return $parts[0];
+	}
+
+	/**
+	 * @return string
```
brightbyte: Explain what the "local part" is, especially in the context of chained foreign IDs.

src/Entity/EntityId.php
brightbyte: Yea... "core ID"? "actual ID"? "local local ID"?
Or the `bar:Q1337` part could be called the "foreign local ID"? Or "relative ID"?... 

Bah. The two hardest things in programming. Cache invalidation, naming things, and off-by-one errors.

src/Entity/PropertyId.php
brightbyte: Actually, getNumericId should fail for foreign IDs. Using the numeric ID of a foreign ID locally can never be right.

src/Entity/ItemId.php
```diff
@@ -24,7 +24,7 @@ class ItemId extends EntityId implements Int32EntityId {
 	 */
 	public function __construct( $idSerialization ) {
 		$this->assertValidIdFormat( $idSerialization );
-		$this->serialization = strtoupper( $idSerialization );
+		parent::__construct( parent::normalizeIdSerialization( $idSerialization ) );
```
JeroenDeDauw: What motivated using a constructor in the parent class? Definitely not worth it IMO.

tests/unit/Entity/EntityIdTest.php
```diff
@@ -81,4 +84,59 @@ public function testReturnTypeOfToString( EntityId $id ) {
...
+	 */
+	public function testSplitSerialization( $serialization, $split ) {
+		$this->assertSame( $split, EntityId::splitSerialization( $serialization ) );
+	}
+
+	/**
+	 * @dataProvider serializationSplitProvider
+	 */
+	public function testJoinSerialization( $serialization, $split ) {
+		$this->assertSame( ltrim( $serialization, ':' ), EntityId::joinSerialization( $split ) );
```
JeroenDeDauw: Having logic such as `ltrim` in tests is an anti-pattern.

src/Entity/EntityId.php
```diff
@@ -30,6 +37,68 @@ public function getSerialization() {
...
+	 * @return array
+	 */
+	public static function splitSerialization( $serialization ) {
+		$parts = explode( ':', $serialization );
+		$localPart = array_pop( $parts );
+		$repoName = array_shift( $parts ) ?: '';
+		$prefixRemainder = implode( ':', $parts );
+
+		return array( $repoName, $prefixRemainder, $localPart );
+	}
```
JeroenDeDauw: This code is procedural and uses a positional array for returning multiple values. Both are not cool

jakobw:
@brightbyte @manicki @JeroenDeDauw thanks for the review!
I updated the patch and tried to address all of the comments.

src/Entity/ItemId.php
jakobw: The idea was to do normalization in the constructor and to make `serialization` private to the parent class which isn't possible yet since it's still used in the `unserialize` methods of the subclasses.

src/Entity/EntityId.php
jakobw: I agree that it's not the prettiest code but we decided to leave it procedural for now since the alternative was less readable. Also left it positional since the order in which the parts are returned is more intuitive than the names we could come up with.

brightbyte: Positional arrays act as tuples, which seem to be a standard feature in several languages, especially functional ones. They are ideomatic in php via the list() syntax.

Also, returning a list seems fine here, since this is a method that splits something.

I agree with jakobw: We could add keys and make the result associative - if we could think of good names for the parts. "start", "middle", "end" are not very helpful...

jakobw:
Updated again to do foreign repo prefix validation in `EntityId` so that the subclasses don't have to know about it as per discussion with @manicki and @brightbyte.

src/Entity/PropertyId.php
thiemowmde: Is `:prefix1:prefix2:Q42` allowed?

src/Entity/EntityId.php
```diff
@@ -17,6 +18,30 @@
 
 	protected $serialization;
 
+	const PATTERN = '/^:?(\w+:)*[^:]+\z/';
```
thiemowmde: Wait, this regular expression **does** allow `:a:b:Q1`, but the expressions in the subclasses do not. I believe this should not be allowed, to make things as simple as possible and not have multiple string representations (`a:Q1` and `:a:Q1`) for the exact same ID.

src/Entity/EntityId.php
```diff
@@ -17,6 +18,30 @@
...
+		$this::assertValidSerialization( $serialization );
+		$this->serialization = self::normalizeIdSerialization( $serialization );
+	}
+
+	private static function assertValidSerialization( $serialization ) {
+		if ( !is_string( $serialization ) ) {
+			throw new InvalidArgumentException( '$serialization must be a string' );
+		}
+
+		if ( $serialization === '' ) {
```
thiemowmde: I find it best practice to merge the first two trivial assertions into a single "must be a non-empty string".

brightbyte: I actually started to do it the other way around, for readability...

src/Entity/EntityId.php
brightbyte: Leading colons can and should always be ignored. They are currently stripped by the constructor.

brightbyte: Why should :a:Q1 fail? It should just be normalized to "a:Q1". 
To wit, ":a:Q1" is valid as input, but should always be normalized. No EntityId would ever return a serialization starting with : from getSerialization.

src/Entity/EntityId.php
```diff
@@ -30,6 +55,91 @@ public function getSerialization() {
...
+	}
+
+	/**
+	 * Builds an ID serialization from the parts returned by EntityId::splitSerialization.
+	 *
+	 * @param string[] $parts
+	 * @return string
+	 */
+	public static function joinSerialization( array $parts ) {
+		return implode( ':', array_filter( $parts ) );
```
thiemowmde: This is not fully compatible with the `splitSerialization` implementation above. `splitSerialization` guarantees the first element is never empty. But this code here returns the exact same result for all these test cases:
* `[ 'Q1' , '', '' ]`
* `[ '' , 'Q1', '' ]`
* `[ '' , '', 'Q1' ]`

This is quite bad. Two of these cases should fail (I believe the first two).

This code also fails for `[ 'wd', '', '0' ]`, `[ '0', '0', '0' ]` and so on, because the string `"0"` is falsy in PHP. But according to the regex on top this string is a valid ID.

src/Entity/EntityId.php
```diff
@@ -30,6 +55,91 @@ public function getSerialization() {
...
+
+	/**
+	 * Returns the serialization without the first repository prefix.
+	 *
+	 * @return string
+	 */
+	public function getLocalPart() {
+		$parts = self::splitSerialization( $this->serialization );
+
+		return self::joinSerialization( array( $parts[1], $parts[2] ) );
```
thiemowmde: This shifts everything and passes the entity ID as the "remainder" part, and the remainder part as a prefix.

```php
joinSerialization( [ '', $parts[1], $parts[2] ] )
```

src/Entity/EntityId.php
```diff
@@ -30,6 +55,91 @@ public function getSerialization() {
...
+	 */
+	private static function normalizeIdSerialization( $id ) {
+		return ltrim( $id, ':' );
+	}
+
+	/**
+	 * @param string $id
+	 * @return string
+	 */
+	protected static function upcaseLocalId( $id ) {
```
thiemowmde: "upcase" is a weird naming convention. Why not "upperCase"?

src/Entity/ItemId.php
```diff
@@ -48,7 +50,11 @@ private function assertValidIdFormat( $idSerialization ) {
 	 * @return int
 	 */
 	public function getNumericId() {
-		return (int)substr( $this->serialization, 1 );
+		if ( $this->isForeign() ) {
+			throw new RuntimeException( 'getNumericId must not be called on foreign ItemIds' );
+		}
+
+		return (int)substr( $this->getSerialization(), 1 );
```
thiemowmde: Do we know this is normalized and does not start with a colon? I do miss this line of code somehow, or don't see it.

src/Entity/PropertyId.php
```diff
@@ -64,7 +70,7 @@ public function getEntityType() {
 	 * @return string
 	 */
 	public function serialize() {
-		return json_encode( array( 'property', $this->serialization ) );
+		return json_encode( array( 'property', $this->getSerialization() ) );
```
thiemowmde: What is the benefit of basically reverting the inlining we did here (and the same in other lines of code in this patch) and calling the getters again?

src/Entity/PropertyId.php
jakobw: Yes, it gets normalized to `prefix1:prefix2:Q42` in the constructor.

src/Entity/EntityId.php
thiemowmde: I will always merge it when I see it. Super trivial and super easy to understand and to read. I wish there would be a "non-empty string" check in Assert. This pattern repeats a thousand times in our code base.

src/Entity/EntityId.php
thiemowmde: The regex in ItemId and PropertyId do not allow `:a:Q1`, but this here does. One must be wrong.

src/Entity/EntityId.php
jakobw: I heard upcase before e.g. in the [Ruby standard library](https://ruby-doc.org/core-2.2.3/String.html#method-i-upcase) but I could change it.

src/Entity/ItemId.php
jakobw: It's normalized in the parent constructor.

src/Entity/EntityId.php
thiemowmde: The fact that this method turns the string upper case is an implementation detail, by the way, and should not be in the name. Something like "normalizeLocalId" would be better, I believe.

src/Entity/EntityId.php
jakobw: @thiemowmde I think you're looking at an outdated version of ItemId and PropertyId. Github shows some strange things on the "Conversation" page of the pull request since they introduced their new review features...

src/Entity/ItemId.php
```diff
@@ -23,8 +24,9 @@ class ItemId extends EntityId implements Int32EntityId {
 	 * @throws InvalidArgumentException
 	 */
 	public function __construct( $idSerialization ) {
-		$this->assertValidIdFormat( $idSerialization );
-		$this->serialization = strtoupper( $idSerialization );
+		parent::__construct( self::upcaseLocalId( $idSerialization ) );
```
thiemowmde: Why is upcaseLocalId called here, when both upcaseLocalId and the constructor call are in the parent class? The way this code is organized is a bit confusing and can mislead people into using this wrong. Basically: Not all entity IDs must be upper case. This is a decision the specific subclass (ItemId and PropertyId) must make. Which raises the question why the helper method is in the base class. Answer: Code sharing.

In the past we tried hard to get rid of code sharing in the Entity and EntityId base classes.

I suggest to rearrange the code in a way that you can reuse the split method, do a trivial `strtoupper( $array[2] )` here in the subclass and the pass the array to the base class.

src/Entity/ItemId.php
```diff
@@ -24,7 +25,7 @@ class ItemId extends EntityId implements Int32EntityId {
 	 */
 	public function __construct( $idSerialization ) {
 		$this->assertValidIdFormat( $idSerialization );
```
jakobw: I'm not sure if you already saw the latest version of this but validating the two parts completely separately is kind of a chicken-egg-problem since the splitting method also wants to validate the serialization. EntityId now looks at the entire string to validate the prefix and also makes sure that there's *something* at the end that does not contain colons. The last part is then validated in the subclass as it should be.

src/Entity/EntityId.php
jakobw: Good point. It's now checking that the last element is not an empty string and also makes sure `'0'` doesn't get filtered.

src/Entity/PropertyId.php
jakobw: The reason is that `serialization` is supposed to be private to the parent class. The only thing blocking that is that it's still set directly in the `unserialize` method.

src/Entity/EntityId.php
jakobw: Normalizing the local ID should be the concern of the subclass. This method is just there for convenience so that all subclasses can easily access it to normalize the local ID.

src/Entity/ItemId.php
jakobw: > I suggest to rearrange the code in a way that you can reuse the split method, do a trivial strtoupper( $array[2] ) here in the subclass and then pass the array to the base class.

I don't understand the last bit here. Where would I pass an array to the base class and what would the base class do with it?

src/Entity/PropertyId.php
thiemowmde: I suggest to not touch these "magic" serialization methods to much, and not discuss this unrelated issue here.

src/Entity/ItemId.php
thiemowmde: This method encodes knowledge about an implementation detail in the wrong class. The base class should not know anything about the fact that some subclasses do a magic "to upper case" conversion. Simply inline the `strtoupper` in the two subclasses.

jakobw: done!

src/Entity/EntityId.php
```diff
@@ -30,6 +55,106 @@ public function getSerialization() {
...
+
+	/**
+	 * Returns the serialization without the first repository prefix.
+	 *
+	 * @return string
+	 */
+	public function getLocalPart() {
+		$parts = self::splitSerialization( $this->serialization );
+
+		return self::joinSerialization( array( $parts[1], $parts[2] ) );
```
jakobw: I agree that having `''` as the first array element does not do anything to the return value. In fact I had it there in the beginning and removed it by request here: https://github.com/wmde/WikibaseDataModel/pull/678#discussion_r79435007. It does seem to confuse people though so I guess I'll add it again :)

manicki: Haha, yeah. After looking at at this code multiple times I even started to think maybe something like `return $parts[1] !== '' ? $parts[1] . $parts[2] : $parts[2];` would do better here :)

src/Entity/EntityId.php
brightbyte: is this fixed?

src/Entity/EntityId.php
brightbyte: is this fixed?

src/Entity/EntityId.php
brightbyte: @thiemowmde just make a pull request, it's "ours" anyway: https://github.com/wmde/Assert

src/Entity/EntityId.php
brightbyte: I discussed with Thiemo that it would be cleaner to drop the convenience function, and just copy the relevant line of code to the child constructors that need it. I don't care terribly much either way, though.

src/Entity/EntityId.php
brightbyte: Didn't we agree this morning to remove the prefix matching part from this pattern?

src/Entity/PropertyId.php
brightbyte: Yea, this can be left for later. If we can't change unserialize(), why change serialize().

JeroenDeDauw:
The new non-private methods still don't have `@since` tags

src/Entity/EntityId.php
```diff
@@ -30,6 +55,92 @@ public function getSerialization() {
...
 	/**
+	 * Returns an array with 3 elements: the foreign repository name as the first element, the local ID as the last
+	 * element and everything that is in between as the second element.
+	 *
+	 * EntityId::joinSerialization can be used to restore the original serialization from the parts returned.
+	 *
+	 * @param string $serialization
+	 * @return string[] Array containing the serialization split into 3 parts.
+	 */
+	public static function splitSerialization( $serialization ) {
```
JeroenDeDauw: As far as I can tell this could be protected and non-static

src/Entity/EntityId.php
```diff
@@ -17,6 +18,30 @@
...
 	protected $serialization;
 
+	const PATTERN = '/^:?(\w+:)*[^:]+\z/';
+
+	/**
+	 * @param string $serialization
+	 */
+	public function __construct( $serialization ) {
+		self::assertValidSerialization( $serialization );
+		$this->serialization = self::normalizeIdSerialization( $serialization );
```
JeroenDeDauw: Looking at the derivatives, I suspect this normalization is not needed

src/Entity/EntityId.php
```diff
@@ -30,6 +55,92 @@ public function getSerialization() {
...
+		);
+	}
+
+	/**
+	 * Returns '' for local IDs and the foreign repository name for foreign IDs. For chained IDs (e.g. foo:bar:Q42) it
+	 * will return only the first part.
+	 *
+	 * @return string
+	 */
+	public function getRepoName() {
```
JeroenDeDauw: getRepositoryName

src/Entity/EntityId.php
```diff
@@ -30,6 +55,92 @@ public function getSerialization() {
...
+
+		return $parts[0];
+	}
+
+	/**
+	 * Returns the serialization without the first repository prefix.
+	 *
+	 * @return string
+	 */
+	public function getLocalPart() {
```
JeroenDeDauw: I'm a bit confused by this naming. If you only remove the first repository prefix, then the remainder can still be non-local right?

tests/unit/Entity/EntityIdTest.php
```diff
@@ -81,4 +85,113 @@ public function testReturnTypeOfToString( EntityId $id ) {
...
+	 */
+	public function testSplitSerialization( $serialization, $split ) {
+		$this->assertSame( $split, EntityId::splitSerialization( $serialization ) );
+	}
+
+	/**
+	 * @dataProvider invalidSerializationProvider
+	 */
+	public function testSplitSerializationFails_GivenInvalidSerialization( $serialization ) {
+		$this->setExpectedException( 'InvalidArgumentException' );
```
JeroenDeDauw: InvalidArgumentException::class

tests/unit/Entity/EntityIdTest.php
```diff
@@ -81,4 +85,113 @@ public function testReturnTypeOfToString( EntityId $id ) {
...
+	 */
+	public function testJoinSerialization( $serialization, $split ) {
+		$this->assertSame( $serialization, EntityId::joinSerialization( $split ) );
+	}
+
+	/**
+	 * @dataProvider invalidJoinSerializationDataProvider
+	 */
+	public function testJoinSerializationFails_GivenEmptyId( $parts ) {
+		$this->setExpectedException( 'InvalidArgumentException' );
```
JeroenDeDauw: InvalidArgumentException::class

tests/unit/Entity/EntityIdTest.php
```diff
@@ -81,4 +85,113 @@ public function testReturnTypeOfToString( EntityId $id ) {
...
+			array( 42 ),
+			array( null ),
+		);
+	}
+
+	/**
+	 * @dataProvider invalidSerializationProvider
+	 */
+	public function testConstructor( $serialization ) {
+		$this->setExpectedException( 'InvalidArgumentException' );
```
JeroenDeDauw: InvalidArgumentException::class

tests/unit/Entity/ItemIdTest.php
```diff
@@ -147,4 +151,9 @@ public function invalidNumericIdProvider() {
 		);
 	}
 
+	public function testGetNumericIdThrowsExceptionOnForeignIds() {
+		$this->setExpectedException( 'RuntimeException' );
```
JeroenDeDauw: RuntimeException::class

tests/unit/Entity/PropertyIdTest.php
```diff
@@ -147,4 +150,9 @@ public function invalidNumericIdProvider() {
 		);
 	}
 
+	public function testGetNumericIdThrowsExceptionOnForeignIds() {
+		$this->setExpectedException( 'RuntimeException' );
```
JeroenDeDauw: RuntimeException:class

tests/unit/Entity/ItemIdTest.php
```diff
@@ -21,24 +21,28 @@ class ItemIdTest extends PHPUnit_Framework_TestCase {
...
+			array( 'q1', 'Q1' ),
+			array( 'q100', 'Q100' ),
+			array( 'q1337', 'Q1337' ),
+			array( 'q31337', 'Q31337' ),
+			array( 'Q31337', 'Q31337' ),
+			array( 'Q42', 'Q42' ),
+			array( ':Q42', 'Q42' ),
+			array( 'foo:Q42', 'foo:Q42' ),
+			array( 'foo:bar:q42', 'foo:bar:Q42' ),
+			array( 'Q42', 'Q42' ),
```
JeroenDeDauw: Now this is there twice

JeroenDeDauw:
I think the responsibility of this string splitting is misplaced and that all this static code is going down the wrong path. Follow up: https://github.com/wmde/WikibaseDataModel/pull/681

src/Entity/EntityId.php
jakobw: Yes, it turned out to be tricky naming this one. Suggestions are very welcome!

src/Entity/EntityId.php
brightbyte: Ah sorry, this is in EntityId. I would have expected the pattern for checking the prefix syntax to have a different constant name than the patterns that check the final parts of the ID.

brightbyte:
@JeroenDeDauw I agree in principle, but I have not yet seen a good alternative. The current solution isn't perfect, but I think the cost in terms of technical dept is low.

The problem as I see it is that there is now a shared responsibility between the EntityId classes and the EntityIdParser implementations: both need to be able to split the ID into the relevant three parts (the EntityId for validation and normalization, the parser for mapping prefixes and picking the right instantiator callback for the EntityId).

I want to keep the knowledge about the syntax used for these parts in one place (currently, in static methods in EntityId). A cleaner way to do this would be an EntityIdPartitioner class I suppose (with "parser" being taken). But the EntityId's constructor would need access to this, which makes things annoying again.

In any case, the syntax used for prefixing IDs is not "pluggable", it has to be the same for all types of entities, and more importantly, it has to be the same across all repos to allow federation. Because of this, I think it's ok to encode it in static methods bound to the EntityId base class: because it is part of the contract of all EntityIds.

tests/unit/Entity/EntityIdTest.php
jakobw: Fixed this and all similar cases!

src/Entity/EntityId.php
jakobw: done!

src/Entity/EntityId.php
jakobw: yes!

src/Entity/PropertyId.php
jakobw: Changed it back to not using `getSerialization`

src/Entity/EntityId.php
thiemowmde: +1

src/Entity/EntityId.php
brightbyte: Does it work with https://github.com/wmde/WikibaseDataModelServices/pull/146 ?

src/Entity/EntityId.php
jakobw: done

src/Entity/EntityId.php
manicki: I've just checked, and yes, it does

brightbyte: I'm confused - I thought the EntityIdParser would need splitSerialization, which is removed in Jeroen's PR? I'll have to look at it more closely.

src/Entity/EntityId.php
brightbyte: we need to make sure that any leading ":" is always stripped.

src/Entity/EntityId.php
brightbyte: I'll merge this for now, we can still discuss Jeroen's change before the next release.

PR #679, opened 2016-09-19
Add support for repo prefixes in DispatchingEntityIdParser

manicki:
~~This includes code from https://github.com/wmde/WikibaseDataModel/pull/678 and should only be merged after `EntityId` handles foreign IDs.~~

### DIFF
--- a/src/Entity/DispatchingEntityIdParser.php
+++ b/src/Entity/DispatchingEntityIdParser.php
@@ -44,7 +44,14 @@ class DispatchingEntityIdParser implements EntityIdParser {
 		}
 
 		foreach ( $this->idBuilders as $idPattern => $idBuilder ) {
-			if ( preg_match( $idPattern, $idSerialization ) ) {
+			try {
+				list( , , $localId ) = EntityId::splitSerialization( $idSerialization );
+			} catch ( InvalidArgumentException $ex ) {
+				// EntityId::splitSerialization performs some sanity checks which
+				// might result in an exception. Should this happen, re-throw the exception message
+				throw new EntityIdParsingException( $ex->getMessage() );
+			}
+			if ( preg_match( $idPattern, $localId ) ) {
 				return $this->buildId( $idBuilder, $idSerialization );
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
So this parser will drop the prefixes?

manicki:
So the idea was to ignore prefixes (foreign repo name etc) while picking the right "builder" to build the Id object. But the prefix is not removed, ie. `EntityId` objects gets "full" serialized id when is constructed.

OTOH as https://github.com/wmde/WikibaseDataModel/pull/678 changes `ItemId::PATTERN` and `PropertyId::PATTERN` in a way that `DispatchingEntityIdParser` will still pick a right builder this change is not really necessary to have parser working as expected.

I am not sure if it is worth to have Parser explicitly skip prefixes in regexp match, or should I just change this PR to simply adding tests ensuring that IDs with prefixes are all parsed correctly?

brightbyte:
@manicki I'm a bit torn about whether ItemId::PATTERN and PropertyId::PATTERN should include the pattern, to be honest. I think I went back and forth three times today. Let's discuss it again tomorrow. For now, doing it "both ways" doesn't do any harm.

JeroenDeDauw:
> But building regular expressions by concatenating strings is icky.

cat-hiding-in-shoe

brightbyte:
@manicki please rebase, I just merged the EntityId changes

src/Entity/DispatchingEntityIdParser.php
```diff
@@ -44,7 +44,8 @@ public function parse( $idSerialization ) {
 		}
 
 		foreach ( $this->idBuilders as $idPattern => $idBuilder ) {
-			if ( preg_match( $idPattern, $idSerialization ) ) {
+			$serializationParts = EntityId::splitSerialization( $idSerialization );
```
manicki: done but called it `$localId` as it is not what is called a local par t:)

manicki:
rebased

PR #682, opened 2016-09-29
Add Foreign Entity Ids specification

manicki:
Based on https://phabricator.wikimedia.org/T133381 (enhanced copy-and-paste).

### DIFF
--- a/docs/foreign-entity-ids.wiki
+++ b/docs/foreign-entity-ids.wiki
@@ -5,11 +5,10 @@ and, similarly, repositories other than local are '''foreign''' repositories.
 Notes:
 * The name used to refer to a given repository is local, and can differ from client to client.
 * The empty repository name "" refers to the local wiki.
-* Repository names can be used to look up the repository Site object, which in turn allows the ID to be mapped to the currect URI, URL, or API path for the repository it belongs to.
+* Repository names can be used to look up the repository Site object, which in turn allows the ID to be mapped to the correct URI, URL, or API path for the repository it belongs to.
 * The serialization of the EntityId follows the pattern <code><repository>:<id></code>. If <code><repository></code> is empty, the leading <code>:</code> is optional. This format follows the convention set by XML namespaces, RDF prefixes, and MediaWiki interwiki links.
 * Repository names can be mapped during serialization and deserialization.
-* When reading data from a another repository, repository names in entity IDs get mapped from the names used on the other repository to the names used on the local wiki. In particular, the empty repository prefix is mapped to the name of the foreign repository while reading IDs.
-* When writing data to a another repository, repository names in entity IDs get mapped from the names used on the local wiki to the names used on the foreign repository. In particular, the foreign wiki's name is mapped to the empty prefix.
+* When reading data from another repository, repository names in entity IDs get mapped from the names used on the other repository to the names used on the local wiki. In particular, the empty repository prefix is mapped to the name of the foreign repository while reading IDs.
 * The repository name in an EntityId object can always be assumed to match the local definition of that name. A repository name in serialized data however cannot be interpreted without knowing which repository it came from.
 
 == Prefix mapping ==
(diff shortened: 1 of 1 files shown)


docs/foreign-entity-ids.wiki
```diff
@@ -0,0 +1,31 @@
...
+* The empty repository name "" refers to the local wiki.
+* Repository names can be used to look up the repository Site object, which in turn allows the ID to be mapped to the currect URI, URL, or API path for the repository it belongs to.
+* The serialization of the EntityId follows the pattern <code><repository>:<id></code>. If <code><repository></code> is empty, the leading <code>:</code> is optional. This format follows the convention set by XML namespaces, RDF prefixes, and MediaWiki interwiki links.
+* Repository names can be mapped during serialization and deserialization.
+* When reading data from a another repository, repository names in entity IDs get mapped from the names used on the other repository to the names used on the local wiki. In particular, the empty repository prefix is mapped to the name of the foreign repository while reading IDs.
+* When writing data to a another repository, repository names in entity IDs get mapped from the names used on the local wiki to the names used on the foreign repository. In particular, the foreign wiki's name is mapped to the empty prefix.
```
jakobw: I might have missed something but are we ever writing data to another repository?

docs/foreign-entity-ids.wiki
```diff
@@ -0,0 +1,31 @@
...
+
+== Prefix mapping ==
+
+When receiving prefixed entities from another repository, prefixes are "chained", like interwiki prefixes:
+
+* When reading <code>d:Q5</code> from repository <code>foo</code>, it is turned into <code>foo:d:Q5</code>, meaning ''<code>Q5</code> at the repo that repo <code>foo</code> calls <code>d</code>''
+* The local repository may have mappings defined for the prefixes used by other wikis, e.g. <code>foo:d</code> would be known to be the same as the local prefix <code>wd</code> for Wikidata. Mappings are defined as a two-dimensional array, e.g. part of mapping definition for example mentioned here would look like below:
+<code> array( ... 'foo' => array( 'd' => 'wd' ), ... )</code>
+
+* During deserialization data from another repository, the name of the source repository is always added as a prefix, and then any known mappings are resolved: <code>d:Q5</code> from <code>foo</code> becomes <code>foo:d:Q5</code> and then <code>wd:Q5</code>.
```
jakobw: "While deserializing data from another repository" or "deserialization of data"

docs/foreign-entity-ids.wiki
```diff
@@ -0,0 +1,31 @@
...
+* If no mapping is known, the "chained" version of the ID (<code>foo:d:Q5</code>) is stored locally. If this kind of ID is sent to yet another repo, that may result in longer "chains" of prefixes, like <code>xyz:foo:d:Q5</code>.
+* During deserialization data from the local database, no prefix is added, but any known mappings are resolved. If <code>foo:d:Q5</code> was stored earlier because there was no mapping defined for <code>foo:d</code> then, but there is a mapping now, that mapping is resolved when loading the item.
```
jakobw: "While deserializing data from another repository" or "deserialization of data"

docs/foreign-entity-ids.wiki
```diff
@@ -0,0 +1,31 @@
...
+* Note that <code>foo:bar:Q5</code> and <code>bar:foo:Q5</code> may mean different things (or the same thing), depending on the mappings defined in repositories <code>foo</code> and <code>bar</code>.
+
+EntityId object always ''knows'' which repository it belongs to, and it always reflects the currently defined mappings.
```
jakobw: > EntityId object always knows which repository it belongs to

Isn't all it knows whether it is foreign or not? I think which repository it belongs to might not even always be known to the local wiki.

jakobw: Should be "An EntityId object"

manicki: It might depend on how you look at it. But there is as public method `EntityId::getRepositoryName` which returns the name of repo.
For the latter part of your comment (" think which repository it belongs to might not even always be known to the local wiki.") I don't think this could happen. I mean there could be a case when an entity is loaded from the foreign repository `foo`. Then the ID that local repo uses for the it is, for instance, `foo:Q123`. If for whatever reason, `foo` is removed from repository config (ie. it's no longer a known repo), `foo:Q123` would be interpretted as `:foo:Q123`. Which obviously would have various consequences (for example: what happens during lookup? Local repo should then have the foreign entity copied before the repo config changes)

docs/foreign-entity-ids.wiki
manicki: I guess you're right, this part is not really relevant for what we're planning to do (at least in first iteration). I believe the description still makes sense but I will remove it for know, so it does not create confusion ("Why you throw exception here when docs define how to write stuffs to foreign repo?")

PR #684, opened 2016-10-14
Release 6.2.0

jakobw:


manicki:
Please wait for https://github.com/wmde/WikibaseDataModel/pull/685 with releasing (whoever does it)!

manicki:
I am wondering shouldn't this also mention in some way that EntityId's constructor has become public again. You don't create instances of abstract classes but the public interface has changed, still.

thiemowmde:
If the constructor parameters changed and a third-party subclass calling `parent::__construct` must be changed because of this, then yes, this is relevant and should be mentioned.

manicki:
I believe this is what happened in https://github.com/wmde/WikibaseDataModel/pull/678/files so I guess this change should be listed too.

jakobw:
Thanks, @manicki. Added a line mentioning the now public constructor.

PR #687, opened 2016-11-01
Add functions to assert a value is a valid repository name, or an array indexed by valid repository names

manicki:
This is re-submission of https://github.com/wmde/WikibaseDataModelServices/pull/155, as it has been suggested it better fits in here.

Background (repeated): similar checks are done in https://github.com/wmde/WikibaseDataModelServices/pull/150,  https://github.com/wmde/WikibaseDataModelServices/pull/150, https://gerrit.wikimedia.org/r/#/c/314682/, https://github.com/wmde/WikibaseDataModelServices/pull/146, https://github.com/wmde/WikibaseDataModelServices/pull/149. It makes sense to have dedicated function(s) for such things. and not duplicate validation code. It would also make it easier to only have to adjust validation code in one place if there are more restrictions added on repository names etc.

Few little things changed here compared to https://github.com/wmde/WikibaseDataModelServices/pull/155:
 - moved to DM component (obviously),
 - class has been renamed,
 - public functions got renamed.

Open questions:
 - should those rather be global functions instead of pretending this code is object-oriented?
 - should this code rather go to `RepositoryName` value object instead of introducing more static code?
 - names....

Ping @brightbyte @jakobw @JeroenDeDauw

### DIFF
--- a/src/Assert/RepositoryNameAssert.php
+++ b/src/Assert/RepositoryNameAssert.php
@@ -1,0 +1,64 @@
+<?php
+
+namespace Wikibase\DataModel\Assert;
+
+use Wikimedia\Assert\Assert;
+use Wikimedia\Assert\ParameterAssertionException;
+
+/**
+ * Provides functions to assure values are allowable repository
+ * names in Wikibase.
+ *
+ * @see docs/foreign-entity-ids.wiki
+ * @see Wikimedia\Assert\Assert
+ *
+ * @since 6.3
+ *
+ * @license GPL-2.0+
+ */
+class RepositoryNameAssert {
+
+	/**
+	 * @param string $value The actual value of the parameter
+	 * @param string $name The name of the parameter being checked
+	 *
+	 * @throws ParameterAssertionException If $value is not a valid repository name.
+	 */
+	public static function assertParameterIsValidRepositoryName( $value, $name ) {
+		if ( !self::isValidRepositoryName( $value ) ) {
+			throw new ParameterAssertionException( $name, 'must be a string not including colons' );
+		}
+	}
+
+	/**
+	 * @param array $values The actual value of the parameter. If this is not an array,
+	 *        a ParameterTypeException is thrown.
+	 * @param string $name The name of the parameter being checked
+	 *
+	 * @throws ParameterAssertionException If any element of $values is not
+	 *         a valid repository name.
+	 */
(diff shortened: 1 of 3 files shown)


brightbyte:
@manicki No to the value object approach, for performance reasons.

I agree that a RepositoryName object would nicely solve the modeling issue for the assertions as well as the splitting/joining.  But object granularity is a thing. Let's not add more overhead to deserializing entities unless we really have to. The performance is already quite bad.

brightbyte:
Re global functions vs. static methods: the are functionally equivalent. We are kind of used to think "function -> bad", but static methods are really the same, and both are ok-ish if they do not access (or worse, modify) global state. Core uses some namespaced functions now, i think.

So, I don't care much either way.

addshore:
To me this seems a bit odd in this library. I mean it uses absolutely nothing in here but pulls in a new dependency (assert).
It seems like right now there may not be a better place for this stuff?
But it is a shame that now my libraries that for example use datamodel how have the assert dependency and also this code when they really don't need it.

+0 from me, I don't know!

addshore:
+0.1

src/Assert/RepositoryNameAssert.php
```diff
@@ -0,0 +1,64 @@
...
+				);
+			}
+		}
+	}
+
+	/**
+	 * @param string $value
+	 * @return bool
+	 */
+	private static function isValidRepositoryName($value ) {
```
thiemowmde: Spacing issue.

tests/unit/Assert/RepositoryNameAssertTest.php
```diff
@@ -0,0 +1,85 @@
...
+	 * @dataProvider provideInvalidRepositoryNames
+	 */
+	public function testGivenInvalidValue_assertParameterIsValidRepositoryNameFails( $value ) {
+		$this->setExpectedException( ParameterAssertionException::class );
+		RepositoryNameAssert::assertParameterIsValidRepositoryName( $value, 'test' );
+	}
+
+	public function provideValidRepositoryNames() {
+		return [
+			[ '' ],
```
thiemowmde: Why is an empty string a valid repository name?

src/Assert/RepositoryNameAssert.php
manicki: Oh dear. Fixed in https://github.com/wmde/WikibaseDataModel/pull/692

PR #690, opened 2016-11-04
FingerprintHolder is deprecated use FingerprintProvider instead

Ladsgroup:


manicki:
Looks good to me, although I don't confident enough to say if it has any bad consequences.
@brightbyte @mariushoch @addshore @thiemowmde thoughts?

Also, could we remove FingerprintHolder for good at this point?

mariushoch:
`FingerprintHolder` says "will be removed in 6.0 in favor of FingerprintProvider", so let's go ahead here.

Verified unused in all Wikidata components, `FingerprintHolder` should be dropped in a follow up.

JeroenDeDauw:
This is a breaking change. If you merge this into master, which you did, you should also update the version to 7.0.x-dev (don't forget composer.json).

To me this makes little sense unless you actually get rid of the deprecated interface as well.

PR #694, opened 2016-11-22
Allow passing a repository name when creating ItemId and PropertyId from number

manicki:


src/Entity/ItemId.php
```diff
@@ -95,16 +96,18 @@ public function unserialize( $serialized ) {
...
 	 *
 	 * @return self
 	 * @throws InvalidArgumentException
 	 */
-	public static function newFromNumber( $numericId ) {
+	public static function newFromNumber( $numericId, $repositoryName = '' ) {
 		if ( !is_numeric( $numericId ) ) {
 			throw new InvalidArgumentException( '$numericId must be numeric' );
 		}
+		RepositoryNameAssert::assertParameterIsValidRepositoryName( $repositoryName, '$repositoryName' );
```
jakobw: Is it intended to exclude nested repository prefixes for ItemIds like `foo:bar:Q123`? This means that not all ItemIds that can be creating with the constructor can be created with `newFromNumber`. I think it should at least be documented why some but not all foreign EntityIds can be created like this.

JeroenDeDauw: This throws `ParameterAssertionException`, not `InvalidArgumentException`, so the constructor doc is now wrong.

As a user of a method such as this one, I kinda expect to get a `InvalidArgumentException` when the argument is invalid. Forcing the user to bind to the assertion library used is not nice.

manicki: So I didn't change the constructor doc because ParameterAssertionException extends InvalidArgumentException. I believe using Assert's exception is OK as long as users expect InvalidArgumentException, no?

thiemowmde:
The code here is trivial and makes sense. But:
* I'm not sure if an optional parameter *after* the number is a good idea. This makes the method signature complicated, non-intuitive and harder to use. The same can be achieved by first constructing an EntityId object with the existing newFromNumber method, and then passing this to some builder code that adds the foreign repository prefix.
* Is this even necessary? Which code would call the newFromNumber method with these two parameters? If there is only a single place in the entire code base that will use this, please do not put this code in the DataModel. Instead please create a DataModel Service or even a local method in the component/namespace that actually needs this.

src/Entity/ItemId.php
brightbyte: In foo:bar:Q123, the repo name is "foo", not "foo:bar". The relative ID, the one that can be resolved by repo foo, is "bar:Q123".

By our spec, repo names cannot contain a colon, and EntityId::getRepositoryName() will never retrun a string that contains a column.

brightbyte:
>  I'm not sure if an optional parameter after the number is a good idea. This makes the method signature complicated, non-intuitive and harder to use.

Only one reason: if you add it at the end of the param list, it's not a breaking change. We can make it a breaking change, if you want. 

> The same can be achieved by first constructing an EntityId object with the existing newFromNumber method, and then passing this to some builder code that adds the foreign repository prefix.

Yes, but you would have to construct two EntityId objects. And how is that better anyway?

>    Is this even necessary? Which code would call the newFromNumber method with these two parameters? If there is only a single place in the entire code base that will use this, please do not put this code in the DataModel. Instead please create a DataModel Service or even a local method in the component/namespace that actually needs this.

Any code that reads numeric IDs and wants to construct an EntityId from it needs to consider which repo the ID belongs to. So most code that currently uses this method will need to change to also add a repo name, so it is able to process data from foreign repositories.

src/Entity/ItemId.php
manicki: @jakobw: as @brightbyte pointed out above, given you only have a number you want to create an entity id instance of, and the repository name (possibly "empty" one), it all simply limits you to only have ids like "foo:Q1234". `ItemId::newFromNumber` is kind of by its definition limited and not suppose to cover all that ItemId's constructor does.

manicki:
@thiemowmde:
>  Which code would call the newFromNumber method with these two parameters? 

So https://gerrit.wikimedia.org/r/#/c/322704/ is currently the only/first patch needing this. But as said above, more places using `newFromNumber` migth/should be needing this, once we get to them.

>  Instead please create a DataModel Service or even a local method in the component/namespace that actually needs this.

Even if it was the only one place needing to create the instance of the foreign entity id (which we think will not be the case) I am not sure if adding another service for dealing with legacy identifiers (with a big fat note "Please do not use me unless really necessary" from its very beginning) would really be better.

src/Entity/ItemId.php
JeroenDeDauw: In this case you should just test for `InvalidArgumentException`. This is probably why I got confoosed about this in the first place

brightbyte:
As per the IRL discussion with @thiemowmde and @manicki, this can be abandoned. Instead of extending the functionaliz of newFromNumber, we'll deprecate it and phase it out in favor of using callbacks defined in the entity type definition, and called from EntityIdComposer.

JeroenDeDauw:
> the entity type definition

What's this? Classes such as `Item` and `Property`?

manicki:
As @brightbyte said, we decided no to go the way this PR was proposing. There will be tickets on phabricator up soon, and we should continue discussion over there. Tickets will be put here, so everyone knows where we continue.

> > the entity type definition
>
> What's this? Classes such as Item and Property?

No, it is things like https://phabricator.wikimedia.org/diffusion/EWBA/browse/master/lib/WikibaseLib.entitytypes.php
In particular, what we have in mind is having `entity-id-composer` callback (including the rename of the filed in the wiring file, but this is different topic), like the one in MediaInfo https://phabricator.wikimedia.org/diffusion/EWBI/browse/master/WikibaseMediaInfo.entitytypes.php;90b82546ca3971d79fa24b21a641747cb82ec469$101 defined for each entity type (ie. there will be no default feedback for Item and Property in EntityIdComposer calling newFromNumber static methods).

JeroenDeDauw:
Kinda nice to have the tickets before closing, so there is continuation.

Please keep in mind that relying on Wikibase Lib, and by extension the whole of MediaWiki, is expensive

manicki:
Related tickets are:
https://phabricator.wikimedia.org/T151573
https://phabricator.wikimedia.org/T151574
https://phabricator.wikimedia.org/T151575
https://phabricator.wikimedia.org/T151576

PR #698, opened 2016-12-05
Avoid calling EntityId::newFromNumber when not needed

thiemowmde:


Ladsgroup:
:+1:

tests/unit/Statement/StatementListTest.php
```diff
@@ -42,7 +42,7 @@ private function getStatement( $propertyId, $guid, $rank = Statement::RANK_NORMA
 
 		$statement->expects( $this->any() )
 			->method( 'getPropertyId' )
-			->will( $this->returnValue( PropertyId::newFromNumber( $propertyId ) ) );
+			->will( $this->returnValue( new PropertyId( 'P' . $propertyId ) ) );
```
thiemowmde: In production code I would agree, but this is a test. All numbers used here are integer constants from test providers in this file, and never user input.

JeroenDeDauw: What difference does it make if this is user input or not?

thiemowmde: Because I can now do this: #699. I can't do this when it's user input.

PR #701, opened 2016-12-24
Avoid PHPUnit deprecation warnings

JeroenDeDauw:
Recent versions of PHPUnit deprecated some stuff
that we can't get rid off since that would mean
using new stuff only available in versions of PHPUnit
that do not support PHP 5.5.

Yay for supporting legacy PHP /o\

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,6 +28,7 @@
 		"wikimedia/assert": "~0.2.2"
 	},
 	"require-dev": {
+		"phpunit/phpunit": "~4.8",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -28,6 +28,7 @@
 		"wikimedia/assert": "~0.2.2"
 	},
 	"require-dev": {
```
thiemowmde: Does not matter much, but alphabetical order would be nice.

JeroenDeDauw: Eh. I think that's a waste of time. Don't let me stop you though

PR #702, opened 2017-01-05
Refactored away HashArray

JeroenDeDauw:
Removed a bunch of unused code and made SnakList simpler.

SnakList can be improved further in follow ups:

* Constructor can hide extending of ArrayObject
* Perhaps can use composition of ArrayObject instead of inheritance
* removeSnakHash -> removeSnakByHash
* ...

The changes made so far should not break anything in Wikibase.git, though clearly there are some breaking changes (public methods no longer there and no more HashArray).


src/Snak/SnakList.php
```diff
@@ -15,17 +19,270 @@
...
+	/**
+	 * @see ArrayObject::__construct
+	 *
+	 * @param array|Traversable|null $input
+	 * @param int $flags
+	 * @param string $iteratorClass
+	 *
+	 * @throws InvalidArgumentException
+	 */
+	public function __construct( $input = null, $flags = 0, $iteratorClass = 'ArrayIterator' ) {
```
thiemowmde: In my opinion this constructor is exposing to much (undocumented) complexity that is not needed by users of the `SnakList` class. Please tell me if I miss something. Otherwise I suggest to merge #439 first and rebase this patch.

src/Snak/SnakList.php
```diff
@@ -15,17 +19,270 @@
...
+		parent::__construct( [], $flags, $iteratorClass );
+
+		if ( $input !== null ) {
```
thiemowmde: This will not be necessary any more with #439.

src/Snak/SnakList.php
JeroenDeDauw: I did not touch this in this PR and don't want to add additional changes to it.

thiemowmde: I consider the `$flags` and `$iteratorClass` parameters implementation details from the `ArrayObject` class. These parameters where never documented for the `SnakList` class, and never used, and should not be copy pasted over here in my opinion. As said I would like to merge #439 first.

JeroenDeDauw:
Merge conflict as well

thiemowmde:
I will happily rebase this patch here when the other patches are merged.

JeroenDeDauw:
Merge?

thiemowmde:
https://github.com/wmde/WikibaseDataModel/pull/702#pullrequestreview-15949890?

src/Snak/SnakList.php
```diff
@@ -15,17 +19,247 @@
...
+	private function preSetElement( $index, $hashable ) {
+		$hash = $hashable->getHash();
+
+		$hasHash = $this->hasSnakHash( $hash );
+
+		if ( $hasHash ) {
+			return false;
+		}
+
+		if ( $hasHash ) {
```
thiemowmde: This is now dead code. See #712, where I remove this entire feature from the base class. As before I will happily rebase this patch when #712 is merged.

src/Snak/SnakList.php
```diff
@@ -15,17 +19,247 @@
...
  * @author Addshore
  */
-class SnakList extends HashArray {
+class SnakList extends ArrayObject implements Hashable, Comparable {
 
 	/**
-	 * @see GenericArrayObject::getObjectType
+	 * Maps snak hashes to their offsets.
+	 *
+	 * @var array [ snak hash (string) => array [ snak offset (string|int) ] | snak offset (string|int) ]
```
thiemowmde: This can not be an array of arrays any more. Multiple `is_array` checks below are not needed any more. See #712.

thiemowmde:
Rebased. I also inlined what was formerly a protected `preSetElement` method. This is now much closer to being mergable, I believe. Still to do:
* Some `@since` tags are missing.
* Make sure the test cases cover everything we want them to cover.
* Rework the constructor. However, we can do this later, but before doing a release.
* The method name `removeSnakHash` is new. It's a bit awkward. Suggestion: `removeSnakByHash`.
* Check if all non-private methods are really needed. Some appear to be duplicates.

JeroenDeDauw:
I do not like this "everything needs to be done before this is merged" approach. Not saying it is bad. I find it unpleasant and think its not efficient. I do not want discuss this with you, and I no longer feel like following up on this change either, it's just not fun.

thiemowmde:
Sorry to hear that. All I wanted was a series of smaller patches that are easier to follow, easier to understand and easier to review.

Thanks for merging the patches I did so far.

JeroenDeDauw:
Just noticed this HasArray thing still exists, went looking for a PR that removes it, found this one, and went faceplam. Forgetti much

thiemowmde:
Thanks. I updated the release notes in 16f54e944cdebd2ea7959026032c9d9d3c4ab386. I used reflection to compile a list of methods SnakList does not support any more.

JeroenDeDauw:
Thanks for updating the release notes

PR #707, opened 2017-01-06
Use assertSame for Statement::getPropertyId

thiemowmde:
This is actually supposed to return the same object, not just an equal one. Not that critical because the returned object is immutable anyway, but still.


JeroenDeDauw:
-2

> In computer science, a value object is a small object that represents a simple entity whose equality is **not based on identity**

https://en.wikipedia.org/wiki/Value_object

thiemowmde:
They should never return different instances, since one is supposed to be a shortcut for the other. The return values should be identical, and this should be tested.

PR #709, opened 2017-01-06
Add missing ReferenceList::addReference test case

thiemowmde:
These two additions (a new test case for inserting a reference between two existing ones, and the exception) solve two of the three mutation testing issues reported in https://github.com/wmde/WikibaseDataModel/pull/654#issuecomment-270764683.

The remaining issue (where `>=` was mutated into `>`) is a trivial performance optimization. Both code paths (straight appending and inserting at the top most position) are supposed to produce the same output.

### DIFF
--- a/tests/unit/ReferenceListTest.php
+++ b/tests/unit/ReferenceListTest.php
@@ -143,7 +143,7 @@ class ReferenceListTest extends PHPUnit_Framework_TestCase {
 			$array->removeReference( $element );
 
 			$this->assertFalse( $array->hasReference( $element ) );
-			$this->assertEquals( --$elementCount, count( $array ) );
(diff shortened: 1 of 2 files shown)


tests/unit/ReferenceListTest.php
```diff
@@ -143,7 +143,7 @@ public function testRemoveReference( ReferenceList $array ) {
 			$array->removeReference( $element );
 
 			$this->assertFalse( $array->hasReference( $element ) );
-			$this->assertEquals( --$elementCount, count( $array ) );
+			$this->assertSame( --$elementCount, count( $array ) );
```
JeroenDeDauw: lololol loop :D

thiemowmde: I don't understand such comments. :-(

PR #710, opened 2017-01-10
Drop unused HashArray::removeDuplicates

thiemowmde:
This is split from #702 to make it easier to review.

Note that we already removed ReferenceList::removeDuplicates in 5.0.

The two testHasElement and testRemoveElement tests are moved without touching them, because they are not applicable to a test setup with duplicates.


JeroenDeDauw:
Merge conflict

thiemowmde:
That was expected. I manually rebased it now. Thanks for the reviews!

PR #713, opened 2017-01-14
Move HashArray::equals to SnakList::equals

thiemowmde:
This goes along with #714 (and will run into a merge conflict – feel free to merge whichever first, I will do the rebase). Both are split from #702 to make that big patch it easier to review later.

Note that I rewrote the tests entirely. This is also the main reason for this patch, to make sure SnakList is properly covered with tests before we are applying the big change in #702.

### DIFF
--- a/tests/unit/Snak/SnakListTest.php
+++ b/tests/unit/Snak/SnakListTest.php
@@ -2,6 +2,7 @@
 
 namespace Wikibase\DataModel\Tests\Snak;
 
+use Comparable;
 use DataValues\StringValue;
 use Hashable;
 use InvalidArgumentException;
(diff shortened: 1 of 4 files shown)


brightbyte:
please rebase. +2 would merge

thiemowmde:
Manually rebased, and made sure the changes are identical to before.

tests/unit/Snak/SnakListTest.php
```diff
@@ -294,6 +295,50 @@ public function testOrderByProperty( SnakList $snakList, SnakList $expected, arr
...
+			],
+			'different objects with same content are equal' => [
+				$oneSnak,
+				new SnakList( [ new PropertyNoValueSnak( 1 ) ] ),
+				true
+			],
+			'different objects with different content are not equal' => [
+				$oneSnak,
+				new SnakList( [ new PropertyNoValueSnak( 2 ) ] ),
+				false
```
manicki: I'd rather see this as a separate test (as in separate method) instead of squashing both test case data, and the expected result in the data provider method but whatever.

thiemowmde: Both are valid approaches, and even if I agree that splitting test cases is the better approach in many cases, I disagree in this case. Readability is key here, and the names of these test cases almost read like a poem. This poem will be destroyed if the last one is not an array key any more but a method name. I don't think this is worth it in this case, if the only benefit is that five booleans can be removed.

manicki: On the off-topic side: if you really want to compare source code to literature works you might consider using "prose" in comparison. I wouldn't consider "readability" or "clarity" an obvious value of "good" poem :)

thiemowmde: I can totally read this out aloud as a poem. Hm. Let me add this as a proposal to the Lightning Talk board.

PR #714, opened 2017-01-14
Move Hashable interface from HashArray to SnakList

thiemowmde:
This goes along with #713 (and will run into a merge conflict – feel free to merge whichever first, I will do the rebase). Both are split from #702 to make that big patch it easier to review later.

Note that I rewrote the tests entirely. This is also the main reason for this patch, to make sure SnakList is properly covered with tests before we are applying the big change in #702.

### DIFF
--- a/tests/unit/Snak/SnakListTest.php
+++ b/tests/unit/Snak/SnakListTest.php
@@ -3,6 +3,7 @@
 namespace Wikibase\DataModel\Tests\Snak;
 
 use DataValues\StringValue;
+use Hashable;
 use InvalidArgumentException;
 use Wikibase\DataModel\Entity\PropertyId;
 use Wikibase\DataModel\Snak\PropertyNoValueSnak;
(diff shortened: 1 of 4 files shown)


tests/unit/Snak/SnakListTest.php
```diff
@@ -292,4 +293,51 @@ public function testOrderByProperty( SnakList $snakList, SnakList $expected, arr
...
+		$this->assertSame( $self->getHash(), $other->getHash() );
+	}
+
+	public function provideEqualSnakLists() {
+		$empty = new SnakList();
+		$oneSnak = new SnakList( [ new PropertyNoValueSnak( 1 ) ] );
+
+		return [
+			'same equal object' => [ $empty, $empty ],
+			'same non-equal object' => [ $oneSnak, $oneSnak ],
```
bekh6ex: Probably you meant 'same empty object' and 'same non-empty object'

thiemowmde: Yea, thanks! Fixed.

PR #716, opened 2017-02-08
Don't use numeric IDs for internal PHP serialization.

brightbyte:
This changes Snak serialization to use the full string representation of property IDs.

NOTE: serialize/unserialize is used mainly when cloning Statements.
NOTE: it's unclear if this completely fixes T157442
NOTE: this changes snak hashes, and consequently reference hashes! This may break reference update/removal!

[Bug: T157442](https://phabricator.wikimedia.org/T157442)

### DIFF
--- a/src/Snak/PropertyValueSnak.php
+++ b/src/Snak/PropertyValueSnak.php
@@ -53,7 +53,7 @@ class PropertyValueSnak extends SnakObject {
 	 * @return string
 	 */
 	public function serialize() {
-		return serialize( [ $this->propertyId->getNumericId(), $this->dataValue ] );
+		return serialize( [ $this->propertyId->getSerialization(), $this->dataValue ] );
 	}
 
 	/**
@@ -64,8 +64,8 @@ class PropertyValueSnak extends SnakObject {
 	 * @param string $serialized
 	 */
 	public function unserialize( $serialized ) {
-		list( $numericId, $dataValue ) = unserialize( $serialized );
-		$this->__construct( $numericId, $dataValue );
+		list( $propertyId, $dataValue ) = unserialize( $serialized );
+		$this->__construct( self::newPropertyId( $propertyId ), $dataValue );
 	}
 
 	/**
--- a/src/Snak/PropertyValueSnak.php
+++ b/src/Snak/PropertyValueSnak.php
@@ -48,7 +48,7 @@ class PropertyValueSnak extends SnakObject {
 	/**
 	 * @see Serializable::serialize
 	 *
-	 * @since 0.1
+	 * @since 7.0
 	 *
 	 * @return string
 	 */
@@ -64,8 +64,14 @@ class PropertyValueSnak extends SnakObject {
 	 * @param string $serialized
 	 */
 	public function unserialize( $serialized ) {
-		list( $propertyId, $dataValue ) = unserialize( $serialized );
-		$this->__construct( self::newPropertyId( $propertyId ), $dataValue );
+		list( $propertyId, $this->dataValue ) = unserialize( $serialized );
+
+		if ( is_string( $propertyId ) ) {
+			$this->propertyId = new PropertyId( $propertyId );
+		} else {
+			// Backwards compatibility with the previous serialization format
+			$this->propertyId = PropertyId::newFromNumber( $propertyId );
+		}
 	}
 
 	/**
--- a/src/Snak/SnakObject.php
+++ b/src/Snak/SnakObject.php
@@ -102,7 +102,7 @@ abstract class SnakObject implements Snak {
 	 * @return string
 	 */
 	public function serialize() {
-		return serialize( $this->propertyId->getSerialization() );
+		return $this->propertyId->getSerialization();
 	}
 
 	/**
@@ -113,19 +113,31 @@ abstract class SnakObject implements Snak {
 	 * @param string $serialized
 	 */
 	public function unserialize( $serialized ) {
-		$this->propertyId = self::newPropertyId( unserialize( $serialized ) );
+		$this->propertyId = self::newPropertyId( $serialized );
 	}
 
 	/**
-	 * @param mixed $unserialized
+	 * @param string $serialized
 	 *
 	 * @return PropertyId
 	 */
-	protected static function newPropertyId( $unserialized ) {
+	protected static function newPropertyId( $serialized ) {
+		if ( is_int( $serialized ) ) {
+			// numeric id, already unserialized
+			return PropertyId::newFromNumber( $serialized );
+		}
+
+		try {
+			// full ID, as a string, not serialized
+			return new PropertyId( $serialized );
+		} catch ( InvalidArgumentException $ex ) {
+			// noop
+		}
+
+		$unserialized = unserialize( $serialized );
 		if ( is_int( $unserialized ) ) {
+			// numeric id, not previously unserialized
 			return PropertyId::newFromNumber( $unserialized );
-		} elseif ( is_string( $unserialized ) ) {
-			return new PropertyId( $unserialized );
 		} else {
 			throw new InvalidArgumentException( 'unexpected property ID serialization' );
 		}
--- a/src/Snak/SnakObject.php
+++ b/src/Snak/SnakObject.php
@@ -97,7 +97,7 @@ abstract class SnakObject implements Snak {
 	/**
 	 * @see Serializable::serialize
 	 *
-	 * @since 0.1
+	 * @since 7.0
 	 *
 	 * @return string
 	 */
@@ -113,33 +113,11 @@ abstract class SnakObject implements Snak {
 	 * @param string $serialized
 	 */
 	public function unserialize( $serialized ) {
-		$this->propertyId = self::newPropertyId( $serialized );
-	}
-
-	/**
-	 * @param string $serialized
-	 *
-	 * @return PropertyId
-	 */
-	protected static function newPropertyId( $serialized ) {
-		if ( is_int( $serialized ) ) {
-			// numeric id, already unserialized
-			return PropertyId::newFromNumber( $serialized );
-		}
-
 		try {
-			// full ID, as a string, not serialized
-			return new PropertyId( $serialized );
+			$this->propertyId = new PropertyId( $serialized );
 		} catch ( InvalidArgumentException $ex ) {
-			// noop
-		}
-
-		$unserialized = unserialize( $serialized );
-		if ( is_int( $unserialized ) ) {
-			// numeric id, not previously unserialized
-			return PropertyId::newFromNumber( $unserialized );
-		} else {
-			throw new InvalidArgumentException( 'unexpected property ID serialization' );
+			// Backwards compatibility with the previous serialization format
+			$this->propertyId = PropertyId::newFromNumber( unserialize( $serialized ) );
 		}
 	}
 
--- a/src/Snak/SnakObject.php
+++ b/src/Snak/SnakObject.php
@@ -126,8 +126,6 @@ abstract class SnakObject implements Snak {
 			return PropertyId::newFromNumber( $unserialized );
 		} elseif ( is_string( $unserialized ) ) {
 			return new PropertyId( $unserialized );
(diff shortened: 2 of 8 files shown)


brightbyte:
pinging @manicki @thiemowmde

src/Entity/EntityIdValue.php
```diff
@@ -31,8 +31,8 @@ public function __construct( EntityId $entityId ) {
 	 */
 	public function serialize() {
 		return json_encode( [
-			$this->entityId->getEntityType(),
-			$this->getNumericId()
+			get_class( $this->entityId ),
```
JeroenDeDauw: Why this change? Kinda scary to do `new $userInput()`

brightbyte: Yes, it's not pretty. But all we use this for is cloning statements.

The change is here because the old code assumes two things: a) the ID can be represented as a number (it can't) and b) we have a well known set of entity types (we don't). 

So, having the choice to try and access some kind of entity id type registry via a global, or putting the class name into the serialization, I opted for the latter.

src/Entity/EntityIdValue.php
```diff
@@ -58,10 +58,14 @@ private function getNumericId() {
 	 * @throws IllegalValueException
 	 */
 	public function unserialize( $serialized ) {
-		list( $entityType, $numericId ) = json_decode( $serialized );
+		list( $type, $id ) = json_decode( $serialized );
 
 		try {
-			$entityId = LegacyIdInterpreter::newIdFromTypeAndNumber( $entityType, $numericId );
+			if ( is_string( $id ) ) {
```
thiemowmde: This needs at least a `class_exists` check and a fallback solution (or return `null`?).

src/Snak/SnakObject.php
```diff
@@ -113,7 +113,24 @@ public function serialize() {
...
+	 * @param mixed $unserialized
+	 *
+	 * @return PropertyId
+	 */
+	protected static function newPropertyId( $unserialized ) {
+		if ( is_int( $unserialized ) ) {
+			return PropertyId::newFromNumber( $unserialized );
+		} elseif ( is_string( $unserialized ) ) {
+			return new PropertyId( $unserialized );
+		} elseif ( $unserialized instanceof PropertyId ) {
```
thiemowmde: Which code path needs this?

src/Entity/EntityIdValue.php
JeroenDeDauw: My remark is not about prettiness, it is about security.

> But all we use this for is cloning statements.

That does no reassure me. Kinda like saying "no need to escape this SQL argument since right now only we call this function with non-user-input".

> So, having the choice to try and access some kind of entity id type registry via a global, or putting the class name into the serialization, I opted for the latter.

Why not just (de)serialize the ID?

brightbyte: Oh, I definitely do not want this in the database, or in our external representation! As far as I am aware, this is not used in any way with our JSON encoding. We only use the php-serialization temporarily, to do deep clones. PHP-Serialization is extremely brittle (sensitive against changes to private members, etc), so it should never be used for persistence.

The alternative would be to change the way we do cloning, and avoid copying immutable objects when we clone. That would be nice, but I can't think of a reliable and easy way of doing this. serialize/unserialize is a very convenient method to do deep cloning safely. But it's somewhat wasteful, since it also copies immutable objects.

src/Snak/SnakObject.php
brightbyte: Probably none atm, it's here for completeness. I don't have strong feelings about having it.

src/Entity/EntityIdValue.php
brightbyte: We can't deserialize the ID without access to a service. How do we get the right service instance? This is particularly tricky for entity parsing, since we use different EntityIdParsers for data coming from different repos.

brightbyte: But I appreciate your concern for security. I'm not very happy about this solution. I just don't see an alternative, apart from not using serialize/unserialize for cloning statements.

JeroenDeDauw: > We can't deserialize the ID without access to a service.

Huh?

```php
deserialize( $entityId );
```

thiemowmde: It's correct that these `serialize` formats are not used in our JSON encoding, but they are used to build summaries.

thiemowmde: I uploaded a draft utilizing this approach as #718.

manicki: For what is worth: if we didn't want to use PHP serialization for EntityIdValue that @thiemowmde is suggesting in https://github.com/wmde/WikibaseDataModel/pull/718 (I am not sure what are implications with that approach, if this serialization is never stored) another option (theoretically) could be using `EntityIdComposer` which is meant as a entity-type-agnostic (kinda) replacement for `LegacyIdInterpreter`. The obvious problem is how to pull `EntityIdComposer` in here, so it makes sense (it is now bound to Wikibase git repo and entity type definitions out there). I cannot think of good way to achieve this, so this comment is mostly noise.

Also theoretically, another option would be to use value of EntityId::getSerialization in EntityIdValue::serialize, and use EntityIdParser in EntityIdValue::unserialize but that seems so wrong.

brightbyte: @JeroenDeDauw Yes, I realized this morning that I misunderstood. deserialize( $this->entityId ) will work, sure.

src/Entity/EntityIdValue.php
brightbyte: If the class doesn't exist, this should fail hard. Ok, we could throw an exception instead of a fatal error.

brightbyte:
I talked to Thiemo about this. We decided to use his patch #718 for EntityIdValue and the EntityId classes, and my changes for SnakObject and PropertyValueSnak. I will remove the changes to EntityIdValue from this pull request to avoid conflicts. The two pull requests do not depend on each other, but are conceptually related, and should become part of the same release. 

Note that this is a breaking change only if we consider the php serialization to be a stable external interface. I don't think we have an explicite statement about this anywhere.

src/Entity/EntityIdValue.php
brightbyte: @thiemowmde you wrote:
>  they are used to build summaries.

when/where?  I can't find it.

brightbyte:
NOTE: this changes snak hashes, and consequently reference hashes! This may break reference update/removal!

src/Snak/SnakObject.php
brightbyte: Still want me to remove this? This is currently blocking the pull  request...

src/Snak/SnakObject.php
```diff
@@ -102,7 +102,7 @@ public function equals( $target ) {
 	 * @return string
 	 */
 	public function serialize() {
-		return serialize( $this->propertyId->getNumericId() );
+		return serialize( $this->propertyId->getSerialization() );
```
thiemowmde: This is a bit pointless. Please replace with a straight `return $this->propertyId->getSerialization();`.

src/Snak/SnakObject.php
```diff
@@ -113,7 +113,22 @@ public function serialize() {
...
+		$this->propertyId = self::newPropertyId( unserialize( $serialized ) );
+	}
+
+	/**
+	 * @param mixed $unserialized
+	 *
+	 * @return PropertyId
+	 */
+	protected static function newPropertyId( $unserialized ) {
+		if ( is_int( $unserialized ) ) {
```
thiemowmde: This also needs redoing with the change above, and should probably be inlined above:
```php
$numericId = unserialize( $serialized );
$this->propertyId = is_int( $numericId )
    ? PropertyId::newFromNumber( $numericId )
    : new PropertyId( $serialized );
```

brightbyte:
NOTE: once this is merged, we should make a release.

src/Snak/SnakObject.php
```diff
@@ -113,7 +113,34 @@ public function serialize() {
 	 * @param string $serialized
 	 */
 	public function unserialize( $serialized ) {
-		$this->propertyId = PropertyId::newFromNumber( unserialize( $serialized ) );
+		$this->propertyId = self::newPropertyId( $serialized );
+	}
+
+	/**
+	 * @param string $serialized
```
thiemowmde: string|int

src/Snak/PropertyValueSnak.php
```diff
@@ -64,8 +64,8 @@ public function serialize() {
 	 * @param string $serialized
 	 */
 	public function unserialize( $serialized ) {
-		list( $numericId, $dataValue ) = unserialize( $serialized );
-		$this->__construct( $numericId, $dataValue );
+		list( $propertyId, $dataValue ) = unserialize( $serialized );
+		$this->__construct( self::newPropertyId( $propertyId ), $dataValue );
```
thiemowmde: Mark the date in your calendar. ;-) This might be the first time I don't like the fact that this relies on a protected method from a base class. Can be much simpler:
```php
$this->__construct( is_int( $propertyId ) ? $propertyId : new PropertyId( $propertyId ), $dataValue );
```

src/Snak/SnakObject.php
```diff
@@ -113,7 +113,34 @@ public function serialize() {
...
+
+		try {
+			// full ID, as a string, not serialized
+			return new PropertyId( $serialized );
+		} catch ( InvalidArgumentException $ex ) {
+			// noop
+		}
+
+		$unserialized = unserialize( $serialized );
+		if ( is_int( $unserialized ) ) {
```
thiemowmde: Sorry, but this code is weird. It does an `is_int` check twice? The reason is that this method is used for two very different use cases:
* In Value snaks this can be called with either `1` or `"P1"`.
* In NoValue and SomeValue snaks this can be called with `"i:42;"` or `"P1"`.

The problem is: there is no need to make these two use cases compatible with each other.

brightbyte:
Thanks @thiemowmde  for refactoring the PropertyId instantiation! +1 for that.

JeroenDeDauw:
+1

tests/unit/Snak/PropertyValueSnakTest.php
```diff
@@ -121,15 +121,62 @@ public function notEqualsProvider() {
...
-	public function testUnserialize() {
+	public function provideDataToUnserialize() {
+		$p2 = new PropertyId( 'P2' );
+		$p2foo = new PropertyId( 'foo:P2' );
+		$value = new StringValue( 'b' );
+
+		return [
+			'legacy' => [
+				new PropertyValueSnak( $p2, $value ),
+				'a:2:{i:0;i:2;i:1;C:22:"DataValues\StringValue":1:{b}}'
```
manicki: It does not affect test result but I believe to be 100% accurate "i:2" here should become "i:1". Minor thing, I am going to fix this, and squash into other commits.

src/Snak/SnakObject.php
```diff
@@ -97,12 +97,12 @@ public function equals( $target ) {
 	/**
 	 * @see Serializable::serialize
 	 *
-	 * @since 0.1
+	 * @since 7.0
```
manicki: I wonder is this correct to change this number here. Change here is indeed breaking change but the interface of the methods does not change. So I guess version number change is to show that serialization format has changed in the release 7.0. I am find with that I just wonder if this is "how it should be done" (tm).

tests/unit/Snak/PropertyValueSnakTest.php
manicki: Please ignore this comment. I looked at wrong spot in the old code. All good here.

src/Snak/SnakObject.php
thiemowmde: I believe the returned value is a crucial part of the contract of a method. If the format of the return value changes, it's a different method. The idea of the since tag is to make this obvious.

JeroenDeDauw: We have some places where we bumped the number but then mention the method was already there in some other form or something along those lines

manicki: Yea, something like this seems reaonable.

PR #717, opened 2017-02-09
Avoid calling getFingerprint where not necesarry

thiemowmde:


JeroenDeDauw:
+1

tests/unit/Entity/PropertyTest.php
```diff
@@ -27,16 +27,18 @@
...
 class PropertyTest extends PHPUnit_Framework_TestCase {
 
+	/**
+	 * @return Property
+	 */
 	private function getNewEmpty() {
 		return Property::newFromType( 'string' );
 	}
 
 	public function testConstructorWithAllParameters() {
```
manicki: Dropping the last argument (new StatementList()) as this defaults to null is OK I believe. But then the name of this test is no longer accurate, no?

thiemowmde: Good catch! Will fix.

PR #718, opened 2017-02-09
Rework native EntityId(Value) serializations

thiemowmde:
This is an alternative draft for #716. Pinging @brightbyte and @JeroenDeDauw.

This patch does 2 things:
1. Change EntityIdValue::serialize to use the native `serialize` instead of JSON.
2. Simplify both ItemId and PropertyId serializations to the straight string.

The 1st change makes the serialized string much longer, the 2nd change minimizes it again.

In all cases I keep compatibility with the previous serialization formats.

### DIFF
--- a/tests/unit/Entity/ItemIdTest.php
+++ b/tests/unit/Entity/ItemIdTest.php
@@ -92,7 +92,7 @@ class ItemIdTest extends PHPUnit_Framework_TestCase {
 
 	public function testSerialize() {
 		$id = new ItemId( 'Q1' );
-		$this->assertSame( '["item","Q1"]', $id->serialize() );
+		$this->assertSame( 'Q1', $id->serialize() );
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


tests/unit/Entity/ItemIdTest.php
```diff
@@ -114,7 +114,7 @@ public function serializationProvider() {
 			[ '["",""]', '' ],
 			[ '["",2]', 2 ],
 			[ '["",null]', null ],
-			[ '', null ],
+			[ '', '' ],
```
brightbyte: Should test new serialization format: [ 'Q2', 'Q2' ],

tests/unit/Entity/PropertyIdTest.php
```diff
@@ -92,7 +92,7 @@ public function testGetEntityType() {
 
 	public function testSerialize() {
```
brightbyte: there is no test for Propertyid::unserialize?!

thiemowmde: There is, see https://github.com/wmde/WikibaseDataModel/pull/718/files/8a3e451d87ae29420ff85354f328f80ba1e72363#diff-4d5b4f754035c1d318ef4e6f8c2382a2L101.

tests/unit/Entity/ItemIdTest.php
thiemowmde: Done for both ItemId and PropertyId.

PR #719, opened 2017-02-13
Release 7.0.0

thiemowmde:
Please have a look at the milestone before merging this: https://github.com/wmde/WikibaseDataModel/milestone/20

[Bug: T157965](https://phabricator.wikimedia.org/T157965)

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,21 +1,52 @@
 # Wikibase DataModel release notes
 
-## Version 7.0.0 (dev)
+## Version 7.0.0 (2017-03-15)
 
-* Removed `FingerprintHolder`
+This release adds support for custom entity types to `EntityIdValue`, and thus changes the hashes of
+snaks, qualifiers, and references.
+
+* Changed the internal `serialize()` format of several `EntityId` related classes. In all cases
+  `unserialize()` still supports the previous format.
+	* Serialization of `SnakObject` (includes `PropertyNoValueSnak` and `PropertySomeValueSnak`)
+	  does not use numeric IDs any more
+	* Serialization of `PropertyValueSnak` (includes `DerivedPropertyValueSnak`) does not use
+	  numeric IDs any more
+	* Serialization of `EntityIdValue` does not use numeric IDs any more
+	* `EntityIdValue` can now serialize and unserialize `EntityId`s other than `ItemId` and
+	  `PropertyId`
+	* Minimized serialization of `ItemId` and `PropertyId` to not include the entity type any more
+
+#### Other breaking changes
+
+* Removed `FingerprintHolder`. Use `TermList::clear` and `AliasGroupList::clear` instead. `Item` and
+  `Property` also still implement `setFingerprint`.
 * Removed class aliases deprecated since 3.0:
 	* `Wikibase\DataModel\Claim\Claim`
 	* `Wikibase\DataModel\Claim\ClaimGuid`
 	* `Wikibase\DataModel\StatementListProvider`
+* Added a `SnakList` constructor that is not compatible with the `ArrayList` constructor any more,
+  and does not accept null any more.
+* Removed `HashArray::equals`, and `HashArray` does not implement `Comparable` any more
+* Removed `HashArray::getHash`, and `HashArray` does not implement `Hashable` any more
 * Removed `HashArray::rebuildIndices`
 * Removed `HashArray::indicesAreUpToDate`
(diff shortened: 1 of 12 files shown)


thiemowmde:
FYI, I just rebased this and made sure it really contains all changes that are currently in master. We talked about this, and indeed: master is in a state we can use for a 7.0 release. The pressing changes are all merged. The unmerged stuff in the 7.0 milestone is optional. I would still love to merge as much as we can agree on, but we can take stuff out.

Let's try to do a release on Tuesday or at least Wednesday.

thiemowmde:
I rebased this again after #649 was merged. I also rearranged [the 7.0.0 milestone](https://github.com/wmde/WikibaseDataModel/milestone/20) an other time. No patch in this list should block the release, but it would still be sooo nice to have as much of them merged as possible.

JeroenDeDauw:
So what do I do to make a grammar fix now? If changes where submitted as dedicated PRs as soon as they are called for, I could just have merged the relevant stuff and made a follow up PR. Things go a lot faster like that if you ask me.

RELEASE-NOTES.md
```diff
@@ -1,8 +1,24 @@
 # Wikibase DataModel release notes
 
-## Version 7.0.0 (dev)
-
-* Removed `FingerprintHolder`
+## Version 7.0.0 (2017-03-07)
+
+This release allows `EntityIdValue` to support custom entity types, and needs to change snak,
+qualifier, and reference hashes because of this.
```
JeroenDeDauw: This release adds support for custom entity types to `EntityIdValue`, and thus to changes the hashes of snaks, qualifiers, and references.

RELEASE-NOTES.md
```diff
@@ -11,11 +27,12 @@
 * Removed `HashArray::indicesAreUpToDate`
 * Removed `HashArray::removeDuplicates`
 * Removed `$acceptDuplicates` feature from `HashArray`
-* Added `clear` to `TermList`, `AliasGroupList` and `StatementList`
+* Fixed exceptions in `DispatchingEntityIdParser` and `ItemIdParser` not forwarding the previous
+  exception
```
JeroenDeDauw: Please separate breaking changes from non-breaking changes as was done in the past

RELEASE-NOTES.md
thiemowmde: I believe "thus to" was a mistake. Besides this is indeed better, and I used it. Thanks.

RELEASE-NOTES.md
thiemowmde: Done.

thiemowmde:
> So what do I do to make a grammar fix now?

You can ask for changes just as you did, or amend this patch by using the edit button in the top right corner of each file. This will create a commit that will be part of this pull request.

JeroenDeDauw:
Asking for a change is slower, it takes more total effort, and is less collaborative IMO. Hence I prefer to make an actual proposal of this change. I could directly modify this PR as you suggest, though this leads to complications, esp when the author of the PR does not agree with the change (which did happen in the past and lead to hurt feelings).

Anyway, this is just a general frustration I have with some PRs lately. It's not specific to this one, and I don't think it really matters for any single one, so no changes requested here from my side.

thiemowmde:
What you are proposing is, from what I understand, that changes to master should be made faster, possibly even skipping review if the author feels like a change is not critical. In case somebody disagrees he can make a pull request any time.

I believe I missed a point or just misunderstood that, but if this is the proposal I need to point out a problem: This creates the situation where one dev (or a small group) will make changes to a code base with almost no discussion (mostly because there is no time and no chance to react), while other devs that disagree are enforced to propose reverts or renames as pull requests that can only be merged when everybody agrees. While this discussion is ongoing master is already in a state that reflects the opinion of the first group. They basically "won", while the others must proof them "wrong". Personally I find this situation more frustrating than anything else.

Commenting, or amending a patch, or chaining pull requests all avoid this situation. This is possible in both Gerrit and GitHub.

manicki:
@thiemowmde: it doesn't seem I am going to have few minutes to look at stuff you put into the 7.0.0 milestone any time soon. I don't think those changes are of high priority and we have more urgent stuff to do. That said how about we tag 7.0.0 with what we have now? If you say it's fine let's just change the date in the release notes, and I'll merge this and tag the release.

thiemowmde:
Can we merge at least #654? It's not just you who should feel responsible for this component. The whole team should. @bekh6ex, can you please have a look?

PR #720, opened 2017-02-15
Add ItemId and PropertyId::newFromRepositoryAndNumber

thiemowmde:
A convenience method that allows us to avoid constructing ugly code (see https://gerrit.wikimedia.org/r/337806).

### DIFF
--- a/src/Entity/ItemId.php
+++ b/src/Entity/ItemId.php
@@ -103,4 +103,21 @@ class ItemId extends EntityId implements Int32EntityId {
 		return new self( 'Q' . $numericId );
 	}
 
+	/**
+	 * CAUTION: Use the full string serialization whenever you can and avoid using numeric IDs.
+	 *
+	 * @param string $repositoryName
+	 * @param int|float|string $numericId
+	 *
+	 * @return self
+	 * @throws InvalidArgumentException
+	 */
+	public static function newFromRepositoryAndNumber( $repositoryName, $numericId ) {
+		if ( !is_numeric( $numericId ) ) {
+			throw new InvalidArgumentException( '$numericId must be numeric' );
+		}
+
+		return new self( self::joinSerialization( [ $repositoryName, '', 'Q' . $numericId ] ) );
+	}
+
 }
--- a/src/Entity/PropertyId.php
+++ b/src/Entity/PropertyId.php
@@ -103,4 +103,21 @@ class PropertyId extends EntityId implements Int32EntityId {
 		return new self( 'P' . $numericId );
 	}
 
+	/**
+	 * CAUTION: Use the full string serialization whenever you can and avoid using numeric IDs.
+	 *
+	 * @param string $repositoryName
+	 * @param int|float|string $numericId
+	 *
+	 * @return self
+	 * @throws InvalidArgumentException
+	 */
+	public static function newFromRepositoryAndNumber( $repositoryName, $numericId ) {
+		if ( !is_numeric( $numericId ) ) {
+			throw new InvalidArgumentException( '$numericId must be numeric' );
+		}
+
+		return new self( self::joinSerialization( [ $repositoryName, '', 'P' . $numericId ] ) );
+	}
+
 }
--- a/tests/unit/Entity/ItemIdTest.php
+++ b/tests/unit/Entity/ItemIdTest.php
@@ -156,4 +156,14 @@ class ItemIdTest extends PHPUnit_Framework_TestCase {
 		];
 	}
 
+	public function testNewFromRepositoryAndNumber() {
+		$id = ItemId::newFromRepositoryAndNumber( 'foo', 1 );
+		$this->assertSame( 'foo:Q1', $id->getSerialization() );
+	}
+
+	public function testNewFromRepositoryAndNumberWithInvalidNumericId() {
+		$this->setExpectedException( InvalidArgumentException::class );
+		ItemId::newFromRepositoryAndNumber( '', 'Q1' );
+	}
+
 }
--- a/tests/unit/Entity/PropertyIdTest.php
+++ b/tests/unit/Entity/PropertyIdTest.php
@@ -156,4 +156,14 @@ class PropertyIdTest extends PHPUnit_Framework_TestCase {
 		];
 	}
 
+	public function testNewFromRepositoryAndNumber() {
+		$id = PropertyId::newFromRepositoryAndNumber( 'foo', 1 );
+		$this->assertSame( 'foo:P1', $id->getSerialization() );
+	}
+
+	public function testNewFromRepositoryAndNumberWithInvalidNumericId() {
+		$this->setExpectedException( InvalidArgumentException::class );
+		PropertyId::newFromRepositoryAndNumber( '', 'P1' );
+	}
+
 }


manicki:
As I was trying to say in https://gerrit.wikimedia.org/r/#/c/337806/ I am not convinced this is something we want to have here. I might not be expressing myself clear enough. I don't know if it makes any sense for @thiemowmde that we re-phrase arguments we have in yet another way, though. Thus I am stepping back and would like to hear from others.

But to the particular change here, I don't why I didn't make this comment before:
I think the name of the new constructor is misleading. One could obviously (and if you want to use it in things like PropertyInfoTable this will be happening on wiki with properties all the time) pass an empty string (which is currently interpretted as "a local repository") as $repostioryName. How this ID is foreign then?
If we wanted to have a "newFromNumber" constructor allowing to pass in the repository name, it could be also added to the old "newFromNumber". For keeping backwards compatibility it would probably come as a last parameter with default empty-string value. This will make it look a bit the wrong way round. My memory is not very good but I am pretty sure this has been discussed already some months ago.

That's from me on this. I will be sending individual-targeted pings if nothing happens here. You've been warned.

src/Entity/ItemId.php
```diff
@@ -103,4 +103,21 @@ public static function newFromNumber( $numericId ) {
...
+	/**
+	 * CAUTION: Use the full string serialization whenever you can and avoid using numeric IDs.
+	 *
+	 * @param string $repositoryName
+	 * @param int|float|string $numericId
+	 *
+	 * @return self
+	 * @throws InvalidArgumentException
+	 */
+	public static function newForeignFromNumber( $repositoryName, $numericId ) {
```
filbertkm: this is not just "...FromNumber" (repository name + a number)  though can't think of a better name right now 

with "...FromNumber", I somewhat would think the $numericId would be the first parameter (the most important thing)

thiemowmde: I feel the name "**foreign** from number" hints that the "foreign" bit comes first. This is also consistent with the static `EntityId::…Serialization` methods.

thiemowmde:
> As I was trying to say […]

Even if I created https://gerrit.wikimedia.org/r/337806 and this patch the same time, they are not depending on each other in any way. All I do here is adding convenience functions for code that already exists. This does not introduce any new concept.

> If we wanted to have a "newFromNumber" constructor allowing to pass in the repository name, it could be also added to the old "newFromNumber".

Personally I do not like this *because* of the wrong parameter order. It does not make anything better or less complex: there are still two possible ways to construct an ItemId from a number, one with and one without a repository prefix.

I understand the argument that `newForeignFromNumber` will create local IDs when the repository prefix is an empty string. It's just that this or something similar is true for pretty much all our code, and I strongly disagree with the conclusion that nothing can have the word "foreign" in it because of this. I find this argument hypercritical. We could only avoid such discussions by picking the most generic name. That would be a pity.

What about `newFromRepositoryAndNumber`?

manicki:
> Even if I created https://gerrit.wikimedia.org/r/337806 and this patch the same time, they are not depending on each other in any way. All I do here is adding convenience functions for code that already exists. 

Why do we need such convenience function then? Where is it intended to be used? Tests?

> I understand the argument that newForeignFromNumber will create local IDs when the repository prefix is an empty string. It's just that this or something similar is true for pretty much all our code, and I strongly disagree with the conclusion that nothing can have the word "foreign" in it because of this. I find this argument hypercritical. We could only avoid such discussions by picking the most generic name. That would be a pity.

I don't quite get this point.
 - In my opinion notion "foreign id" can only make sense when we say there are also "local ids". Nothing is wrong with having such a distinction. 
 - Said we have "foreign vs local" distinction and then have a method with a name stating it creates "foreign" objects while what it can be creating both "foreign" and "local" objects, I argue that the name of the method is not accurate. There is no problem itself in having the method that does create both kind of objects (maybe we need something like it). Just the name is the problem as it suggests something that is not true.
 - I didn't really get what you meant by "It's just that this or something similar is true for pretty much all our code, and I strongly disagree with the conclusion that nothing can have the word "foreign" in it because of this. I find this argument hypercritical." and how it relates to the case here.
 - I don't think names of methods, classes, etc should be "generic". And I don't think I said anything like this above. But I do think names should be accurate.

"newFromRepositoryAndNumber" is an accurate name, yes.

thiemowmde:
> Why do we need such convenience function […]

I find this question surprising. We do not need it, but it's convenient to have something like this (somewhere, does not need to be here) in all places that construct an EntityId from a known entity type, repository prefix, and number.

> […] the name is the problem as it suggests something that is not true.

How is the fact that `newForeignFromNumber( 'foo', 1 )` does create a foreign EntityId "not true"?

About the argument being hypercritical: A good example is `EntityHandler::makeEntityContent`. The fact that this returns an ItemContent does not mean that the name "makeEntityContent" is wrong.

thiemowmde:
I changed the name to get this unstuck. I still disagree with the idea that `newForeignFromNumber` is somehow "bad". It makes the *intent* so much more obvious. It should only be used in contexts with variable repository names, and not like `newFromRepositoryAndNumber( '', $numericId )`.

manicki:
> How is the fact that newForeignFromNumber( 'foo', 1 ) does create a foreign EntityId "not true"?

It is not "not true". I wasn't referring to that.
What is problematic is the following:
 - I call ItemId::newForeignFromNumber( '', 1 ), and I get ItemId $x
 - on this $x I do $x->isForeign() and it returns false
 - with such result I figure either a) there is a bug in "newForeignFromNumber", or b) method works correct it was just named wrong.

> A good example is EntityHandler::makeEntityContent. The fact that this returns an ItemContent does not mean that the name "makeEntityContent" is wrong.

ItemContent is a EntityContent so I don't see contradiction here. But this is also completely different case.
If a fictive EntityHandler::makeNonItemContent returned ItemContent it would be a similar case to what I was referring to here.

What I wanted to say from the beginning is that as we've introduced the exclusive distinction between "foreign" and "non-foreign" entity ids (one could argue if it makes sense but it is different discussion), and there are places in Wikibase that rely on this distinction, as far as I can tell, it is kind of critical that we don't create false assumptions and guarantees in the code. Having "newForeignFromNumber" that with no problem returns a non-foreign id is such a false guarantee IMO.

Really, I didn't criticise for the sake of criticism. I wish I explained my points more clear so we don't use resources we're short on no some better things than discussing names.

thiemowmde:
> […] this is also completely different case.

I almost expected you to say this. But it's not.

Being "foreign" is not a specialization, it's a **generalization**. Not all foreign EntityIds are local but all local EntityIds are foreign.

manicki:
Right, I've finally understood that this all fuss is because of us having different understanding of concepts we're referring to. I wasn't aware of this until now. Knowing that, this is my last comment here, I have nothing more to say here.

In all my previous comments when referring to "newForeignFromNumber" being a problematic name for method with a given implementation I was assuming that:
1. ID of every entity always contains the name of the repository the entity belongs to (from docs/foreign-entity-ids.wiki: "EntityId contains the logical name of the repository the ID refers to").
2. Local entity IDs have an empty string as a repository name part of the ID (ibid, possibly not very explict: "The empty repository name "" refers to the local wiki")m
3. Foreign entity IDs have a non-empty string value as a repository name part of the ID (not explicitly said in the documentation, in my opinion it is the only interpretation foreign-entity-id provides as a whole).

Based on 1, 2, 3 in my understanding the following statements are axioms:
4. Every entity ID is either local or foreign.
5. Entity ID is never both local and foreign.

4 and 5 are clearly contradicting your understanding of those two concepts, especially
> Being "foreign" is not a specialization, it's a generalization. 

and
> all local EntityIds are foreign.

If my understanding of concepts "local entity ID" and "foreign entity ID" was wrong, I can only apologize for distracting you and others who have to dig through this whole discussion, and for not making myself clear early enough. In such case I feel I need to state that I find the situation where we ended up creating ourselves a language where "local" is a subset of "foreign" very unfortunate.

thiemowmde:
I believe this question is the critical one: Do we have a use case where we want code to work with "foreign" entity IDs, but **fail** for entity IDs with the repository set to an empty string? If so, where, and why?

Or, in other words: why exactly are "local" and "foreign" required to be *exclusive*?

I believe this just can not work out in the long run. These concepts can not be exclusive. To make them exclusive, **all** code consuming EntityId objects **must** check `isLocal` before doing anything with the object. But this can not be guaranteed, obviously. There is no type hint against `LocalEntityId`, nor for the opposite. **All** code **must** assume that each EntityId object it gets may have whatever repository prefix – which makes this the generalization.

We can call the generalized concept `RepositoryAwareEntityId` or something like this, to avoid the disputed term "foreign".

thiemowmde:
See #722.

Note that the discussion should not block this patch. I agree that we should avoid the term "foreign". I'm perfectly fine with the current name `newFromRepositoryAndNumber`. If you agree then please merge this.

PR #722, opened 2017-03-02
[DNM] Replace isForeign with isUnresolved

thiemowmde:
Triggered by the discussion in #720.

It appears we implemented a concept of being "foreign" like it is a specialization. Multiple issues with that:
* All we really know is if an entity ID does have prefixes.
  * If there is no prefix, we know the ID refers to an entity in the local repository.
  * But we can't conclude the opposite. A prefixed ID may still refer to a local entity, either if it's prefixed with the local repository name, or with multiple prefixes that happen to point back to the local repository. The ID of such an entity may be in a "foreign" form, but if the addressed entity is local it's confusing to still describe this concept as being "foreign".
* All entity IDs are prefixed. (In a sense all are "foreign".) It's just that there is special case handling for IDs that happen to have no prefix.
* I don't think we have code that needs the guarantee that an addressed entity is "foreign" (which is a guarantee we can not give anyway, as explained above). But we have code that needs the guarantee that an addressed entity is local, typically when we are going to store it.

### DIFF
--- a/src/Entity/EntityId.php
+++ b/src/Entity/EntityId.php
@@ -58,7 +58,8 @@ abstract class EntityId implements Comparable, Serializable {
 	 * Returns an array with 3 elements: the foreign repository name as the first element, the local ID as the last
 	 * element and everything that is in between as the second element.
 	 *
-	 * EntityId::joinSerialization can be used to restore the original serialization from the parts returned.
+	 * self::joinSerialization can be used to restore the original serialization from the parts
+	 * returned.
 	 *
 	 * @since 6.2
 	 *
@@ -81,7 +82,7 @@ abstract class EntityId implements Comparable, Serializable {
 	}
 
 	/**
-	 * Builds an ID serialization from the parts returned by EntityId::splitSerialization.
+	 * Builds an ID serialization from the parts returned by self::splitSerialization.
 	 *
 	 * @since 6.2
 	 *
@@ -104,8 +105,8 @@ abstract class EntityId implements Comparable, Serializable {
 	}
 
 	/**
-	 * Returns '' for local IDs and the foreign repository name for foreign IDs. For chained IDs (e.g. foo:bar:Q42) it
-	 * will return only the first part.
+	 * Returns an empty string for unprefixed, or a repository name for prefixed IDs. For chained
+	 * prefixes (e.g. foo:bar:Q42) it will return only the first prefix.
 	 *
 	 * @since 6.2
 	 *
@@ -131,14 +132,18 @@ abstract class EntityId implements Comparable, Serializable {
 	}
 
 	/**
-	 * Returns true iff EntityId::getRepoName returns a non-empty string.
+	 * Returns true if the ID does have a prefix and is therefore guaranteed to be unresolved. This
+	 * is the same as checking if getRepositoryName() returns a non-empty string. Note that the
+	 * inverse is not necessarily true! An unprefixed ID may still need to be resolved, e.g. when
+	 * the local repository does not support a specific entity type, but a known remote repository
+	 * does.
 	 *
-	 * @since 6.2
+	 * @since 7.0
 	 *
 	 * @return bool
 	 */
-	public function isForeign() {
-		// not actually using EntityId::getRepoName for performance reasons
+	public function isUnresolved() {
+		// Not actually using getRepositoryName() for performance reasons
 		return strpos( $this->serialization, ':' ) > 0;
 	}
 
--- a/tests/unit/Entity/EntityIdTest.php
+++ b/tests/unit/Entity/EntityIdTest.php
@@ -85,12 +85,12 @@ class EntityIdTest extends \PHPUnit_Framework_TestCase {
 		$this->assertInternalType( 'string', $id->__toString() );
 	}
 
-	public function testIsForeign() {
-		$this->assertFalse( ( new ItemId( 'Q42' ) )->isForeign() );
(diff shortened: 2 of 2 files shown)


manicki:
While I think flipping isForeign to isLocal might be a thing we want to do (I agree we have use cases where we only want to have local entities, and I am not aware of use case where we only need foreign ones) I am concerned that there is some misunderstanding around the concepts of "foreign", "local", and "prefix" which is reflected in some (secondary) changes in this PR. I strongly believe this should be sorted out at this point, to avoid confusion in the future. With the experience from the other PR, I don't feel like trying to discuss it in comments. I'd rather let @brightbyte speak, and/or discuss it IRL and write up conclusions afterwards.

thiemowmde:
On hold for now. Should not block a release. However, I'm not closing this because I still think we should aim for a better name than `isForeign`. In todays discussion things like "hasNonEmptyRepositoryPrefix" or "isUnresolved" came up. But note that "isResolved" will not work for the same reason as "isForeign", "isLocal", or "exists": all these claim to know something about the entity it can not know.
* "d:Q1" looks "foreign", and "isForeign" will return true because of this, but it may as well be a local entity.
* "Q1" looks "local" and "resolved", and both "isLocal" and "isResolved" will return true because of this, but it may still address an item on a remote repository and need to be resolved because of this.

For the moment, and considering the actual use cases for this, `isUnresolved` feels best.

brightbyte:
@thiemowmde wrote:

> "Q1" looks "local" and "resolved", and both "isLocal" and "isResolved" will return true because of this, but it may still address an item on a remote repository and need to be resolved.

That's true for the string "Q1", it should never be true for an ItemId( "Q1" ). If items are not on the local repo, the string "Q1" would turn into "whatever:Q1" during parsing.

A further complication: in client code (with repo disabled), "local" means "on the default repo", bot "on this wiki".

thiemowmde:
> That's true for the string "Q1", it should never be true for an ItemId( "Q1" ).

You said that before, and I strongly disagree with this argument. `ItemId( 'Q1' )` can not give any more guarantees than "Q1".

brightbyte:
@thiemowmde said:
> You said that before, and I strongly disagree with this argument. ItemId( 'Q1' ) can not give any more guarantees than "Q1".

Sure it can. ItemId can guarantee that having an empty prefix (aka repo name) means the ID is local (that is, it refers to an entity on the default repo). This will be true for all wikis, regardless of configuration. It's part of the contract of ItemId (and EntityId).

The string "Q1" is local on Wikidata, where it becomes ItemId( 'Q1' ). It's currently also "local" on Wikipedia and Commons. But once Commons is a repo itself, we MUST NOT parse "Q1" as ItemId( 'Q1' ), the parser must turn it into ItemId( 'wikidata:Q1' ). ItemId( 'Q1' ) simply be invalid on Commons.

thiemowmde:
> ItemId can guarantee that having an empty prefix […] means the ID is local […]

If "Q1" may as well refer to an item on a remote repository, then how on earth can the constructor of *any* class give a guarantee that goes way beyond what the code actually knows? Such a guarantee does not appear out of thin air by just adding a "do not use" section to the classes documentation.

> once Commons is a repo itself, we MUST NOT parse "Q1" as ItemId( 'Q1' ), the parser must turn it into ItemId( 'wikidata:Q1' ). ItemId( 'Q1' ) simply be invalid on Commons.

I'm able to understand this, but it's entirely irrelevant for the discussion here. This here is the DataModel component. It does not know anything about Commons. It does not know anything about the EntityIdFactory we are going to use. The guarantee you are talking about is in code the DataModel does not know about.

It appears we need to have a meeting about decoupling and component design.

brightbyte:
@thiemowmde wrote
> It appears we need to have a meeting about decoupling and component design.

Yea, let's do this. Because I think it's entirely in the domain of this component to specify the meaning and intended usage of the value object it defines. Saying "empty string means the local/default repo" needs no knowledge about parsers and factories. It's the other way around: we write these parsers and factories to conform to the contract of EntityId.

I do agree that the contract of EntityId in general and the getRepositoryName and isLocal methods should be made more clear in the documentation. Currently, their intended semantics isn't really documented. My point is that they can and should be specified here, on the EntityId class.

Perhaps the best place would be the class level documentation of EntityId. That could specify the syntax, and the interpretation and intended use of the parts of the ID.

thiemowmde:
We established that a string like "wikidata:Q1" may address a foreign as well as a local entity.

We established that a string like "Q1" may address a local as well as a foreign entity.

We established that `new ItemId( 'Q1' )` can not encode more knowledge than the string "Q1".

The only knowledge in this component is if there is a prefix or not. The wisdom of what this means is in code outside of this component. I really don't understand why we are still arguing about this simple fact, and it already made me angry yesterday. Please just stop.

Naming a method in this component "isForeign" is misleading, confusing and just wrong. This component can not know this.

No amount of "you should not" documentation in this component will change this.

brightbyte:
> I really don't understand why we are still arguing about this simple fact, and it already made me angry yesterday. Please just stop.

We are still arguing about it because we still disagree. Whether that makes you angry or not is besides the point, and entirely your own business.

We can bikeshed about names, as long as semantics is well-defined. There is nothing worse than underspecified data models.

thiemowmde:
Did you just said you disagree with the fact that there is no service in this component that resolves repository prefixes and entity types and knows if "Q1" is "local" or "foreign"?

JeroenDeDauw:
Please reopen if this should still show up in the open PR list

PR #723, opened 2017-03-06
Improve documentation of EntityId

brightbyte:
This patch documents the intended semantics of the methods in the
EntityId base class, to align them to the behavior defined in
docs/foreign-entity-ids.wiki

### DIFF
--- a/docs/foreign-entity-ids.wiki
+++ b/docs/foreign-entity-ids.wiki
@@ -1,10 +1,10 @@
 EntityId contains the logical name of the repository the ID refers to. This serves as a namespace mechanism for entity IDs.
-EntityIds from repositories other than local repository are called '''foreign''' EntityIds,
-and, similarly, repositories other than local are '''foreign''' repositories.
+EntityIds from repositories other than the local default repository are called '''foreign''' EntityIds,
+and, similarly, repositories other than the local default are '''foreign''' repositories.
 
 Notes:
 * The name used to refer to a given repository is local, and can differ from client to client.
-* The empty repository name "" refers to the local wiki.
+* The empty repository name "" refers to the local application's default repository. If the application is a Wikibase repo, "" refers to this repository itself. If the application is a Wikibase client, it refers to that client's default repository.
 * Repository names can be used to look up the repository Site object, which in turn allows the ID to be mapped to the correct URI, URL, or API path for the repository it belongs to.
 * The serialization of the EntityId follows the pattern <code><repository>:<id></code>. If <code><repository></code> is empty, the leading <code>:</code> is optional. This format follows the convention set by XML namespaces, RDF prefixes, and MediaWiki interwiki links.
 * Repository names can be mapped during serialization and deserialization.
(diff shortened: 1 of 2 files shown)


brightbyte:
Ping @thiemowmde @manicki @filbertkm

docs/foreign-entity-ids.wiki
```diff
@@ -17,7 +17,7 @@ When receiving prefixed entities from another repository, prefixes are "chained"
 
 * When reading <code>d:Q5</code> from repository <code>foo</code>, it is turned into <code>foo:d:Q5</code>, meaning ''<code>Q5</code> at the repo that repo <code>foo</code> calls <code>d</code>''
 * The local repository may have mappings defined for the prefixes used by other wikis, e.g. <code>foo:d</code> would be known to be the same as the local prefix <code>wd</code> for Wikidata. Mappings are defined as a two-dimensional array, e.g. part of mapping definition for example mentioned here would look like below:
-<code> [ ... 'foo' => [ 'd' => 'wd' ], ... ]</code>
+<code> array( ... 'foo' => array( 'd' => 'wd' ), ... )</code>
```
manicki: Is this change intended?

src/Entity/EntityId.php
```diff
@@ -43,12 +85,16 @@ private static function assertValidSerialization( $serialization ) {
 	}
 
 	/**
-	 * @return string
+	 * @return string The type of the designated Entity.
```
manicki: This description could be omitted I believe.

src/Entity/EntityId.php
```diff
@@ -7,6 +7,38 @@
...
+ * the EntityId object was instantiated in, or, for Wikibase  clients,
+ * the default repo of the client the EntityId object was instantiated in.
+ *
+ * A repository can be specified in the string representation of an EntityId
+ * as a prefix, separated by a colon (":") from the local part of the ID:
+ * "<repo>:<local-id>". Multiple such prefixes can be "chained", as in
+ * "foo:bar:rudd:X123"; in that case, only the first prefix ("foo") is
+ *  considered the repository name, the remaining prefixes are part of the
+ * "local part" of the ID ("bar:rudd:X123"), since they are to be resolved
+ * "locally" in the context of the context given as the first prefix.
```
manicki: "in the context of the **repository** given..."

src/Entity/EntityId.php
```diff
@@ -104,8 +150,18 @@ public static function joinSerialization( array $parts ) {
 	}
 
 	/**
-	 * Returns '' for local IDs and the foreign repository name for foreign IDs. For chained IDs (e.g. foo:bar:Q42) it
-	 * will return only the first part.
+	 * Returns the name of the repository which serves as the context for resolving this EntityId.
+	 * This method is the complement of getLocalPart().
+	 *
+	 * For local IDs with no repository name attached, getRespoitoryName() will return the empty
```
manicki: getRepositoryName mispelled slightly

src/Entity/EntityId.php
```diff
@@ -7,6 +7,38 @@
...
+ * which they can be resolved. IDs that do not provide a repository name
+ * (or, technically, have the repository name set to the empty string, '')
+ * are referred to as "local". IDs that do provide a repository name are
+ * referred to as "foreign" (even if they, by some detour, are equivalent to
+ * a local ID).
+ *
+ * The intended interpretation of "local" IDs is that they can be resolved
+ * in the default context - that is, they belong to the Wikibase repository
+ * the EntityId object was instantiated in, or, for Wikibase  clients,
+ * the default repo of the client the EntityId object was instantiated in.
```
thiemowmde: This is confusing and I believe even wrong. What this sentence basically says is that the interpretation of "local" is meaningless. Could mean anything you want. May not even be "local". Also it's unclear to what "instantiated" refers to if an ID is, for example, read from a database or dump. Same for "default". To what does this refer to? Who defines said "default"? Who knows about it? Certainly not the DataModel.

brightbyte:
> Furthermore, this component does not need this knowledge.

The data model needs to provide a specification for the interpretation of IDs. That specification needs to contain all information required by application code (including 3rd party clients) to successfully resolve an EntityId. 

Even if the code in this component does not need this knowledge, this component needs to define the respective contract, so that different users of the data model are interoperable.

brightbyte:
For the record, I don't insist on the terms "local" and "foreign". Maybe "relative" and "qualified" are better. I do maintain that it's important to clearly define the meaning of prefixes in general, and the empty prefix in particular, here in this component.

I agree that this component should not know about wikis, or their "location". But it needs to know about "repositories" as an abstract concept: a repo is a storage location, name resolution context, and identifier namespace. We need to make use of this concept in the specification of the data model. It's the only way to make sure all users of this component share the same interpretation of the data.

thiemowmde:
"Local" and "foreign" **are** locations.

In parallel I'm in an email thread with @brightbyte and it appears we have two different ideas of what this component represents. I believe this is the PHP implementation of Wikibase' base DataModel and not a conceptual data model that describes the interpretation of details like what an empty repository prefix in the serializations means.

JeroenDeDauw:
General observation here, not entirely specific to this PR: I'm not happy with foundational concepts being made more complicated. Needed to have long documentation on value objects is also kinda disconcerting.

JeroenDeDauw:
Please reopen if this should still show up in the open PR list

PR #727, opened 2017-03-15
Inline all abstract stuff from HashArrayTest to the implementations

thiemowmde:
All I do here is inlining the three methods from the abstract base class into the two implementations.

You will notice that the base class is almost empty now. A single helper method is left, and this is even used by only one of the two implementations. I'm aware of this, but I stopped working on the code at this point on purpose to make this patch as trivial as it can be. #702 will get rid of the base class along with the second implementation. There is just no point in making this code nicer before killing it.

This patch does the following things:
* It moves all relevant code to `SnakListTest`, making it more robust. This is the test we really care about.
* It removes a redundant test case from `SnakListTest`. Before, it was creating two instances with no arguments and `[]`. These are identical anyway.
* It also removes the second test case from `HashArrayWithoutDuplicatesTest`. It appears this was meant to test the "without duplicates" feature, and created an array with duplicates to do this. But there was never an assertion that tested this behaviour. What the `instanceProvider` returned always was an array with two equal instances. I just removed one now.

### DIFF
--- a/tests/unit/HashArray/HashArrayWithoutDuplicatesTest.php
+++ b/tests/unit/HashArray/HashArrayWithoutDuplicatesTest.php
@@ -4,6 +4,7 @@ namespace Wikibase\DataModel\Tests\HashArray;
 
 use Hashable;
 use Wikibase\DataModel\Fixtures\HashArrayElement;
+use Wikibase\DataModel\Fixtures\HashArrayWithoutDuplicates;
 use Wikibase\DataModel\HashArray;
 use Wikibase\DataModel\Snak\PropertyNoValueSnak;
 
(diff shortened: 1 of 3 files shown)


tests/unit/HashArray/HashArrayWithoutDuplicatesTest.php
```diff
@@ -19,17 +20,10 @@
  */
 class HashArrayWithoutDuplicatesTest extends HashArrayTest {
 
-	public function constructorProvider() {
-		$argLists = [];
-
-		$argLists[] = [ HashArrayElement::getInstances() ];
-		$argLists[] = [ array_merge( HashArrayElement::getInstances(), HashArrayElement::getInstances() ) ];
```
JeroenDeDauw: This one did not make sense?

thiemowmde: This is what I talked about in the commit message. These two constructors construct objects with the same elements. The second constructor call duplicates all elements, but because this is the "without duplicates" test, these duplicates are stripped on construction time. There are no assertions for this, so I came to the conclusion this second test case is redundant and can as well be removed.

JeroenDeDauw: Somehow did not see the body of the commit message... derp

thiemowmde:
Thanks a lot, @JeroenDeDauw! I rebased #702 again and find it really awesome now. If you want you can approve it, and I will merge it.

PR #728, opened 2017-03-16
Improve serialization support for custom ID types and foreign IDs.

brightbyte:
We need EntityIdValues to support plain old PHP serialization, so we can put objects that contain EntityIdValues into the object cache. 

This changes does two things:
* add tests that make sure we can apply PHP serialization and JSON serialization fr custom and foreign entity types
* Add support for JSON serialization of entity types IDs the not have a numeric representation.

Bug: https://phabricator.wikimedia.org/T160516

### DIFF
--- a/src/Entity/EntityIdValue.php
+++ b/src/Entity/EntityIdValue.php
@@ -13,6 +13,7 @@ use Wikibase\DataModel\LegacyIdInterpreter;
  * @license GPL-2.0+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
  * @author Thiemo Mättig
+ * @author Daniel Kinzler
  */
 class EntityIdValue extends DataValueObject {
 
@@ -27,22 +28,12 @@ class EntityIdValue extends DataValueObject {
 	 *
 	 * @since 7.0 serialization format changed in an incompatible way
 	 *
+	 * @note Do not use PHP serialization for persistence! Use a DataValueSerializer instead.
+	 *
 	 * @return string
 	 */
 	public function serialize() {
 		return serialize( $this->entityId );
-	}
-
-	/**
-	 * This method gets the numeric id from the serialization.
-	 * It makes assumptions we do not want to make about the id format,
-	 * though cannot be removed until we ditch the "numeric id" part
-	 * from the serialization.
-	 *
-	 * @return float Numeric id as a whole number. Can not be int because of 32-bit PHP.
-	 */
-	private function getNumericId() {
-		return floatval( substr( $this->entityId->getSerialization(), 1 ) );
 	}
 
 	/**
@@ -123,11 +114,16 @@ class EntityIdValue extends DataValueObject {
 	 * @return array
 	 */
 	public function getArrayValue() {
-		return [
+		$array = [
 			'entity-type' => $this->entityId->getEntityType(),
-			'numeric-id' => $this->getNumericId(),
-			'id' => $this->entityId->getSerialization(),
 		];
+
+		if ( $this->entityId instanceof Int32EntityId ) {
+			$array['numeric-id'] = $this->entityId->getNumericId();
+		}
+
+		$array['id'] = $this->entityId->getSerialization();
+		return $array;
 	}
 
 	/**
--- a/tests/unit/Entity/EntityIdValueTest.php
+++ b/tests/unit/Entity/EntityIdValueTest.php
@@ -3,9 +3,11 @@
 namespace Wikibase\DataModel\Tests\Entity;
 
 use PHPUnit_Framework_TestCase;
+use Wikibase\DataModel\Entity\EntityId;
 use Wikibase\DataModel\Entity\EntityIdValue;
 use Wikibase\DataModel\Entity\ItemId;
 use Wikibase\DataModel\Entity\PropertyId;
+use Wikibase\DataModel\Fixtures\CustomEntityId;
 
 /**
  * @covers Wikibase\DataModel\Entity\EntityIdValue
@@ -16,6 +18,7 @@ use Wikibase\DataModel\Entity\PropertyId;
  * @license GPL-2.0+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
  * @author Thiemo Mättig
+ * @author Daniel Kinzler
  */
 class EntityIdValueTest extends PHPUnit_Framework_TestCase {
 
@@ -29,26 +32,29 @@ class EntityIdValueTest extends PHPUnit_Framework_TestCase {
 	 * @dataProvider instanceProvider
 	 */
 	public function testSerialzationRoundtrip( EntityIdValue $id ) {
-		$newId = unserialize( serialize( $id ) );
+		$serialized = serialize( $id );
+		$newId = unserialize( $serialized );
 
 		$this->assertEquals( $id, $newId );
 	}
 
 	public function instanceProvider() {
 		$ids = [
-			new ItemId( 'Q1' ),
-			new ItemId( 'Q42' ),
-			new ItemId( 'Q31337' ),
-			new ItemId( 'Q2147483647' ),
-			new PropertyId( 'P1' ),
-			new PropertyId( 'P42' ),
-			new PropertyId( 'P31337' ),
+			'Q1' => new ItemId( 'Q1' ),
+			'Q42' => new ItemId( 'Q42' ),
+			'Q31337' => new ItemId( 'Q31337' ),
+			'Q2147483647' => new ItemId( 'Q2147483647' ),
+			'P1' => new PropertyId( 'P1' ),
+			'P42' => new PropertyId( 'P42' ),
+			'P31337' => new PropertyId( 'P31337' ),
+			'X567' => new CustomEntityId( 'X567' ),
+			'foo:P678' => new PropertyId( 'foo:P678' ),
 		];
 
 		$argLists = [];
 
-		foreach ( $ids as $id ) {
-			$argLists[] = [ new EntityIdValue( $id ) ];
+		foreach ( $ids as $k => $id ) {
+			$argLists[$k] = [ new EntityIdValue( $id ) ];
 		}
 
 		return $argLists;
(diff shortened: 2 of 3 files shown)


brightbyte:
Fails code style tests, will fix that soon. 
Please go ahead and review for logic.

tests/unit/Entity/EntityIdValueTest.php
```diff
@@ -186,4 +243,58 @@ public function invalidArrayProvider() {
...
+	 *
+	 * @return EntityId
+	 */
+	private function newCustomId( $string ) {
+		return new EntityIdValueTest_CustomEntityId( $string );
+	}
+
 }
+
+class EntityIdValueTest_CustomEntityId extends EntityId {
```
JeroenDeDauw: New file please

tests/unit/Entity/EntityIdValueTest.php
```diff
@@ -75,41 +78,95 @@ public function testGetSortKey( EntityIdValue $id ) {
 		$this->assertInternalType( 'string', $id->getSortKey() );
 	}
 
+	public function provideGetArrayValue() {
+		return [
+			'Q2147483647' => [
```
thiemowmde: The names for these test cases are really pointless and can as well be removed. The exact same string is in the next line.

tests/unit/Entity/EntityIdValueTest.php
```diff
@@ -75,41 +78,95 @@ public function testGetSortKey( EntityIdValue $id ) {
...
-	public function testGetArrayValueRoundtrip( EntityIdValue $id ) {
-		$newId = EntityIdValue::newFromArray( $id->getArrayValue() );
+	public function testGetArrayValue( EntityId $id, array $expected ) {
+		$value = new EntityIdValue( $id );
+		$array = $value->getArrayValue();
 
-		$this->assertEquals( $id, $newId );
+		ksort( $expected );
+		ksort( $array );
+		$this->assertEquals( $expected, $array );
```
thiemowmde: Please use `assertSame` and remove the two `ksort`. Order should be part of the contract, in my opinion. This was tested before in `testGetArrayValueCompatibility`, which you removed. The removal is fine, but the fact that these details are also tested should not be removed entirely.

tests/unit/Entity/EntityIdValueTest.php
```diff
@@ -186,4 +243,58 @@ public function invalidArrayProvider() {
...
 }
+
+class EntityIdValueTest_CustomEntityId extends EntityId {
+
+	/**
+	 * @param string $idSerialization
+	 *
+	 * @throws InvalidArgumentException
+	 */
+	public function __construct( $idSerialization ) {
```
thiemowmde: I believe you can remove this entire constructor and all tests will still succeed just fine.

tests/unit/Entity/EntityIdValueTest.php
```diff
@@ -186,4 +243,58 @@ public function invalidArrayProvider() {
...
+	}
+
+	/**
+	 * @see Serializable::unserialize
+	 *
+	 * @param string $serialized
+	 */
+	public function unserialize( $serialized ) {
+		$array = json_decode( $serialized );
+		$this->serialization = is_array( $array ) ? $array[1] : $serialized;
```
thiemowmde: Why does this custom type need this compatibility layer?

tests/unit/Entity/EntityIdValueTest.php
```diff
@@ -249,52 +254,8 @@ public function invalidArrayProvider() {
 	 * @return EntityId
 	 */
 	private function newCustomId( $string ) {
-		return new EntityIdValueTest_CustomEntityId( $string );
+		new HashArrayElement( 'Bla ');
```
thiemowmde: Huh?

brightbyte: Oops...

src/Entity/EntityIdValue.php
```diff
@@ -67,7 +60,7 @@ public function unserialize( $serialized ) {
 		try {
 			$entityId = LegacyIdInterpreter::newIdFromTypeAndNumber( $entityType, $numericId );
 		} catch ( InvalidArgumentException $ex ) {
-			throw new IllegalValueException( 'Invalid EntityIdValue serialization.' );
+			throw new IllegalValueException( 'Invalid EntityIdValue serialization.', 0, $ex );
```
brightbyte: It's here because it improves the error report when trying to use the LegacyIdInterpreter for custom IDs.

src/Entity/EntityIdValue.php
```diff
@@ -103,6 +96,7 @@ public function getSortKey() {
 	 * @return self
 	 */
 	public function getValue() {
+		// TODO: this should be changed to return the EntityId.
```
brightbyte: Because the actual value is the EntityId. But this method is pointless anyway. We should just delete it.

src/Entity/EntityIdValue.php
```diff
@@ -123,11 +117,16 @@ public function getEntityId() {
...
 	 */
 	public function getArrayValue() {
-		return [
+		$array = [
 			'entity-type' => $this->entityId->getEntityType(),
-			'numeric-id' => $this->getNumericId(),
 			'id' => $this->entityId->getSerialization(),
 		];
+
+		if ( $this->entityId instanceof Int32EntityId ) {
```
brightbyte: Sure. This is added here so this doesn't break for custom IDs that don't support numeric.

src/Entity/EntityIdValue.php
```diff
@@ -139,6 +138,8 @@ public function getArrayValue() {
 	 *
 	 * @param mixed $data
 	 *
+	 * @deprecated since 7.1, use callbacks in DataValueDeserializer instead
```
brightbyte: It's related in that it specifies what needs to be done in order to support custom IDs.

tests/unit/Entity/EntityIdValueTest.php
```diff
@@ -29,26 +32,25 @@ public function testCanConstruct() {
...
-			new ItemId( 'Q31337' ),
-			new ItemId( 'Q2147483647' ),
-			new PropertyId( 'P1' ),
-			new PropertyId( 'P42' ),
-			new PropertyId( 'P31337' ),
+			'Q1' => new ItemId( 'Q1' ),
+			'Q2147483647' => new ItemId( 'Q2147483647' ),
+			'P31337' => new PropertyId( 'P31337' ),
+			'X567' => $this->newCustomId( 'X567' ),
+			'foo:P678' => new PropertyId( 'foo:P678' ),
```
brightbyte: Asserting support for foreign and custom IDs is the point of this patch.

tests/unit/Entity/EntityIdValueTest.php
brightbyte: The key is shown when the test fails. Without it, there is no good way to see while case failed.

tests/unit/Entity/EntityIdValueTest.php
brightbyte: I disagree. Order should not be part of the contract. This array is used in the JSON serialization. The order of object fields is undefined in JSON.

brightbyte: changed to assertSame. The behavior of === on arrays seems odd to me. I guess I'm still thinking of arrays as objects. They are mutable, after all...

thiemowmde: We had this discussion before. There is no JSON here. This is a PHP array. The order is defined in a PHP array. The order was tested before, you removed it, and I disagree with this removal. Nothing ever unexpectedly broke because of us testing this.

tests/unit/Entity/EntityIdValueTest.php
thiemowmde: Sorry. What I meant is that the *removal* of test cases is unrelated and should be done in a separate patch.

src/Entity/EntityIdValue.php
thiemowmde: See #729.

tests/unit/Entity/EntityIdValueTest.php
brightbyte: If we guarantee the order of the fields, we have to preserve it over serialization/deserialization. We cannot do this with JSON, unless we add some magic field.

So, we can either drop the guarantee (which isn't explicitly stated anywhere, as far as I know), or we somehow ensure that it survives a JSON round trip (without relying on implementation details, such as the fact that php's json_encode/json_decode methods do preserve order).

Which one do you prefer?

tests/unit/Entity/EntityIdValueTest.php
brightbyte: Ok, I'll revert that. Seems pointless, though. These tests are redundant. To me, this is a case of "small fix when touching code close by".

tests/unit/Entity/EntityIdValueTest.php
thiemowmde: It sounds like you are deliberately confusing dependencies here. Again, this code here does not know anything about JSON. This outputs a PHP array and can easily give this guarantee. This does not mean unrelated code in a serializer has to obey.

thiemowmde: We are again talking about two entirely different idea of what "data model" means.

This is a PHP implementation that encodes a major part, but not al knowledge about the conceptual data model. Giving a guarantee about the order of array keys in a serialization is convenient and makes code that depends on this simpler and easier to maintain. It does not mean all code consuming this array must keep the key order, especially not if it just can't, as it is the case in JSON.

tests/unit/Entity/EntityIdValueTest.php
thiemowmde: The `EntityId` base class already has this constructor.

tests/unit/Entity/EntityIdValueTest.php
brightbyte: You are right regarding PHP vs. abstract data model. 
I disagree with your conclusion though.
The interface should not give guarantees no needed by the caller, based on the current implementation. 
Guaranteeing this order makes calling code more brittle, and the code itself harder to maintain.

Point in case is this patch. This test fails without the ksort. So should I make the code more complex to uphold a pointless guarantee, which is not stated anywhere, and which no production code relies on (I hope)? Or should I just drop the assumption, and make the test ablivious to the order? The latter seems easier, and gives more flexibility for the future.

Interface guarantees should never include anything that just happens to hold for the current implementation. An interface's contract needs to strike a balance between the stability needed by the callers, and the flexibility wanted to change the implementation. Needlessly restricting that flexibility is not a good thing.

thiemowmde: > […] makes calling code more brittle, and the code itself harder to maintain.

I hope you realize that I just said the exact opposite?

> […] not stated anywhere

The guarantee is "stated" as a fact in every single serialization we export.

> […] which no production code relies on (I hope)?

I also hope no production and no third party code relies on this. This does not mean it's "pointless" to give guarantees. It's simply convenient for developers to be able to look at an export and always find the keys in the same order.

Fixing this is super trivial:
```php
$array = [ 'first key' => … ];
if ( … ) {
    $array['second key'] = …;
}
$array['third key'] = …;
```

brightbyte: Sure it's trivial to fix in the production code. It's also trivial to fix in the test code, which is what I did.

Basically, my argument is: implementations should have as much "wiggle-room" as possible within the constraints of their interface contract. Making that contract more restrictive than it needs to be takes away autonomy from the imlementation.

The more guarantees we give, the more changes have to be considered breaking. That's not a good thing.

Relyling on the order of key/value pairs is generally a bad thing. We should not encourage it.

thiemowmde: I'm not saying we should encourage this. I'm not even saying we should promote this as a guarantee. All I say is that there is just no need to change the order now. We will break guarantees anyway the moment we drop the two legacy array keys entirely.

src/Entity/EntityIdValue.php
manicki: FTR: I've merged the other PR but seriously, I think it has taken more effort to make that PR, get it reviewed and now rebase this PR, than the distraction caused by sneaking in this particular change in this PR.

tests/unit/Entity/EntityIdValueTest.php
```diff
@@ -75,41 +82,98 @@ public function testGetSortKey( EntityIdValue $id ) {
...
-	public function testGetArrayValueRoundtrip( EntityIdValue $id ) {
-		$newId = EntityIdValue::newFromArray( $id->getArrayValue() );
+	public function testGetArrayValue( EntityId $id, array $expected ) {
+		$value = new EntityIdValue( $id );
+		$array = $value->getArrayValue();
 
-		$this->assertEquals( $id, $newId );
+		ksort( $expected );
+		ksort( $array );
+		$this->assertSame( $expected, $array );
```
manicki: Wouldn't assertEquals work here without needing to sort both arrays? Or does assertEquals is also less strict than assertSame in some other aspect than just the element order?

manicki:
Looks OK after a brief look

tests/unit/Entity/EntityIdValueTest.php
brightbyte: might work, will check

brightbyte:
Thanks Thiemo. I'm fine with the patch as it is now. IF you are fine with it too, go ahead and merge it.

PR #730, opened 2017-03-22
Remove redundant EntityIdValue test cases

thiemowmde:
This was originally part of #728. The idea of removing these test cases is that testing with a minimal and a larger ID should be enough. The other test cases do not add anything new.


thiemowmde:
Ping?

Issue #733, opened 2017-03-24
Remove deprecated Item::getSiteLinks


JeroenDeDauw:

JeroenDeDauw:
I propose to do this in 8.0. Has been deprecated since 0.8

thiemowmde:
Appears to be unused in all Wikibase related code I have on my machine, so: :+1:

PR #734, opened 2017-03-24
Undeprecate SiteLink methods in Item

JeroenDeDauw:
They where deprecated years ago and never removed, presumably cause there is no clear motivation to do so.

IIRC I deprecated this stuff, but at present I no longer think this makes things better.

Note that this leaves one of them in place (since I do think that one does not make sense to have here): https://github.com/wmde/WikibaseDataModel/issues/733

### DIFF
--- a/src/Entity/Item.php
+++ b/src/Entity/Item.php
@@ -204,7 +204,6 @@ class Item implements EntityDocument, FingerprintProvider, StatementListHolder,
 	 * If there already is a site link with the site id of the provided site link,
 	 * then that one will be overridden by the provided one.
 	 *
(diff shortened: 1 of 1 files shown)


thiemowmde:
* Note: `getSiteLinks` should be removed entirely. This is  already issued as #733.
* I suggest to remove `hasSiteLinks`. It's unused, and feels a bit awkward to me. Just get the list and check if it's empty. That appears much more expressive to me.
* I suggest to remove `removeSiteLink`. Also unused. This is a very rare use case, Just get the list and remove the element there. Maintaining the shortcut doesn't give much benefit.
* I agree to undeprecate the remaining 3 methods (add, has and get).

PR #736, opened 2017-04-04
Update DispatchingEntityIdParser.php

magnusmanske:
Show the problematic $idSerialization value, rather then the variable name...

### DIFF
--- a/src/Entity/DispatchingEntityIdParser.php
+++ b/src/Entity/DispatchingEntityIdParser.php
@@ -68,7 +68,7 @@ class DispatchingEntityIdParser implements EntityIdParser {
 	 */
 	private function assertIdIsString( $idSerialization ) {
 		if ( !is_string( $idSerialization ) ) {
-			throw new EntityIdParsingException( '$idSerialization must be a string' );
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
Tests fail:

> 1) Wikibase\DataModel\Tests\Entity\BasicEntityIdParserTest::testCannotParseInvalidId with data set #3 (array())

> Array to string conversion

probably want to put in a `var_export`

thiemowmde:
What is the actual problem you want to solve? Do you have a stack trace?

I don't think this is the right place for the message you want. This particular line checks the type and should probably be `throw new EntityIdParsingException( '$idSerialization must be a string, got ' . gettype( $idSerialization ) );`, but not dump the entirety of whatever it got with a magic `(string)` typecast.

If you are passing user input to this parser, I think you should do a type check closer to where the user input is coming from.

thiemowmde:
Closing now because there was no response for about 3 months, and I don't think outputting the type would solve the actual problem the user does have. Feel free to reopen or create a Phabricator ticket. Thanks.

PR #737, opened 2017-04-06
Disallow periods in repository names

manicki:
Periods would be confusing when entity ID is used in URIs/URLs as they might be considered "separators" of parts of URL/URI. Also periods are about to be used instead of colons in "chained" entity IDs in cases where colons are not allowed (e.g. RDF).


brightbyte:
@Thiemo: In wikitext and now in foreign IDs, we use recursive prefix handling: "foo:bar:x" is split into "foo" and "bar:x", and whatever server "foo" resolves to is responsible for resolving the rest of the ID. That server would then split "bar:x" into "bar" and "x", and would in turn as the server "bar" for resource "x", which that server could resolve locally.

While RDF Turtle allows such "chained" identifiers (":" is allowed in the local name production, https://www.w3.org/TR/turtle/#grammar-production-PN_LOCAL), XML qnames do not allow additional ":"  https://www.w3.org/TR/REC-xml-names/#NT-NCName. To avoid problems with the XML/RDF mapping, we replace any extra columns in EntityIds with dots (so "foo:bar:x" becomes "foo:bar.x"). Which in turn means repo names cannot contain dots (or columns). 

Note that such "chained" identifiers are really a failsafe. With proper and complete ID mappings set up, they should never occur. But we need *some* mechanism to represent "the thing that is called 'boo:x' on repo foo" internally, and in RDF.

manicki:
Thanks @brightbyte for providing more detail explanation. I've indeed failed to be specific in the PR description.
Regarding the "confusing" part: what I find a bit confusing is that with repository names containing periods, the identifier starts to "look like" the URI (e.g. "foo.test:Q123") is not far from the hypothetical URI.  As I get it, EntityIDs considered here should not be considered URIs.

brightbyte:
@manicki: It should be "foo:test.Q123", not "foo.test:Q123". Unless I'm *very* confused now.

PR #738, opened 2017-04-06
Remove no longer used AutoloadingAliasesTest

manicki:
Class aliases have been removed in 5a1e0823ca9003eb7163716dce5465d064b10c2e.
With this change, alias autoloading test is no longer used.


manicki:
yeah, it was not clear to me there are some class aliases coming which would make it somewhat justified to keep the no-op test in place

PR #740, opened 2017-04-10
Remove deprecated Item::hasSiteLinks

thiemowmde:
Unused. The replacement is much more expressive, in my opinion:
```php
$item->getSiteLinkList()->isEmpty()
```
See #734.

### DIFF
--- a/src/Entity/Item.php
+++ b/src/Entity/Item.php
@@ -267,16 +267,6 @@ class Item implements EntityDocument, FingerprintProvider, StatementListHolder,
 	}
 
 	/**
-	 * @deprecated since 0.8, use getSiteLinkList()->isEmpty() instead.
-	 * @since 0.5
-	 *
-	 * @return bool
-	 */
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
I don't see any reason to not have this shortcut. Why is

```php
$item->getSiteLinkList()->isEmpty()
```

better than

```php
$item->hasSiteLinks()
```

The second approach provides the benefit that users of it do not bind to the structure of the objects (law of demeter).

brightbyte:
For the record: we should either remove the method, or un-deprecate it. I don't care much either way. But given that I can find *no* code that checks whether a SiteLinkList is empty, it seems pointless to have a convenience method for doing so. I can follow the Law of Demeter argument, but only if it's useful in practice. Otherwise we end up duplicating the interfaces of internal structures.

My recommendation: don't waste time on arguing about unused stuff. Merge or abandon. I don't care.

PR #747, opened 2017-06-15
Flip mixed $expected and $actual parameters

thiemowmde:


Ladsgroup:
Travis failure does not seem to be related. Needs to be addressed ASAP though.

PR #749, opened 2017-06-15
Fix broken PHPCS rule set

thiemowmde:
The changes I made in #745 accidentally removed crucial exclude patterns.

Having these additional sniffs here, locally, and not in the Wikibase rule set, is intended exactly because of these exceptions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: php
 
+dist: trusty
+
 php:
   - 5.5
   - 5.6
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -14,10 +14,17 @@
 	<rule ref="Generic.Metrics.CyclomaticComplexity" />
 	<rule ref="Generic.Metrics.NestingLevel" />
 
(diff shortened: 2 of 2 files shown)


Ladsgroup:
Travis is still failing

manicki:
I believe it uses the right standard file, and I believe was using it already without this change. The thing seems to be that local phpcs.xml turns on some sniffs turned off in wikibase cs ruleset?
Namely Squiz.Strings.DoubleQuoteUsage.ContainsVar, PSR1.Files.SideEffects, and PSR1.Methods.CamelCapsMethodName in tests are excluded in wikibase ruleset, and then enabled locally which is a mistake I suppose?

thiemowmde:
\o/

.travis.yml
```diff
@@ -1,5 +1,7 @@
 language: php
 
+dist: trusty
```
manicki: out of curiosity: was this needed as well?

Ladsgroup: hhvm in travis only works in trusty and not the default distro (precise)

manicki: oh, right

PR #751, opened 2017-06-29
Add strict type hint to StatementGuid constructor

thiemowmde:
The only caller is in StatementGuidParser, and the function argument is guaranteed to be of this type there. No other caller exists (outside of tests).

I believe this was a mistake or fragment from a time where strict type hints did not existed.

### DIFF
--- a/src/Statement/StatementGuid.php
+++ b/src/Statement/StatementGuid.php
@@ -39,10 +39,7 @@ class StatementGuid implements Comparable {
 	 *
 	 * @throws InvalidArgumentException
 	 */
-	public function __construct( $entityId, $guid ) {
-		if ( !( $entityId instanceof EntityId ) ) {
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
> There were 5 errors:
> 
> 1) Wikibase\DataModel\Tests\Statement\StatementGuidTest::testBadConstruction with data set #0 ('foobar', 'foobar')
> 
> Argument 1 passed to Wikibase\DataModel\Statement\StatementGuid::__construct() must be an instance of Wikibase\DataModel\Entity\EntityId, string given, called in /home/travis/build/wmde/WikibaseDataModel/tests/unit/Statement/StatementGuidTest.php on line 62 and defined

thiemowmde:
Thanks for the note. Fixed now.

PR #753, opened 2017-07-24
Do not let int as value in Item::setId and Property::setId

Ladsgroup:
Bug: T151577

### DIFF
--- a/src/Entity/Item.php
+++ b/src/Entity/Item.php
@@ -87,15 +87,13 @@ class Item implements EntityDocument, FingerprintProvider, StatementListHolder,
 	 * Can be ItemId since 0.5.
 	 * Can be null since 1.0.
 	 *
-	 * @param ItemId|int|null $id
+	 * @param ItemId|null $id
 	 *
 	 * @throws InvalidArgumentException
 	 */
 	public function setId( $id ) {
 		if ( $id === null || $id instanceof ItemId ) {
 			$this->id = $id;
-		} elseif ( is_int( $id ) ) {
-			$this->id = ItemId::newFromNumber( $id );
 		} else {
 			throw new InvalidArgumentException( '$id must be an instance of ItemId, an integer,'
 				. ' or null' );
--- a/src/Entity/Property.php
+++ b/src/Entity/Property.php
@@ -85,15 +85,13 @@ class Property implements EntityDocument, FingerprintProvider, StatementListHold
 	 * Can be PropertyId since 0.5.
 	 * Can be null since 1.0.
 	 *
-	 * @param PropertyId|int|null $id
+	 * @param PropertyId|null $id
 	 *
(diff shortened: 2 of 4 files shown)


JeroenDeDauw:
I actually misread the code at first due to the title. You are not changing `Entity::setId`, you are changing `Item::setId` and `Property::setId`.

I can see why accepting an integer in `Entity::setId` is problematic but don't see much improvement in changing this for Item and Property. Seems to mainly create busywork for yourself.

Ladsgroup:
The more details about why we are doing it can be found in https://phabricator.wikimedia.org/T151577
 Also I don't appreciate judging how useful my work is or guessing my intentions.

JeroenDeDauw:
I'm sorry if I upset you somehow. This comes as a big surprise to me as all I did was saying I do not get the reason for doing this and raising a concern. I did not say there is no improvement and neither did I object to you making this change. Not sure what format would be better to provide this kind of feedback. Any suggestion?

JeroenDeDauw:
Briefly talked to someone about my and your comment. To clarify, what I wrote does not contain hidden accusation or implication. My assumption is you are making this change in good faith. In particular, I do not think you are trying to create busywork for the sake of it. Still, it seems to me like that it was this change does, which is why I let you know my impression.

brightbyte:
Rationale: Removing support for plain integer IDs throughout the code base is intended to avoid errors, and flush out old code that still uses numeric IDs. Being strict is generally better than being "nice".

brightbyte:
Regarding busy-work: the discussion here has probably cost three people more time and energy already than writing this code.

manicki:
Exception message has not been updated and is now incorrect.

BUSYWORK WASTE O'TIME ETC WARNING!!!!
Speaking of busywork, having noticed this PR I was wondering why not change the whole EntityDocument interface and not explicitly say setId should get an EntityId object? This seemed almost trivial after this change except the fact that apparently setId also accepts null. I wonder whether this makes change from the data model perspective to set identifier of the entity that already has some ID to the blank ID? I thought that maybe it does not, so why not change it only expect non-null EntityId object?

As I had a spare minute I went ahead and made this change: https://github.com/wmde/WikibaseDataModel/tree/entity-setid-nonull, and also made a rough proof of concept adjusting setId's uses in Wikibase: https://gerrit.wikimedia.org/r/368418 (as hopefully could be seen there, mostly usages in tests needed to be changed). I intentionally do not submit change in this repo as PR (at least not yet) as this is not critical change and I don't want to create impression I make people use their time looking at this kind of experiment. Just look at this if you find this interesting. You've been warned!

Also note: I am not aware of specifics of usages of this lib outside of Wikibase so it may be all wrong

manicki:
meh, the wikibase patch on gerrit fails badly because of other libs requirements. Will hack it all way to have tests run on Jenkins later.

PR #754, opened 2017-07-25
Update release notes for 8.0.0

Ladsgroup:
The change that recently got merged is a breaking change.
So it needs a major release.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -2,10 +2,12 @@
 
 ## Version 8.0.0 (dev)
 
+This release removes integers as acceptable entity IDs.
+
 #### Breaking changes
-* `Item:setId` as well as `Property::setId` do not accept integers any more.
+* `Item::setId` and `Property::setId` do not accept integers any more.
 * Removed `Item::getSiteLinks` and `Item::hasSiteLinks`.
-* `RepositoryNameAssert` now also disallows periods, in addition to colons.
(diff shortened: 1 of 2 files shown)


RELEASE-NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Wikibase DataModel release notes
 
+## Version 8.0.0 (2017-06-25)
+
+This release removes integar as acceptable value in Item::setId() and Property::setId()
```
manicki: Typo. "integer" not "integar".

JeroenDeDauw:
In the past we tended to wait till we had some breaking changes before merging them and making a release. This because making a release comes with a cost, especially if it is a major one, since then consumers often need to update their list of supported version numbers. We did this "wait till there is sufficient reason for the work" especially for changes not strictly needed by consumers, such as code cleanup, which the one change this release contains so far is. You can see a huge difference between this major release and the past ones in the release notes.

So my recommendation is to revert the change already merged, submit a new PR that reverts the revert, tag it as 8.0 and merge it when you have some other breaking changes. (And close this PR.) Of course if you think this one change is worth spending time on like this I won't stop you.

bekh6ex:
@JeroenDeDauw Valid point. Lets wait.

JeroenDeDauw:
@thiemowmde it'd have been good to amend the commit message though, I just got confoosed by a "released version 8.0.0" on master

PR #756, opened 2017-08-02
Fix forgotten Item/Property::setId documentation and exception

thiemowmde:
Introduced in #753.

Three people discussed there, and a fourth person noticed this mistake, but 5 days later nobody was able to fix it. I find this quite fascinating.

I agree with Jeroen in so far that the removal done in #753 was not required to fulfill what the ticket asked for: it's perfectly possible to **not** call `setId` with integers, even if the method still accepts integers. However, it's obviously much easier to find unwanted usages if `setId` fails with an exception if ever called with an integer.

The only downside of the removal is that test code becomes harder to write. I agree that calls like `$item->setId( 1 )` must be burned with fire. But `$item->setId( new ItemId( 'Q1' ) )` is quite ugly. What about allowing `$item->setId( 'Q1' )`? Is this worth it? (I know we already have `NewItem::withId( 'Q1' )->build()`.)

### DIFF
--- a/src/Entity/Item.php
+++ b/src/Entity/Item.php
@@ -83,21 +83,18 @@ class Item implements EntityDocument, FingerprintProvider, StatementListHolder,
 	}
 
 	/**
-	 * Can be integer since 0.1.
-	 * Can be ItemId since 0.5.
-	 * Can be null since 1.0.
+	 * @since 0.5, can be null since 1.0
 	 *
 	 * @param ItemId|null $id
 	 *
 	 * @throws InvalidArgumentException
 	 */
 	public function setId( $id ) {
-		if ( $id === null || $id instanceof ItemId ) {
-			$this->id = $id;
-		} else {
-			throw new InvalidArgumentException( '$id must be an instance of ItemId, an integer,'
-				. ' or null' );
+		if ( !( $id instanceof ItemId ) && $id !== null ) {
+			throw new InvalidArgumentException( '$id must be an ItemId or null' );
 		}
+
+		$this->id = $id;
 	}
 
 	/**
--- a/src/Entity/Property.php
+++ b/src/Entity/Property.php
@@ -81,21 +81,18 @@ class Property implements EntityDocument, FingerprintProvider, StatementListHold
 	}
 
 	/**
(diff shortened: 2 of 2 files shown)


src/Entity/Item.php
```diff
@@ -83,21 +83,18 @@ public function getId() {
...
 	 *
 	 * @throws InvalidArgumentException
 	 */
 	public function setId( $id ) {
-		if ( $id === null || $id instanceof ItemId ) {
-			$this->id = $id;
-		} else {
-			throw new InvalidArgumentException( '$id must be an instance of ItemId, an integer,'
-				. ' or null' );
+		if ( !( $id instanceof ItemId ) && $id !== null ) {
```
manicki: personally I'd have flipped the condition here but whatever.

thiemowmde: I did this because the documentation also says `ItemId|null`.

thiemowmde:
> What is "quite ugly" about it?

`$item->setId( new ItemId( 'Q1' ) )` is just painfull to write. If I can't pass anything but an ItemId (or null), why do I have to repeat the ItemId constructor call each time? `$item->setId( 'Q1' )` would be so much easier, with no negative consequence.

But this is only relevant in tests, and we do have the NewItem builder, so this is only a minor complain. ;-)

src/Entity/Item.php
manicki: I meant that "if ( $id instanceof ItemId || $id === nulll )" also means ItemId|null

thiemowmde: Ok. But then the assignment would be in the if, and the exceptional case in the else (or below the if). I wanted to use the guard pattern here, with the exceptional case first.

manicki: yeah, I don't have better option that put the condition in the private method. But it is all minor and, if this wasn't clear, I am not saying this MUST be changed now to any other form.

thiemowmde:
@manicki, I believe what we discussed here is unrelated to the actual changes made in this pull request. I believe this is nothing but a fixup for a mistake done in #753 that should be merged soon, so we do not forget it when doing the 8.0.0 release.

PR #760, opened 2017-09-18
EXPERIMENT: infrastructure for hierarchicaly adressable entities

brightbyte:


src/Entity/HierarchicalEntityId.php
```diff
@@ -0,0 +1,124 @@
...
+ * consist of the base (parent) EntityId and a relative part. The base entity may itself be
+ * hierarchical. Only the first (root) entity ID in a chain may be a foreign ID.
+ *
+ * Hierarchical entity IDs are serialized as the serialization of the base entity IDs, followed
+ * by a dash, followed by the child part of the ID.
+ *
+ * @license GPL-2.0+
+ * @author Daniel Kinzler
+ */
+abstract class HierarchicalEntityId extends EntityId {
```
bekh6ex: It should not extend EntityId, because EntityId can contain prefixes which doesn't make sense in this case, so those two should be distinguished.

brightbyte: A hierarchical ID can have a prefix - just not in the relative part. A hierarchical ID's prefix is the prefix of the root ID.

But even if it could not:  It's fine for any subclass of EntityId to guarantee that it does not have a prefix. This does not violate LSP (because there is no setter). It just adds a guarantee. So I see no reason to not use a subclass.

Indeed, making the two incompatible would defeat the purpose. The idea is to be able to use a hierarchical ID wherever you could use a non-hierarchical ID.

PR #761, opened 2017-09-23
Heavily optimize SnakList::moveSnaksToBottom

mariushoch:
I found this to be rather inefficient when profiling the dump generation, thus I looked into it. Currently when dumping the first 1,000 entities this takes about 10% of the time, the new version only takes about 6% of the time, thus making the dump approx. 4% faster. This will also effect the API and many other DM users to some extent.

Note: Instead of using $this->offsetHashes, I could have also either used an Iterator or an array derived from the object, but that's slower as it requires generating a new object/ array.

### DIFF
--- a/src/Snak/SnakList.php
+++ b/src/Snak/SnakList.php
@@ -180,6 +180,21 @@ class SnakList extends ArrayObject implements Comparable, Hashable {
 	 * @param Snak[] $snaks to remove and re add
 	 */
 	private function moveSnaksToBottom( array $snaks ) {
+		// Skip Snaks that are already at the bottom:
+		// Find the last element in the array by looking at the last element
+		// in the offsetHashes array (this works as they are always added side-by-side).
+		$offsets = $this->offsetHashes;
+
+		$snakCount = count( $snaks );
+		for ( $i = 0; $i < $snakCount; $i++ ) {
+			$lastOffset = array_pop( $offsets );
+			if ( $this[$lastOffset] === $snaks[$i] ) {
+				unset( $snaks[$i] );
+			} else {
+				break;
+			}
+		}
+
 		foreach ( $snaks as $snak ) {
 			$this->removeSnak( $snak );
 			$this->addSnak( $snak );


JeroenDeDauw:
Heavy?

image

---

/me whispers StatementGroup

thiemowmde:
I'm afraid the orderByProperty method affected by this patch is not covered enough. There are only tests with either none or a single element in the $order array. Please add a few more complex tests that succeed with both the old and the new implementation. Suggestion for such a test:
```php
[
	[
		new PropertyValueSnak( $id1, new StringValue( 'a' ) ),
		new PropertyValueSnak( $id2, new StringValue( 'b' ) ),
		new PropertyValueSnak( $id1, new StringValue( 'c' ) ),
		new PropertyValueSnak( $id3, new StringValue( 'd' ) ),
		new PropertyValueSnak( $id4, new StringValue( 'e' ) ),
		new PropertyValueSnak( $id2, new StringValue( 'f' ) ),
		new PropertyValueSnak( $id4, new StringValue( 'g' ) ),
	],
	[
		new PropertyValueSnak( $id2, new StringValue( 'b' ) ),
		new PropertyValueSnak( $id2, new StringValue( 'f' ) ),
		new PropertyValueSnak( $id3, new StringValue( 'd' ) ),
		new PropertyValueSnak( $id1, new StringValue( 'a' ) ),
		new PropertyValueSnak( $id1, new StringValue( 'c' ) ),
		new PropertyValueSnak( $id4, new StringValue( 'e' ) ),
		new PropertyValueSnak( $id4, new StringValue( 'g' ) ),
	],
	[ 'P2', 'P3', 'P1' ]
],
```
I also played around with the code a little bit. The following implementation reduces the runtime for an arbitrary test setup on my local system from 2.9s to 2.6s. @mariushoch, can you run this code with the same test data you used for your version?
```php
public function orderByProperty( array $order = [] ) {
	$byProperty = array_combine( $order, array_fill( 0, count( $order ), [] ) );

	/** @var Snak $snak */
	foreach ( $this as $snak ) {
		$byProperty[$snak->getPropertyId()->getSerialization()][] = $snak;
	}

	$ordered = [];
	foreach ( $byProperty as $snaks ) {
		$ordered = array_merge( $ordered, $snaks );
	}

	$this->exchangeArray( $ordered );

	$this->offsetHashes = [];
	$index = 0;
	/** @var Snak $snak */
	foreach ( $this as $snak ) {
		$this->offsetHashes[$snak->getHash()] = $index++;
	}
}
```

thiemowmde:
**PS:** I edited and simplified the code above even more.

Also note that the test should check that the state of the internal $this->offsetHashes array is correct after orderByProperty was called. This is easy to miss and will mess up a bunch of other methods.

mariushoch:
I just benchmarked your code as well (profile excerpts `dumpJson.php --wiki wikidatawiki --sharding-factor 4 --shard 0 --limit 1000 --snippet`, one run only):

Currently deployed version (run just now):
`13.40% 7919.955  51332 - Wikibase\DataModel\Snak\SnakList::orderByProperty`
My version (run from October 23):
`8.77% 4832.099  51027 - Wikibase\DataModel\Snak\SnakList::orderByProperty`
Your version:
`4.83% 2606.631  51332 - Wikibase\DataModel\Snak\SnakList::orderByProperty`

The output didn't differ at all.

@thiemowmde Do you want to take over this PR or shall I?

thiemowmde:
I uploaded https://github.com/wmde/WikibaseDataModel/pull/762. @mariushoch, please review it and either merge it if you think it's fine, or document your +2 in a comment and I will merge it.

I'm willing to merge this patch here no matter what, but I think it needs at least the tests I now added to https://github.com/wmde/WikibaseDataModel/pull/762.

Also please tell me to which Phabricator ticket this refactoring belongs. Is it https://phabricator.wikimedia.org/T176844?

mariushoch:
> I'm willing to merge this patch here no matter what, but I think it needs at least the tests I now added to #762.

This patch is superfluous after #762 (the function modified is removed there), so I'll rather close it.

>Also please tell me to which Phabricator ticket this refactoring belongs. Is it https://phabricator.wikimedia.org/T176844?

There's no ticket about this specifically. I'll look into that.

PR #763, opened 2017-10-05
[WIP] Do not build SnakList on top of ArrayObject

thiemowmde:
The main reason to touch this file again is the awkward complexity it had before. Note that the data structure was build in a way that it kept track of **two** keys for each element. One key is the snak hash, the other one whatever you provided via the constructor.

WIP mostly because of the Iterator methods (rewind, current, and so on). These expose the internal state of the $this->snaks array. I'm not sure if this is robust enough. The array pointer might change when using other SnakList methods.

### DIFF
--- a/src/Snak/SnakList.php
+++ b/src/Snak/SnakList.php
@@ -2,16 +2,18 @@
 
 namespace Wikibase\DataModel\Snak;
 
-use ArrayObject;
 use Comparable;
+use Countable;
 use Hashable;
 use InvalidArgumentException;
+use Iterator;
+use Serializable;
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
The index change is nice though technically a breaking change. Since making a major release is currently very expensive, it might be better to avoid this one. We could also just present it is not a breaking change... though that is a bit scary. Something could be relying on the old behavior.

As to the Iterator stuff: you can implement IteratorAggregate instead. And in its getIterator method, just do

    yield from $this->snaks;

PR #764, opened 2017-10-05
Optimized DispatchingEntityIdParser

mariushoch:
This made the function quite a bit faster:
> 14.48% 48797.778 817218 - Wikibase\DataModel\Entity\DispatchingEntityIdParser::parse

to

> 12.64% 42330.389 817218 - Wikibase\DataModel\Entity\DispatchingEntityIdParser::parse
when JSON-dumping 10,000 Wikidata entities.

Related to https://phabricator.wikimedia.org/T177486.

### DIFF
--- a/src/Entity/DispatchingEntityIdParser.php
+++ b/src/Entity/DispatchingEntityIdParser.php
@@ -37,20 +37,19 @@ class DispatchingEntityIdParser implements EntityIdParser {
 	 * @return EntityId
 	 */
 	public function parse( $idSerialization ) {
-		$this->assertIdIsString( $idSerialization );
-
-		if ( empty( $this->idBuilders ) ) {
+		if ( $this->idBuilders === [] ) {
 			throw new EntityIdParsingException( 'No id builders are configured' );
 		}
 
+		try {
+			list( , , $localId ) = EntityId::splitSerialization( $idSerialization );
+		} catch ( InvalidArgumentException $ex ) {
+			// EntityId::splitSerialization performs some sanity checks which
+			// might result in an exception. Should this happen, re-throw the exception message
+			throw new EntityIdParsingException( $ex->getMessage(), 0, $ex );
+		}
+
 		foreach ( $this->idBuilders as $idPattern => $idBuilder ) {
-			try {
-				list( , , $localId ) = EntityId::splitSerialization( $idSerialization );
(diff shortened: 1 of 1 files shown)


src/Entity/DispatchingEntityIdParser.php
```diff
@@ -27,6 +27,10 @@ class DispatchingEntityIdParser implements EntityIdParser {
 	 * @param callable[] $idBuilders
 	 */
 	public function __construct( array $idBuilders ) {
+		if ( $this->idBuilders === [] ) {
+			throw new InvalidArgumentException( '$idBuilders must not be empty' );
```
thiemowmde: Uh. I find this a little problematic. I believe this was done as it was before for a reason: The idea was to allow a setup with *no* custom entity types. As long as the parse method is not called, it is not a problem to have an empty ID parser.

Even if such a situation can never happen in production, this change makes this entire patch a breaking one. We could avoid this by keeping the previous handling of empty arrays.

The least minimum is to documented this as a `@throws`.

src/Entity/DispatchingEntityIdParser.php
```diff
@@ -37,20 +41,19 @@ public function __construct( array $idBuilders ) {
 	 * @return EntityId
 	 */
 	public function parse( $idSerialization ) {
-		$this->assertIdIsString( $idSerialization );
+		try {
+			list( , , $localId ) = EntityId::splitSerialization( $idSerialization );
+		} catch ( InvalidArgumentException $ex ) {
+			// Only do this assert in case something is wrong, EntityId::splitSerialization already
+			// checks for string.
+			$this->assertIdIsString( $idSerialization );
```
thiemowmde: Why keep this additional check *inside* the catch? This looks awkward. What does this do in addition to what splitSerialization already does? Let's remove it, please.

src/Entity/DispatchingEntityIdParser.php
mariushoch: Moved back, this probably hardly has any performance impact. It mostly looked weird to me.

src/Entity/DispatchingEntityIdParser.php
mariushoch: changing it will change the Exception message slightly, but I guess that's ok.

mariushoch:
Note: This is no longer breaking

PR #765, opened 2017-10-23
Release 7.2.0

thiemowmde:
@mariushoch, please review this carefully. The master branch contains other changes to the SnakList class and I'm really not sure if we can backport this so easily. On the other hand, if all tests succeed, why not?

[Bug: T177486](https://phabricator.wikimedia.org/T177486)

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,11 @@
 # Wikibase DataModel release notes
 
+## Version 7.2.0 (2017-10-23)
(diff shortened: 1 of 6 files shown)


mariushoch:
The only relevant change to master was e23e3c098c93153159c26d13eb97b8d3ddd0adec, which should not affect any of the functionality your patch touches. `DispatchingEntityIdParser` had no relevant changes at all. So this should be good to go.

PR #767, opened 2017-11-06
Optimize EntityId::getRepositoryName and ::getLocalPart

manicki:
This is implementing some changes suggested in https://phabricator.wikimedia.org/T162371 (see https://phabricator.wikimedia.org/T162371#3514607 in particular).

This change is then meant mostly as a, possibly temporary, means to have a code generating RDF output in Wikibase more performant. I can see this kind of change might not be something which is generally wanted, approved, etc. The only alternative to this I can see regarding the performance of RDF generating would be quite thorough re-structuring of the RDF-related code in Wikibase. While I actually thing it would be something which would (very) good in general, it is also (sadly) not something that is going to happen any time soon.

Final note: this is a PR against the 7.x branch. If the change is approved, I will also submit the similar change onto master, so this is not somehow obscurely hidden in 7.x only.

### DIFF
--- a/src/Entity/EntityId.php
+++ b/src/Entity/EntityId.php
@@ -16,6 +16,16 @@ abstract class EntityId implements Comparable, Serializable {
 
 	protected $serialization;
 
+	/**
+	 * @var string
+	 */
+	protected $repositoryName;
+
+	/**
+	 * @var string
+	 */
+	protected $localPart;
+
 	const PATTERN = '/^:?(\w+:)*[^:]+\z/';
 
 	/**
@@ -26,6 +36,8 @@ abstract class EntityId implements Comparable, Serializable {
 	public function __construct( $serialization ) {
 		self::assertValidSerialization( $serialization );
 		$this->serialization = self::normalizeIdSerialization( $serialization );
+
+		list ( $this->repositoryName, $this->localPart ) = $this->getRepositoryNameAndLocalPart( $serialization );
 	}
 
 	private static function assertValidSerialization( $serialization ) {
@@ -68,7 +80,11 @@ abstract class EntityId implements Comparable, Serializable {
 	public static function splitSerialization( $serialization ) {
 		self::assertValidSerialization( $serialization );
 
-		$parts = explode( ':', self::normalizeIdSerialization( $serialization ) );
+		return self::getSerializationParts( self::normalizeIdSerialization( $serialization ) );
+	}
+
+	private static function getSerializationParts( $serialization ) {
+		$parts = explode( ':', $serialization );
 		$localPart = array_pop( $parts );
 		$repoName = array_shift( $parts );
 		$prefixRemainder = implode( ':', $parts );
@@ -112,9 +128,7 @@ abstract class EntityId implements Comparable, Serializable {
 	 * @return string
 	 */
 	public function getRepositoryName() {
-		$parts = self::splitSerialization( $this->serialization );
-
-		return $parts[0];
+		return $this->repositoryName;
 	}
 
 	/**
@@ -125,9 +139,7 @@ abstract class EntityId implements Comparable, Serializable {
 	 * @return string
 	 */
 	public function getLocalPart() {
-		$parts = self::splitSerialization( $this->serialization );
-
-		return self::joinSerialization( [ '', $parts[1], $parts[2] ] );
+		return $this->localPart;
 	}
 
 	/**
@@ -177,6 +189,12 @@ abstract class EntityId implements Comparable, Serializable {
 
 		return $target instanceof self
 			&& $target->serialization === $this->serialization;
+	}
+
+	protected function getRepositoryNameAndLocalPart( $serialization ) {
+		list( $repoName, $prefixRemainder, $localId ) = self::getSerializationParts( $serialization );
+		$localPart = self::joinSerialization( [ '', $prefixRemainder, $localId ] );
+		return [ $repoName, $localPart ];
 	}
 
 }
--- a/src/Entity/EntityId.php
+++ b/src/Entity/EntityId.php
@@ -17,11 +17,15 @@ abstract class EntityId implements Comparable, Serializable {
 	protected $serialization;
 
 	/**
+	 * @since 7.3
+	 *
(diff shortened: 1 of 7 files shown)


src/Entity/EntityId.php
```diff
@@ -179,4 +191,10 @@ public function equals( $target ) {
 			&& $target->serialization === $this->serialization;
 	}
 
+	protected function getRepositoryNameAndLocalPart( $serialization ) {
```
brightbyte: This could be static, and perhaps should be.
Also, the name sounds like a getter, which this is not. Perhaps "computeRepositoryNameAndLocalPart" - or "splitIntoRepositoryNameAndLocalPart".
Also, please document contract and return type.

src/Entity/EntityId.php
```diff
@@ -68,7 +80,11 @@ public function getSerialization() {
 	public static function splitSerialization( $serialization ) {
 		self::assertValidSerialization( $serialization );
 
-		$parts = explode( ':', self::normalizeIdSerialization( $serialization ) );
+		return self::getSerializationParts( self::normalizeIdSerialization( $serialization ) );
+	}
+
+	private static function getSerializationParts( $serialization ) {
```
brightbyte: The name sounds like a getter, which this is not. "splitSerialization" would be descriptive, but it's take. splitSerializationInternal is ugly, but ok for a private method, I guess.

Alternatively, don't add this method at all. Why can't we just use splitSerialization?

src/Entity/EntityId.php
```diff
@@ -179,4 +191,10 @@ public function equals( $target ) {
 			&& $target->serialization === $this->serialization;
 	}
 
+	protected function getRepositoryNameAndLocalPart( $serialization ) {
+		list( $repoName, $prefixRemainder, $localId ) = self::getSerializationParts( $serialization );
```
brightbyte: This bypasses validation. That seems odd, but if this is really desired, the contract documentation should make this clear.

src/Entity/EntityId.php
```diff
@@ -68,7 +80,11 @@ public function getSerialization() {
 	public static function splitSerialization( $serialization ) {
 		self::assertValidSerialization( $serialization );
 
-		$parts = explode( ':', self::normalizeIdSerialization( $serialization ) );
+		return self::getSerializationParts( self::normalizeIdSerialization( $serialization ) );
+	}
+
```
brightbyte: Please document at least the return type.

src/Entity/EntityId.php
manicki: This indeed was a point. Namely, it is meant to be used in unserialize methods. As it has been bypassing validation for the whole "serialization", I went for the same behaviour here. I don't claim it is a right thing to do, but OTOH it was kind of something Stas was talking about :) (do not validate stuff you know is right).

src/Entity/EntityId.php
manicki: True. Renamed to extractRepositoryNameAndLocalPart, and made static.

src/Entity/EntityId.php
manicki: So it was introduced to be able to chop serialization into pieces but not run validation checks.

src/Entity/EntityId.php
manicki: done

src/Entity/EntityId.php
```diff
@@ -179,4 +203,18 @@ public function equals( $target ) {
...
+	 * Returns a pair (repository name, local part of ID) from the given ID serialization.
+	 * Note: this does not perform any validation of the given input. Calling code should take
+	 * care of this!
+	 *
+	 * @param string $serialization
+	 * @return string[] Array of form [ string $repositoryName, string $localPart ]
+	 */
+	protected static function extractRepositoryNameAndLocalPart( $serialization ) {
+		list( $repoName, $prefixRemainder, $localId ) = self::extractSerializationParts( $serialization );
+		$localPart = self::joinSerialization( [ '', $prefixRemainder, $localId ] );
```
thiemowmde: The fact that the two last elements are never used independent from each other makes this look like it should be optimized even further: Just do `explode( ':', $serialization, 2 )` and return the second element. Done.

brightbyte: IIRC, we considered this when we first designed this class, and we found that we sometimes need only the first element, and sometimes only the last. But I don't remember why, exactly, but there was *some* reason.

manicki: The reason why splitSerialization does return always three parts was indeed that sometimes caller might want first element and second+third together, sometimes it might though only want third element separately.

In this case though, what @thiemowmde suggest would be fine. Only possible concern (I am not saying it is) is then the detail about splitting by colon is in two places instead of just one. Having the separator details encapsulated in a single method used to be considered important if I recall correctly.

manicki:
Re
> which change in the code made ["",""] invalid?

Unserializing '' as an ID became invalid when unserialize called extractRepositoryNameAndLocalPart, which was using joinSerialization that has sanity check for ID being an empty string. With the suggestion two use explode instead of split/joinSerialization in extractRepositoryNameAndLocalPart, it is again fine to unserialize empty string as an Item or Property ID.
Fine with me as it used to be like this. It seems a bit iffy though, why allow unserializing obvious erroneous data.

src/Entity/EntityId.php
```diff
@@ -179,4 +203,18 @@ public function equals( $target ) {
...
+	 * Note: this does not perform any validation of the given input. Calling code should take
+	 * care of this!
+	 *
+	 * @param string $serialization
+	 * @return string[] Array of form [ string $repositoryName, string $localPart ]
+	 */
+	protected static function extractRepositoryNameAndLocalPart( $serialization ) {
+		$parts = explode( ':', $serialization, 2 );
+		return count( $parts ) > 1 ? [ $parts[0], $parts[1] ] : [ '', $parts[0] ];
+		return [ $repoName, $localPart ];
```
brightbyte: Unreachable code.

manicki: oops, thanks Thiemo for fixing.

PR #771, opened 2017-11-22
Streamline ItemId and PropertyId constructors a little

thiemowmde:
* Remove two is_string checks that are effectively dead code. The variable is coming from a splitSerialization call, which is guaranteed to return strings. And even if, the next preg_match check will still catch all errors.
* I move the strtoupper down, mostly for performance reasons.
* The shorter variable name $parts makes this much better readable. Before, this short method repeated the word "serialization" 8 times.

### DIFF
--- a/src/Entity/ItemId.php
+++ b/src/Entity/ItemId.php
@@ -22,19 +22,14 @@ class ItemId extends EntityId implements Int32EntityId {
 	 * @throws InvalidArgumentException
 	 */
 	public function __construct( $idSerialization ) {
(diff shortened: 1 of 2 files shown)


mariushoch:
@thiemowmde Do you want to apply this to master? Would make sense to me.

thiemowmde:
Yea, that was the intention. Thanks for the reminder. Done via 18f87fcc2521da9bdbf4b65a712f1563c40f9eb7.

PR #774, opened 2018-03-07
phpdoc: TermList::getIterator can return an Iterator

addshore:
Without this type hint we get an IDE error (and probably further static analysis errors) as
SpecialNewLexeme::createSummary calls getIterator and calls current() on the result
which doesn't exist in the current typehint.

This is a 2 part fix, getIterator should document that it can return an Iterator, and also
the createSummary method should probably check that it actually has a Iterator instead
of a just a Traversable.

### DIFF
--- a/src/Term/TermList.php
+++ b/src/Term/TermList.php
@@ -6,6 +6,7 @@ use ArrayIterator;
 use Comparable;
 use Countable;
 use InvalidArgumentException;
+use Iterator;
 use IteratorAggregate;
 use OutOfBoundsException;
 use Traversable;
@@ -66,7 +67,7 @@ class TermList implements Countable, IteratorAggregate, Comparable {
 
 	/**
 	 * @see IteratorAggregate::getIterator
-	 * @return Traversable|Term[]
+	 * @return Traversable|Iterator|Term[]
 	 */
 	public function getIterator() {
 		return new ArrayIterator( $this->terms );
--- a/src/Term/TermList.php
+++ b/src/Term/TermList.php
@@ -9,7 +9,6 @@ use InvalidArgumentException;
 use Iterator;
 use IteratorAggregate;
 use OutOfBoundsException;
-use Traversable;
 
 /**
  * Unordered list of Term objects.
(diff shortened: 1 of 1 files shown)


manicki:
Seems right, plus "and also
the createSummary method should probably check that it actually has a Iterator instead
of a just a Traversable." also makes sense.

The patch made my wonder though, maybe the code using this method should not rely on it returning the iterator, i.e. the use of current() there in SpecialNewLexeme should be somehow fixed?

addshore:
The related change on gerrit is https://gerrit.wikimedia.org/r/#/c/416918/

> The patch made my wonder though, maybe the code using this method should not rely on it returning the iterator, i.e. the use of current() there in SpecialNewLexeme should be somehow fixed?

Perhaps, for now the gerrit change just adds an assert.

I did a quick google and found https://github.com/wmde/traversable-iterator (hehe, it is ours)...
Perhaps we should check if we have a traversable or iterator and if we have a traversable use that library to get an iterator?

manicki:
we could consider this yes. That lib is PHP >=7 though, so nothing to do tomorrow.
I am going to merge this today. Maybe @thiemowmde manages to chime in here and prevent me from doing something silly

src/Term/TermList.php
```diff
@@ -66,7 +67,7 @@ public function toTextArray() {
 
 	/**
 	 * @see IteratorAggregate::getIterator
-	 * @return Traversable|Term[]
+	 * @return Traversable|Iterator|Term[]
```
thiemowmde: I'm fine with stating Iterator here. I remember when I had to decide what to use, and I went for the more narrow Traversable because that's all that is needed in a foreach loop.

Note that the parent's documentation says "Iterator or Traversable".

Since Iterator extends Traversable, Traversable should be removed here. :+1: otherwise.

addshore:
I did this in the github UI, so lets squash and merge when this PR gets merged.

thiemowmde:
If you want to release this ASAP, don't forget to backport to the 7.x branch first.

PR #787, opened 2018-08-24
[WIP] Implement non-standard "Q" prefixes

nyurik:
Allow Wikibase installations to use prefixes
other than 'Q' and 'P' for their items.

This pull request should not be merged yet, but seeks to start
the discussion on the viability of this approach.

See https://phabricator.wikimedia.org/T202676


nyurik:
@thiemowmde prefixes won't work very well for the usecase described in my reply at https://phabricator.wikimedia.org/T202676#4529858   What would be the best way to solve that?

JeroenDeDauw:
Tracked and discussed on the linked phabricator ticket. This PR can be reopened if the approach it takes is deemed viable.

Thanks for contributing @nyurik

PR #788, opened 2018-09-12
Add TermList::add

JeroenDeDauw:
So TermListMerger in WikibaseLexeme can be removed

Note that TermListMerger does a not needed hasTerm check which has been omitted here

### DIFF
--- a/src/Term/TermList.php
+++ b/src/Term/TermList.php
@@ -200,7 +200,7 @@ class TermList implements Countable, IteratorAggregate, Comparable {
 	 *
 	 * @param iterable|Term[] $terms
 	 */
-	public function add( /* iterable */ $terms ) {
+	public function addAll( /* iterable */ $terms ) {
 		foreach ( $terms as $term ) {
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
I think I've been successfully brainwashed by Rich Hickey because needing to do basic collection stuff over and over again for different types is irking me.

This functionality (and many more possibilities) can also be achieved by letting the constructor accept `iterable`. Then you could simply do

```php
$mergedList = new TermList( new MultipleIterator( $firstList, $secondList ) );
```

Or use any of the other iterator/traversable/generator functionality PHP provides. And this opens the door to laziness.

Thoughts on that approach?

JeroenDeDauw:
Seems like the constructor approach will even be less nice than I thought. Bloody PHP. Geez.

image

Not only do you need to create this iterator class, you also can't add the child iterators in the constructor, and it actually only takes iterators rather than traversables such as TermList in general
 >_> omg

PR #789, opened 2018-09-13
Take iterable in TermList constructor

JeroenDeDauw:
Follow-up-ish to https://github.com/wmde/WikibaseDataModel/pull/788

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -2,7 +2,7 @@
 
 ## Version 8.1.0 (dev)
 
-* The `TermList` constructor now take any `iterable` instead of just `array`
+* The `TermList` constructor now takes any `iterable` instead of just `array`
 * Added `TermList::addAll`
 
 ## Version 8.0.0 (2018-08-03)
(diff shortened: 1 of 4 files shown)


RELEASE-NOTES.md
```diff
@@ -1,5 +1,10 @@
 # Wikibase DataModel release notes
 
+## Version 8.1.0 (dev)
+
+* The `TermList` constructor now take any `iterable` instead of just `array`
```
wiese: "now take**s**"

src/Term/TermList.php
```diff
@@ -199,9 +193,14 @@ public function clear() {
 	 * @since 8.1
 	 *
 	 * @param iterable|Term[] $terms
+	 * @throws InvalidArgumentException
 	 */
 	public function addAll( /* iterable */ $terms ) {
 		foreach ( $terms as $term ) {
+			if ( !( $term instanceof Term ) ) {
+				throw new InvalidArgumentException( 'Every element in $terms must be an instance of Term' );
```
wiese: Not sure if this is by (some) design but neither the constructor did have nor `addAll()` now appear to be having a test case that asserts the `InvalidArgumentException`.

JeroenDeDauw: Good catch, added those

PR #790, opened 2018-09-14
Drop PHP 5.5 support

JeroenDeDauw:


thiemowmde:
Note that, as far as I'm aware of, these tests won't run on the Wikimedia CI any more. I'm not sure how this can be resolved when the first patches get red flags on Gerrit. @addshore, since you are the tech lead for this code now, can you have a look if this is worth it?

JeroenDeDauw:
Huh? Are there tests run on Wikimedia CI somehow? That is pretty non-standard and you need to do something special so they even get loaded since they are `autoload-dev`

manicki:
> Are there tests run on Wikimedia CI somehow?

I agree they shouldn't be run. That's the point of this being a library.
There are things in Wikimedia CI which are rather non-standard, but I believe this is not the case here.
@addshore has been pinged, so he will definitely shed some ultimate light here!

addshore:
As far as I know the tests are not run anywhere in WMF CI.
My opinion also is that they should probably not be.

PR #802, opened 2018-11-06
Add SubEntityProvider interface

lucaswerkmeister:
This provides users with a generic way to enumerate the subentities of an entity.

Bug: [T204066](https://phabricator.wikimedia.org/T204066)

---

Note that I have omitted some important formation from the interface documentation: what *is* a subentity?

Do we have a clear definition of that? E. g. do the subentities of an entity share the parent entity’s storage mechanism (in Wikibase: MediaWiki pages)?

Would it be acceptable to merge this PR without any such definition?

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,9 @@
 # Wikibase DataModel release notes
 
+## Version 10.0.0 (dev)
+
+* Added `SubEntityProvider`
+
 ## Version 9.0.0 (2018-11-01)
 
 * Breaking change: `EntityDocument` no longer extends `ClearableEntity` (8.0.0 revert)
--- a/src/Entity/SubEntityProvider.php
+++ b/src/Entity/SubEntityProvider.php
@@ -1,0 +1,25 @@
+<?php
+
+
+namespace Wikibase\DataModel\Entity;
+
+/**
+ * Common interface for classes (typically entities) that may contain subentities.
+ *
+ * This only provides a way to enumerate the subentities;
+ * there is no generic mechanism for modifying them.
+ *
+ * @license GPL-2.0-or-later
+ */
+interface SubEntityProvider {
+
+	/**
+	 * Returns the subentities of this object,
+	 * not necessarily in any particular order.
+	 * Note that the returned list may be empty.
+	 *
+	 * @return EntityDocument[]
+	 */
+	public function getSubEntitities(): array;
+
+}


thiemowmde:
I'm curious. Am I right that, when applied on the Lexeme data model, this will return Forms and Senses as one array? What is the use case for this? Most code should know how to call `getForms` and `getSenses`. What code doesn't need to know this, and can still do something useful with an array of essentially unknown entity types?

Tpt:
An example of use case is the RDF serialization: with this interface the RDF serializer could call its delegate serializers for the main entity and its sub entities easily instead of relying on a delegate serializer to ask for the serialization of the sub entities.

thiemowmde:
Hm, this doesn't sound right to me. Why would you run an RDF serializer on an entity the serializer does not understand? How would the serializer create a meaningful output if it doesn't know it is running on a Lexeme?

Tpt:
The RDF serialization is using delegates: you have an RDF builder for creating the serialization of full statements (`FullStatementRdfBuilder`) of the entity if it implements `StatementListProvider`, a `PropertyRdfBuilder` that adds the property related triples only if the entity is a `Property`...
All these builders implements the same interface and are hold by an `RdfBuilder` that calls all its delegate on each `Entity`. This way it is very easy to add support of new entity types without having to change the root builder and to add/remove builders according to options (in order to, e.g., create the truthy dumps).

Having a `SubEntityProvider` interface would allow the `RdfBuilder` class to call its delegates on the sub entities instead of having the specific delegate for lexemes to have to call `RdfBuilder` back to serialize them

RELEASE-NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Wikibase DataModel release notes
 
+## Version 10.0.0 (dev)
```
JeroenDeDauw: 9.1 since there are no breaking changes

thiemowmde:
I think the better solution here is to pass these subentities as separate, individual entities to the RDF builder, and not expect it to traverse a tree structure it does not really need to understand. Is there anything the RDF builder could do with the information how entities are nested? I think it shouldn't, even if it could, because each subentity should (and does) contain enough information to be able to identify it's parent.

I think the issue your team is running into is this: Does the RDF builder traverse entity IDs or page Titles? You know, before subentities have been introduced there was a 1:1 mapping between the two (never guaranteed, but de-facto there). With sub-entities, multiple entity IDs can point to the same Content object. It appears like the RDF builder is either traversing a namespace, or when it traverses entity IDs it probably tries to process the whole serialization it finds in the Content object. In both cases the RDF builder is not traversing Forms and Senses as separate entities.

@brightbyte, do you have the time to say something about how subentities have been meant to behave in such situations?

brightbyte:
> @brightbyte, do you have the time to say something about how subentities have been meant to behave in such situations?

There are basically two choices, as you said: either the code that enumerates entities has to enumerate sub-entities as top level entities, or the code that processes the top-level entities has to recurse into sub-entities. The former seems nicer, but has more potential for confusion, since it means that *no* code triggered from such an enumeration must ever recurse into sub-entities, otherwise you'd get the twice (once nested, and once stand-alone). The latter is probably easier to do.

The solution for the RdfBuilder is to have LexemeRdfBuilder recurse into Forms and Senses after it is finished with it's own stuff. As far as I can tell, that's indeed how this is already implemented.

I see no problem with doing the same in a generic way:: if Wikibase supports the notion of sub-entities (and hierarchical IDs) anywhere, it can and should support it fully and transparently throughout the code. So if there is now agreement that this notion is something that Wikibase should support, handling for sub-entities should indeed be added to RdfBuilder, EntityTitleLookup, EntityRevisionLookup, EntityStore, etc.

I actually remember writing the relevant code for EntityRevisionLookup and EntityStore a couple of years ago. Maybe these patches are still hanging around somewhere.

On the other hand, if there is no commitment to support sub-entities throughout Wikibase, then the current handling via LexemeRdfBuilder is probably fine.

brightbyte:
...to come back to the enumeration idea: it would also be possible to have an EntityIdPager that "flattens" the sub-entity hierarchy on the fly. But this becomes tricky when you want to enumerate *just* the sub-entities, e.g. senses. Then you'd have to somehow know which top-level entities to look in. All doable, but doesn't seem simpler to do, nor easier to understand. I'd only go for that if there is indeed a need to enumerate the sub-entities without their parent entities.

manicki:
I don't want to interfere with constructive discussions (there might be several running in parallel in this thread actually), but to step, few remarks before you guys go to far into solving problems which might not be actual problems or so:
- the concept of "sub entity" is indeed not defined, and especially not defined in the scope of this library. Not knowing much of the background that lead to this pull request, I am nonetheless skeptical about modelling concepts that are not really defined. I am happy to be convinced of course.
- If all this is related to 1) generating RDF output, and 2) Lexemes, has it been considered to have a solution in a place which is closer to both of these?

TL;DR: there are many comments on this PR already. How about taking a step back, and thinking of two questions first: What is the problem we're trying to solve with the code? Where (what product, code base, use case) does this problem live/occur/happen?

lucaswerkmeister:
> What is the use case for this?

> I think the issue your team is running into is this:

> What is the problem we're trying to solve with the code?

…This is all answered right in the PR description? *My* motivation for this patch is [T204066](https://phabricator.wikimedia.org/T204066) – if it turns out to be useful in RDF generation as well, that’s nice, but I can’t really defend the patch against any criticisms coming from this angle :)

The issue in T204066 is that we want links from entities to Commons files to appear on Commons description pages. Wikibase currently implements this by iterating through the statements of an entity, and all the snaks in those statements (main, qualifier, reference), and looking for Commons references there. (It does something similar for collecting URLs, geographical coordinates, and some other things.) For lexicographical data, that isn’t enough: it also needs to iterate through the statements of a lexeme’s senses and forms. [I3964dac698](https://gerrit.wikimedia.org/r/472019) (Wikibase) and [I344a1371ae](https://gerrit.wikimedia.org/r/472020) (WikibaseLexeme) implement that, using the new interface introduced here: without knowing anything about WikibaseLexeme in particular, Wikibase can iterate through all the subentities of an entity, and then through their statements (if they’re `StatementListProvider`s) and so on.

brightbyte:
Ah, right. The alternative would be to have per-type extract-all-links callbacks.  Yet another reason to properly support sub-entities in Wikibase. I'm all for it, but I's not my call any more :)

manicki:
Thanks for explanation @lucaswerkmeister - I admit I couldn't connect https://phabricator.wikimedia.org/T204066 with RDF really, hence my uncertainty.

It looks like an instance of applying a rather generic solution to a specific/local problem. It might be okay to decide to use this approach, but the decision should probably be made considering its implications, and it might be worth looking into what alternatives would be.
It seems @JeroenDeDauw could suggest something, or others too. Please do!

brightbyte:
> The general approach I would pursue is to have WikibaseLexeme deal with Lexemes (and Senses and Forms) instead of adjusting Wikibase(Repository) to work with Lexemes. This way you avoid both indirection and strong binding across component boundaries. There might be some duplication in WikibaseLexeme then, though that is much cheaper than a poor abstraction. And it can be minimized by using composition of code defined in Wikibase(Repository).

On the other hand, the proliferation of special case solution, and the resulting duplication of code, due to the lack of proper abstraction, can also lead to unmaintainable code. We have both problems in MediaWiki core.

I don't think it's possible to decide whether pulling the idea of sub-entities into the wikibase data model makes things better or worse overall, purely based on principles. As most engineering decisions it's a treade-off, and the costs and benefits very much depend on the present needs, the present code structure, and the anticipated future.

JeroenDeDauw:
To me it seems overwhelmingly likely using standard design approaches will result in significantly more maintainable code than going with generic entity code. In other words, not a close trade-off at all.

JeroenDeDauw:
Will go with a different approach as per https://phabricator.wikimedia.org/T204066#4736308

PR #804, opened 2018-11-09
Fix earlier ClearbaleEntity revert

JeroenDeDauw:
This completes/fixes the revert started in https://github.com/wmde/WikibaseDataModel/pull/799

With this change the behavior is back to that of DM 7.5.x. (see https://github.com/wmde/WikibaseDataModel/blob/7.x/src/Entity/Item.php#L30 for verification)

Needed to make DM 9.x work in WB.git. Change in behavior spotted by WB.git tests

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,9 @@
 # Wikibase DataModel release notes
 
+## Version 9.0.1 (2018-11-09)
+
+* `Item` and `Property` now implement `ClearableEntity` again
+
 ## Version 9.0.0 (2018-11-01)
 
 * Breaking change: `EntityDocument` no longer extends `ClearableEntity` (8.0.0 revert)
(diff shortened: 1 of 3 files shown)


RELEASE-NOTES.md
```diff
@@ -1,5 +1,9 @@
 # Wikibase DataModel release notes
 
+## Version 9.0.1 (2018-11-09)
```
mariushoch: Shouldn't this rather be `9.1.0`? I don't think this is technically fixing a bug, given the (reverted) change was intended?

JeroenDeDauw: The revert was meant to undo to 8.0 changes and go back to the 7.5 behavior. I fucked up. So yeah, this is a bugfix. If it is not a bugfix, then it is a breaking change (10.0), which we have incentive to avoid (costly).

mariushoch: Ok, makes sense

PR #807, opened 2019-01-21
Update wikimedia/assert, minimum PHP version

tgr:
Upgrade wikimedia/assert to 0.4.0 (no major changes but needed to avoid incompatible dependencies in mediawiki/vendor). Upgrade PHP requirement to 7.x or HHVM (needed by wikimedia/assert 0.4 and the end-of-life for 5.6 was January 1, anyway).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: php
 dist: trusty
 
 php:
-  - 5.6
   - 7.0
   - 7.1
   - 7.2
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
 	"require": {
 		"php": ">=5.6.99",
 		"data-values/data-values": "~0.1|~1.0|~2.0",
-		"wikimedia/assert": "~0.4.0"
+		"wikimedia/assert": "~0.2.2|~0.3.0|~0.4.0"
 	},
 	"require-dev": {
 		"ockcyp/covers-validator": "~0.5.0",
--- a/composer.json
+++ b/composer.json
@@ -23,9 +23,9 @@
 		"irc": "irc://irc.freenode.net/wikidata"
 	},
 	"require": {
-		"php": ">=5.6",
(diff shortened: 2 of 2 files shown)


tgr:
@JeroenDeDauw @thiemowmde not sure what to do here. Scrutinizer claims to autodetect the PHP version but apparently does not do a very good job of it, despite it being right there in composer.json. Should I pin it to 7.0?

Also, this needs to be released as 10.0.0 to cut the dependency knot in mediawiki/vendor. I assumed the release notes bump should be its own commit but can add it here if preferred.

(Or an alternative approach would be to relax the wikimedia/assert version and allow anything between 0.2-0.4, release the change as a new patch version, and hope composer can figure out that it needs to install the old version of wikimedia/assert when on PHP 5.6. But I'm not sure it can, specifying multiple major versions is not nice, and 5.6 is end-of-lifed anyway so I figured this approach is preferable.)

thiemowmde:
> an alternative approach would be to relax the wikimedia/assert version and allow anything between 0.2-0.4 […]

I would go for this, to be honest.

tgr:
Thanks! (Although at this point there was no need to bump the PHP version, IMO.)

JeroenDeDauw:
Then why are you submitting a change that does that????

tgr:
As I said, IMO the two options are:
* set wikimedia/assert to `~0.4.0`, bump PHP version, do major release
* set wikimedia/assert to `~0.2.2|~0.3.0|~0.4.0`, do patch release, hope composer is clever enough to install an old version of it on old PHP

What I meant is that if you do the version bump no point in retaining compatibility with the older versions IMO. Or was that done to reduce the chance of dependency conflicts in non-MediaWiki projects?

Anyway, as long as it unblocks the wikimedia/assert version upgrade in mediawiki/vendor, I'm happy with that :) Thanks for merging.

PR #808, opened 2019-01-23
9.1.0 release

tgr:


JeroenDeDauw:
Bumping PHP version (or that of other packages) is not a breaking change. So this should be 9.1

JeroenDeDauw:
And if it where a breaking change, it would not be worth making, since major releases of this package are costly. See https://www.entropywins.wtf/blog/2018/08/30/base-libraries-should-be-stable/

addshore:
We just merged something else and will likely make a 9.1.0 (including that other thing) today

PR #809, opened 2019-01-29
Prevent redirecting an entity to itself

Ladsgroup:
[Bug: T214920](https://phabricator.wikimedia.org/T214920)

### DIFF
--- a/src/Entity/EntityRedirect.php
+++ b/src/Entity/EntityRedirect.php
@@ -37,6 +37,10 @@ class EntityRedirect {
 			);
 		}
 
+		if ( $entityId->getSerialization() === $targetId->getSerialization() ) {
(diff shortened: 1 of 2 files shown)


src/Entity/EntityRedirect.php
```diff
@@ -37,6 +37,10 @@ public function __construct( EntityId $entityId, EntityId $targetId ) {
 			);
 		}
 
+		if ( $entityId->getSerialization() === $targetId->getSerialization() ) {
```
thiemowmde: Note since repo prefixes have been introduced it's possible to have different serializations that – when resolved – all refer to the same entity. For example `foo:bar:Q1` vs. `bar:foo:Q1`, which might or might not be the same entity.

But I think this is unrelated and no reason to not have this basic check. From my perspective it makes a lot of sense.

PR #812, opened 2019-04-01
Always use vendor PHPUnit

JeroenDeDauw:


manicki:
meh, this HHVM thing on travis is annoying. Gonna fix it.

JeroenDeDauw:
You can just merge this since only hhvm failed for known other reason 
image

PR #814, opened 2019-04-02
Use HHVM 3.18 on Travis CI

manicki:


manicki:
I'm confused about Scrutinizer failure, looking into it

manicki:
I am not claiming the way I "fixed" Scrutinizer here is the right one, so please feel free to adjust the PR with something better if you know it!

manicki:
that was fast :)

PR #815, opened 2019-05-28
Optimize EntityId::isForeign()

lucaswerkmeister:
When this method was first introduced, `getRepositoryName()` would split the serialization each time it was called, and using `strpos()` in `isForeign()` was an optimization to avoid that cost. However, since #769 (commit eb6656155b), the serialization is only split once in the constructor, so checking `strpos` each time instead of deferring to the already populated `$repositoryName` is probably slightly more expensive.

As a very minor optimization, we directly access `$this->repositoryName` instead of calling `getRepositoryName()`. Subclasses that for some reason override `getRepositoryName()` to do something different must override `isForeign()` as well.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -4,6 +4,7 @@
 
 * `TermList` now throws `InvalidArgumentException` when given non-iterable rather than failing silently
 * `SiteLinkList` now throws `InvalidArgumentException` when given non-iterable rather than failing silently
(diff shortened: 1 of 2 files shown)


lucaswerkmeister:
> As a very minor optimization, we directly access `$this->repositoryName` instead of calling `getRepositoryName()`.

And no, PHP7 really doesn’t inline this trivial getter. HHVM does, but we’re moving away from that. (I didn’t test this case specifically, but I tried a similar case a while ago and there was a clear difference on PHP7.)

PR #816, opened 2019-06-06
Make use of language features like end() and array_pad()

thiemowmde:


mariushoch:
I don't like that `end` interferes with the internal array pointer… I guess this can have weird side-effects later :/

thiemowmde:
I'm not aware of any code that makes use of the internal array pointer, or assumes it is in a specific position. To my knowledge such code just doesn't exist in our realm. Furthermore, the arrays in question here are, if I'm not mistaken, private implementation details where the array pointer can't have any influence on the public interface of this code.

But feel free to use something else (e.g. `array_pop`) and update this patch accordingly if you feel better about it. I won't have the time to continue working on this the next months. It would be sad when this is still sitting here when I come back. Better revert parts that are somewhat controversial and merge the rest. Thanks! :-)

mariushoch:
None of these arrays seem to leave our scope here (they aren't returned or given by reference AFAIS), so let's just do this :)

PR #820, opened 2020-01-24
Update code sniffer to 1.1.0

addshore:


tests/unit/Assert/RepositoryNameAssertTest.php
```diff
@@ -50,6 +50,7 @@ public function provideValidRepositoryNames() {
 	 */
 	public function testGivenValidValue_assertParameterIsValidRepositoryNamePasses( $value ) {
 		RepositoryNameAssert::assertParameterIsValidRepositoryName( $value, 'test' );
+		$this->assertTrue( true );
```
thiemowmde: Note while this is a valid and very common workaround, there a dedicated feature for situations like this: `$this->addToAssertionCount(1);`

addshore: I thought there was but couldn't' find the doc for it at the time.
Will fix!

addshore: Done in https://github.com/wmde/WikibaseDataModel/commit/a55334a301d09d57f5d378a72fbbd1335f7d290f

PR #821, opened 2020-01-30
Add TermTypes interface with term type constants

jakobw:
Useful when passing term types to services like TermBuffer.

### DIFF
--- a/src/Term/TermTypes.php
+++ b/src/Term/TermTypes.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace Wikibase\DataModel\Term;
+
+/**
+ * Term types used on Items and Properties
+ *
+ * @since 9.3
+ *
+ * @license GPL-2.0-or-later
+ */
+final class TermTypes {
+
+	public const TYPE_LABEL = 'label';
+	public const TYPE_DESCRIPTION = 'description';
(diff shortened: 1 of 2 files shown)


src/Term/TermTypes.php
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * @since 9.3
+ *
+ * @license GPL-2.0-or-later
+ */
+interface TermTypes {
+
+	public const TYPE_LABEL = 'label';
+	public const TYPE_DESCRIPTION = 'description';
+	public const TYPE_ALIAS = 'alias';
```
Ladsgroup: Maybe having one for aliases would be good too? Just thinking out loud.

jakobw: The scary thing is that some services (e.g. `TermBuffer` and by association `PrefetchingTermLookup` and friends) expect to be passed `'alias'` not `'aliases'` to fetch all aliases. That can probably be solved by documenting it properly, but my original motivation was actually to have these constants to avoid exactly this mix-up.

JeroenDeDauw:
Maybe `final class` is better than `interface`

jakobw:
> Maybe `final class` is better than `interface`

Done!

PR #823, opened 2020-02-27
Add getter for statement guid part

matthiasmullie:
This would be useful for reconstructing another StatementGuid based on its existing parts.
EntityId can already be gotten from a StatementGuid object, and now the GUID would too - that's the 2 required arguments for constructing a new StatementGuid with the same (or derived) data.

Use case: https://gerrit.wikimedia.org/r/c/mediawiki/extensions/WikibaseMediaInfo/+/573315/1/src/WikibaseMediaInfoHooks.php

### DIFF
--- a/src/Statement/StatementGuid.php
+++ b/src/Statement/StatementGuid.php
@@ -31,6 +31,11 @@ class StatementGuid implements Comparable {
 	/**
 	 * @var string
 	 */
+	private $guid;
+
+	/**
+	 * @var string
+	 */
 	private $serialization;
 
 	/**
@@ -46,6 +51,7 @@ class StatementGuid implements Comparable {
 
 		$this->serialization = $entityId->getSerialization() . self::SEPARATOR . $guid;
 		$this->entityId = $entityId;
+		$this->guid = $guid;
 	}
 
 	/**
@@ -53,6 +59,13 @@ class StatementGuid implements Comparable {
 	 */
 	public function getEntityId() {
 		return $this->entityId;
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getGuid() {
+		return $this->guid;
 	}
 
 	/**
--- a/src/Statement/StatementGuid.php
+++ b/src/Statement/StatementGuid.php
@@ -31,7 +31,7 @@ class StatementGuid implements Comparable {
 	/**
 	 * @var string
 	 */
-	private $guid;
+	private $guidPart;
 
 	/**
 	 * @var string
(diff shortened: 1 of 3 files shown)


addshore:
I have replied to this patch, and the gerrit patch and comment there at https://phabricator.wikimedia.org/T244207#5954555

matthiasmullie:
Any other reservations? Can we merge this?

src/Statement/StatementGuid.php
```diff
@@ -46,6 +51,7 @@ public function __construct( EntityId $entityId, $guid ) {
 
 		$this->serialization = $entityId->getSerialization() . self::SEPARATOR . $guid;
 		$this->entityId = $entityId;
+		$this->guid = $guid;
```
addshore: Lets call this guidPart just to avoid any confusion for people looking in there for the first time.

src/Statement/StatementGuid.php
```diff
@@ -55,6 +61,13 @@ public function getEntityId() {
 		return $this->entityId;
 	}
 
+	/**
+	 * @return string
+	 */
+	public function getGuid() {
```
addshore: @since 9.4 please

addshore: We probably also want to call this getGuidPart.

Otherwise someone will get confused at some point as we have $statement->getGuid which returns a string which is the whole guid, and then $guid->getGuid which is just the later part.

src/Statement/StatementGuid.php
```diff
@@ -55,6 +61,13 @@ public function getEntityId() {
 		return $this->entityId;
 	}
 
+	/**
+	 * @return string
```
JeroenDeDauw: You can ditch the comment in favor of a return type hint.

JeroenDeDauw:
> Please also add this to a 9.4 section in the release notes and then LGTM

Just merge and add it yourself ;) Most stuff was added to release notes after merge in the past. Less merge conflicts that way.

addshore:
> > Please also add this to a 9.4 section in the release notes and then LGTM
> 
> Just merge and add it yourself ;) Most stuff was added to release notes after merge in the past. Less merge conflicts that way.

Totally fine with adding it myself, but also, this PR is being touched again anyway so would be just as easy to add the relnotes then, especially as it doesn't look like we will be merging anything else for 9.4

src/Statement/StatementGuid.php
matthiasmullie: Happy to, but looks like there aren't any return type hints in this repo yet? (or at least not in the few files I looked at)

PR #824, opened 2020-05-26
WikibaseDataModel: PHP 7.4 CI fails

toban:
PHP 7.4 exposes two new magic functions __serialize and __unserialize
for custom object serialization. These are required to be implemented to
support migration from older PHP versions.

see https://wiki.php.net/rfc/custom_object_serialization for more
detail.

Bug: T243590

### DIFF
--- a/src/ReferenceList.php
+++ b/src/ReferenceList.php
@@ -26,7 +26,6 @@ use Wikibase\DataModel\Snak\Snak;
  * @author Bene* < benestar.wikimedia@gmail.com >
  *
  * @phpcs:disable MediaWiki.NamingConventions.LowerCamelFunctionsName.FunctionName
- * @SuppressWarnings(PHPMD.TooManyPublicMethods)
  */
 class ReferenceList implements Comparable, Countable, IteratorAggregate, Serializable {
 
--- a/src/ReferenceList.php
+++ b/src/ReferenceList.php
@@ -24,6 +24,9 @@ use Wikibase\DataModel\Snak\Snak;
  * @author H. Snater < mediawiki@snater.com >
  * @author Thiemo Kreuz
  * @author Bene* < benestar.wikimedia@gmail.com >
+ *
+ * @phpcs:disable MediaWiki.NamingConventions.LowerCamelFunctionsName.FunctionName
+ * @SuppressWarnings(PHPMD.TooManyPublicMethods)
  */
 class ReferenceList implements Comparable, Countable, IteratorAggregate, Serializable {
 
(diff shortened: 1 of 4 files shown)


JeroenDeDauw:
Looks good apart from the style issues

JeroenDeDauw:
7.4 tests still fail: https://travis-ci.org/github/wmde/WikibaseDataModel/jobs/691215198

toban:
> 7.4 tests still fail: https://travis-ci.org/github/wmde/WikibaseDataModel/jobs/691215198

Yes, there is one last test that does not pass without altering the expected output for 7.4. I suspect this is related to what is described here https://www.php.net/manual/en/migration74.incompatible.php but unsure if that is the case and what the actual consequences would be for allowing a schema change which in turn could generate changes to hash values. Waiting for insights from @thiemowmde @addshore

addshore:
So having a fresh look at this today I do not think that the stability of the ReferenceList getValueHash is important. Other hashes are important and we should check to see if we have stability tests for "getHash" methods in all other Datamodel data objects IMO so that we have some visibility on what is changing when (I believe some of these already exist).

If anything else is changing also with 7.4 then we perhaps need to make some more changes here regarding the calculation of the hashes.

Thinking about how this will end up being rolled out on Wikidata.org, there will be a period of time where both 7.4 and another PHP version run side by side, meaning that we will have different request executions running and using different hashes internally for reference lists.
For that reason the cleanest way forward might be to release a new major version of WikibaseDataModel that includes a different hash calculation for at least RefernceList (and anything else that will have a changed hash as detected by the other stability checks we want to add / make sure that we have.

This would allow us to control exactly when the change in hashes would happen, rather than being in some odd middle ground.

addshore:
Sorry, I started that PR review before the last force push and since then CI is now green.
I'll review this now and I'm pretty sure I can get it merged today.
Another followup to ensure nothing odd is happening would be to make sure that all datamodel objects have hash stability tests and nothing is changing without us noticing. (I'll write a phab ticket for that)

src/ReferenceList.php
```diff
@@ -24,6 +24,9 @@
  * @author H. Snater < mediawiki@snater.com >
  * @author Thiemo Kreuz
  * @author Bene* < benestar.wikimedia@gmail.com >
+ *
+ * @phpcs:disable MediaWiki.NamingConventions.LowerCamelFunctionsName.FunctionName
+ * @SuppressWarnings(PHPMD.TooManyPublicMethods)
```
addshore, line 29: How would you feel about altering the mess detector configuration to also add __ to the ignorepattern for the TooManyPublicMethods rule?
I think that might make sense, and then this suppression could also be removed.

src/ReferenceList.php
```diff
@@ -24,6 +24,9 @@
  * @author H. Snater < mediawiki@snater.com >
  * @author Thiemo Kreuz
  * @author Bene* < benestar.wikimedia@gmail.com >
+ *
+ * @phpcs:disable MediaWiki.NamingConventions.LowerCamelFunctionsName.FunctionName
```
addshore, line 28: Looks like the mediawiki-codesniffer needs to be fixed for these new magic methods
https://phabricator.wikimedia.org/T253633
Patch already up https://gerrit.wikimedia.org/r/598745

Fine to merge it as it is, and try to remember to remove the disable line when we get to use the new version!
(I could even file a ticket for that)

toban: Ok cool, I think this might be a good route!

src/ReferenceList.php
toban: Yes i can try adding this now!

PR #830, opened 2020-10-16
Remove unneeded @phpcs:disable comments

lucaswerkmeister:


micgro42:
Thanks, looks good so far. Could you check if we also still need the extended ignore syntax added in this line:

https://github.com/wmde/WikibaseDataModel/pull/824/files#diff-86663aeec12838bb61855e163b38da1d1e38d18276b2d25c9dd6761127914a43R18

micgro42 (approved):
Mh, we probably do, the line mentioned above is only related to the maxmethods rule.  Will merge once Travis is happy.

lucaswerkmeister:
Hm, locally `composer ci` passes even if I completely remove that `ignorepattern` line.

PR #832, opened 2020-10-21
Retire interfaces

itamargiv:
Removes interface usages as they were blocking the [release of DataValues/DataValues v 3.0.0](https://github.com/DataValues/DataValues/issues/58#issuecomment-700998895)

Related to: [T260915](https://phabricator.wikimedia.org/T260915)

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -9,6 +9,7 @@
   please update your code:
   for instance, change `->addNewReference( [ $x, $y ] )` to `->addNewReference( $x, $y )`,
   and `->addNewReference( $snaks )` to `->addNewReference( ...$snaks )`.
+* `Statement`, `Reference`, `SnakList` and `Snak` no longer depend on the `Hashable` and `Immutable` interfaces from `DataValues/DataValues`.
 
 ## Version 9.5.1 (2020-06-03)
 
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -9,7 +9,7 @@
   please update your code:
   for instance, change `->addNewReference( [ $x, $y ] )` to `->addNewReference( $x, $y )`,
   and `->addNewReference( $snaks )` to `->addNewReference( ...$snaks )`.
-* `Statement`, `Reference`, `SnakList` and `Snak` no longer depend on the `Hashable` and `Immutable` interfaces from `DataValues/DataValues`.
+* `Statement`, `Reference`, `SnakList` and `Snak` no longer implement the `Hashable` and `Immutable` interfaces from `DataValues/DataValues`.
 
 ## Version 9.5.1 (2020-06-03)
 
--- a/src/Snak/SnakList.php
+++ b/src/Snak/SnakList.php
@@ -208,6 +208,9 @@ class SnakList extends ArrayObject implements Comparable {
 	 */
 	public function offsetUnset( $index ) {
 		if ( $this->offsetExists( $index ) ) {
+			/**
+			 * @var Snak $element
+			 */
 			$element = $this->offsetGet( $index );
 			$hash = $element->getHash();
 			unset( $this->offsetHashes[$hash] );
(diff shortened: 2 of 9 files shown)


src/Snak/SnakList.php
```diff
@@ -211,9 +208,6 @@ private function getNewOffset() {
 	 */
 	public function offsetUnset( $index ) {
 		if ( $this->offsetExists( $index ) ) {
-			/**
-			 * @var Hashable $element
```
lucaswerkmeister, line 215: I assume this could also be `@var Snak $element`? Or can the `$element` also be something else?

tests/unit/ReferenceListTest.php
```diff
@@ -368,9 +367,6 @@ public function testGetValueHashIsTheSameForClone( ReferenceList $array ) {
 	public function testHasReferenceHash( ReferenceList $references ) {
 		$this->assertFalse( $references->hasReferenceHash( '~=[,,_,,]:3' ) );
 
-		/**
-		 * @var Hashable $reference
```
lucaswerkmeister, line 372: Similarly, I assume `$reference` is here known to be a `Reference`.

lucaswerkmeister (approved):

src/Snak/SnakList.php
itamargiv: I just didn't see the reason to keep it as there was no consistent rule regarding where an interface or class of a variable is mentioned, don't mind putting it back though if it's a blocker.

lucaswerkmeister: I think the rule is more or less to include `@var` comments wherever “the IDE” can’t figure out the type on its own; in this case, `offsetGet` is a general `ArrayAccess` method, so there’s no way for the IDE to know that `$element` will be anything other than `mixed`. (At least, this is still the case in IntelliJ 202.7502.)

lucaswerkmeister (approved):
I was thinking about whether this should be added to the release notes, but that actually leads into the question of whether the next version (with this change) should be 9.6.0 or 10.0.0, so maybe it’s okay if we defer that for now and merge this pull request first.

itamargiv:
> I was thinking about whether this should be added to the release notes, but that actually leads into the question of whether the next version (with this change) should be 9.6.0 or 10.0.0, so maybe it’s okay if we defer that for now and merge this pull request first.

I think that if we want to document this in the release notes we should add it now so we don't for get to do that later, and decide on whether it's a minor or major release later (if it's just this change that we're talking about I would see it as a patch, since it is not a breaking change, and we are not introducing a new feature).

RELEASE-NOTES.md
```diff
@@ -9,6 +9,7 @@
   please update your code:
   for instance, change `->addNewReference( [ $x, $y ] )` to `->addNewReference( $x, $y )`,
   and `->addNewReference( $snaks )` to `->addNewReference( ...$snaks )`.
+* `Statement`, `Reference`, `SnakList` and `Snak` no longer depend on the `Hashable` and `Immutable` interfaces from `DataValues/DataValues`.
```
lucaswerkmeister, line 12: I find “no longer depend on” a strangle way to phrase this… they no longer implement those interfaces, and that’s a potentially breaking change to users of these classes. The current phrasing makes it sound too harmless, in my opinion – it sounds like the same functionality is being provided, just without an internal dependency.

lucaswerkmeister:
In my opinion it’s a breaking change in principle… but I could see one argument against bumping the major version, that we’d also be bumping the data-values dependency, and so if someone else uses data-values and this library, they can’t “accidentally” or “automatically” get the incompatible data model version, because composer would only install the latest version compatible with the older data-values.

RELEASE-NOTES.md
itamargiv: I was under the impression we were providing the same functionality in each of the implementing classes already. But of course you're right in case people decide to check for `Hashable` or `Implementable` in their code then it is a breaking change,  It just didn't cross my mind at the time.

itamargiv:
@lucaswerkmeister Sorry, I completely forgot that we were working on this change until I saw this PR on my dashboard. Let me know if there's anything else to change.

lucaswerkmeister (approved):

PR #833, opened 2020-12-03
Also test with PHP 8.0

JeroenDeDauw:


JeroenDeDauw:
This does not work since the package has indirect dev dependencies on stuff that also pins to 7.x. Can by removing WB codesnifer

PR #845, opened 2021-09-22
build: Update mediawiki/codesniffer to 37.0.0

ZabeMath:


ZabeMath:
Pushed change to gerrit instead.

